PR #10, opened 2026-05-21
Fix #9: Wire generator tasks as dependencies of Jar tasks

### LINKED ISSUE #9: sourcesJar (KMP root) consumes generated sources without a task dependency
Follow-up to #7. After upgrading to 0.4.3 the generated sdbus tasks are correctly wired as inputs to `compileKotlinLinuxX64` / `compileKotlinLinuxArm64` / `compileLinuxMainKotlinMetadata`, but the **KMP aggregate `sourcesJar` task** (the one vanniktech-maven-publish uses to build the `-sources.jar` artifact) still picks up the generated source directory without a declared dependency.

## Repro

Same minimal project from #8 (or any KMP module that applies `com.monkopedia.sdbus.plugin` and has sources under `src/dbus/`), with `com.vanniktech.maven.publish` applied to produce source jars. On Gradle 9:

```
$ ./gradlew :engine:publishToMavenLocal

> Task :engine:sourcesJar FAILED
  Reason: Task ':engine:sourcesJar' uses this output of task
  ':engine:generateSdbusWrappersOrg.bluez.Adapter1' without declaring
  an explicit or implicit dependency.
  (repeated for every generated interface)
```

`compileKotlinLinuxX64` etc. run fine — the wiring from #7 covers them. Only the root `sourcesJar` is missed. `linuxX64SourcesJar` / `linuxArm64SourcesJar` might have the same issue, I haven't tested them in isolation.

## Workaround

```kotlin
tasks.matching { it.name == "sourcesJar" || it.name.endsWith("SourcesJar") }
    .configureEach { dependsOn("generateSdbusWrappers") }
```

## Fix sketch

In the gradle plugin, after wiring the generator into the Kotlin compile tasks, also attach it to any `Jar` task that includes the generated directory in its `from(...)`. Kotlin Multiplatform creates one source jar per target plus an aggregate root `sourcesJar` when the publish plugins are applied; the plugin needs to cover all of them, not just the compile tasks.

## Environment

- Gradle 9.4.1
- Kotlin 2.3.20
- sdbus-kotlin 0.4.3
- com.vanniktech.maven.publish 0.36.0

monkopedia-coder:
## Summary

- Wires `Jar` tasks (including `sourcesJar` and target-specific `*SourcesJar`) to depend on the `generateSdbusWrappers` root task, fixing the Gradle 9 missing-dependency error when maven-publish produces source jars.
- Adds a test verifying `sourcesJar --dry-run` includes `generateSdbusWrappers` in its task graph.

Fixes #9

## Test plan

- [x] `plugin:test` passes (6 tests including the new `wiresGeneratorAsDependencyOfSourcesJar`)
- [ ] CI: full test suite on x64
- [ ] CI: ARM build + test

### DIFF
--- a/plugin/src/main/kotlin/com/monkopedia/sdbus/plugin/SdbusPlugin.kt
+++ b/plugin/src/main/kotlin/com/monkopedia/sdbus/plugin/SdbusPlugin.kt
@@ -4,6 +4,7 @@ import com.monkopedia.sdbus.capitalized
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.jetbrains.kotlin.gradle.dsl.kotlinExtension
+import org.gradle.jvm.tasks.Jar
 import org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompileTool
 
 class SdbusPlugin : Plugin<Project> {
@@ -18,6 +19,9 @@ class SdbusPlugin : Plugin<Project> {
                 }
             }
             target.tasks.withType(AbstractKotlinCompileTool::class.java).configureEach { task ->
+                task.dependsOn(rootTask)
+            }
+            target.tasks.withType(Jar::class.java).configureEach { task ->
                 task.dependsOn(rootTask)
             }
             ext.sources.asFileTree.filter { it.isFile && it.extension == "xml" }.forEach { file ->
--- a/plugin/src/test/kotlin/com/monkopedia/sdbus/plugin/SdbusPluginTest.kt
+++ b/plugin/src/test/kotlin/com/monkopedia/sdbus/plugin/SdbusPluginTest.kt
@@ -61,6 +61,20 @@ class SdbusPluginTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Looks good. The change is minimal, correct, and follows the existing `AbstractKotlinCompileTool` wiring pattern exactly. Wiring all `Jar` tasks via `withType(Jar::class.java).configureEach` is the right approach — it catches `sourcesJar`, target-specific `*SourcesJar`, and any other jar tasks that consume the generated source directory.

The new test (`wiresGeneratorAsDependencyOfSourcesJar`) is solid: it applies `maven-publish` to get the `sourcesJar` task, does a `--dry-run`, and asserts the generator appears in the task graph.

**Category:** `bug_fix` (targeted fix for Gradle 9 missing-dependency error, no API surface change)
**Tier:** 1 (default for `bug_fix`, no project override for sdbus-kotlin)
**Verdict:** approve — auto-merging per policy.

PR #22, opened 2026-06-06
Post-2.4.0 warning + hygiene cleanup (#17, #19, #20, #21)

### LINKED ISSUE #17: Add -Xexpect-actual-classes to the jvm target to silence ~20 Beta warnings
**Category:** inconsistency
**Where:** `build.gradle.kts` — `kotlin { jvm { compilerOptions { ... } } }` (lines 45-53), versus `linuxX64`/`linuxArm64` (lines 63-70, 82-89)

**Current:** `-Xexpect-actual-classes` is passed via `freeCompilerArgs` to the `linuxX64` and `linuxArm64` targets, but the `jvm` target only sets `jvmTarget`. Because the library uses `expect`/`actual` classes extensively (`Message`, `MethodCall`, `MethodReply`, `PlainMessage`, `Proxy`, `Types`, `Signal`, `TypeTraits`, plus their `.jvm` actuals), the JVM compile emits ~20 warnings of the form:
```
w: ...Message.kt:37:1 'expect'/'actual' classes ... are in Beta. Consider using the '-Xexpect-actual-classes' flag to suppress this warning.
```
The native targets suppress these; the JVM target does not, so the same intentional pattern produces a wall of warnings on only one target.

**Proposed:** Add `-Xexpect-actual-classes` to the `jvm` target's `compilerOptions.freeCompilerArgs` (matching what the native targets already do), so the suppression is applied consistently across all targets.

**Why:** Removes ~20 build warnings on the JVM compile that are pure noise (the project already opted into this pattern deliberately on native). Quieter builds make genuine new warnings visible; the flag choice is already established elsewhere in the same file, so this just makes it consistent.

**Effort:** mechanical

### LINKED ISSUE #19: Remove dead OSSRH publishing block + snapshot repo from codegen/build.gradle.kts
**Category:** dead-code
**Where:** `codegen/build.gradle.kts` — `mavenPublishing { repositories { maven(OSSRH) { ... } } }` (lines 106-114), and the snapshot repo at line 17. Plus a related deprecation at line 58.

**Current:** The codegen module's `mavenPublishing` block declares a manual OSSRH staging repository with credentials from `MAVEN_USERNAME`/`MAVEN_PASSWORD`:
```kotlin
repositories {
    maven(url = "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
        name = "OSSRH"
        credentials {
            username = System.getenv("MAVEN_USERNAME")
            password = System.getenv("MAVEN_PASSWORD")
        }
    }
}
publishToMavenCentral(automaticRelease = true)
```
But `.github/workflows/publish.yaml` never sets `MAVEN_USERNAME`/`MAVEN_PASSWORD` — it sets `ORG_GRADLE_PROJECT_mavenCentralUsername`/`mavenCentralPassword`, which feed the Vanniktech plugin's `publishToMavenCentral(...)`. The root `build.gradle.kts` publishes correctly with *only* `publishToMavenCentral(...)` and no manual `repositories {}` block. So the OSSRH block in codegen is a dead publish target whose env vars are never populated (it predates the current Vanniktech-based publishing). Line 17's `maven("https://oss.sonatype.org/content/repositories/snapshots/")` is likewise an unused snapshot source.

**Proposed:** Delete the manual `repositories { maven(OSSRH) { ... } }` block (lines 106-114) and the snapshot repo on line 17, leaving codegen's publishing to match root's clean `publishToMavenCentral`-only pattern. While in this file, also fix the related deprecation on line 58: `freeCompilerArgs.add("-Xjvm-default=all-compatibility")` now warns `-Xjvm-default is deprecated. Use -jvm-default instead.` under Kotlin 2.4.0 — switch to `-jvm-default=all-compatibility`.

**Why:** Removes dead, misleading publish config (a maintainer reading it would think OSSRH creds are wired when they aren't) and aligns codegen with the root module's publishing pattern; clears the `-Xjvm-default` deprecation warning introduced by the Kotlin bump. Verify against `publish.yaml` before removing.

**Effort:** mechanical

### LINKED ISSUE #20: Add missing @OptIn(ExperimentalXmlUtilApi) in Xml2Kotlin.kt (7 warnings)
**Category:** hygiene
**Where:** `codegen/src/main/kotlin/Xml2Kotlin.kt` — the `policy = object : DefaultXmlSerializationPolicy(policy) { ... }` override (lines 114-122); file opt-in at line 69

**Current:** The file opts in only to `ExperimentalSerializationApi` (`@file:OptIn(ExperimentalSerializationApi::class)`, line 69), but the custom `DefaultXmlSerializationPolicy` subclass and its `handleUnknownContentRecovering` override touch `nl.adaptivity.xmlutil` API that requires `ExperimentalXmlUtilApi`. Result is 7 warnings on every codegen build (confirmed):
```
w: ...Xml2Kotlin.kt:115:30 Base declaration of supertype 'DefaultXmlSerializationPolicy' needs opt-in.
   ... The declaration override should be annotated with '@nl.adaptivity.xmlutil.ExperimentalXmlUtilApi'
   or '@OptIn(nl.adaptivity.xmlutil.ExperimentalXmlUtilApi::class)'
w: ...Xml2Kotlin.kt:114:13 This XML Serialization code is not stable ...
(and 5 more at 114/121)
```

**Proposed:** Add `ExperimentalXmlUtilApi` to the opt-in — either extend the file-level annotation to `@file:OptIn(ExperimentalSerializationApi::class, ExperimentalXmlUtilApi::class)` (with the matching import), or scope a `@OptIn(ExperimentalXmlUtilApi::class)` to the policy override. This is the clean opt-in the compiler explicitly asks for; the experimental API usage is load-bearing (the custom policy is how the generator tolerates unknown D-Bus introspection content), so suppressing via the documented opt-in — not removing the code — is correct.

**Why:** Clears all 7 warnings with the exact annotation the compiler recommends, and makes the experimental-API dependency explicit and intentional rather than implicitly warned-about on every build.

**Effort:** mechanical

### LINKED ISSUE #21: codegen test pins kotlin-compiler-embeddable to 2.3.0 (toolchain is 2.4.0)
**Category:** hygiene
**Where:** `codegen/build.gradle.kts:25` — `testImplementation("org.jetbrains.kotlin:kotlin-compiler-embeddable:2.3.0")`; consumed by `codegen/src/test/kotlin/CompilationIntegrationTest.kt`

**Current:** `CompilationIntegrationTest` invokes `K2JVMCompiler` from `kotlin-compiler-embeddable` to actually compile the generated proxy/adaptor/interface Kotlin and assert it builds clean. That dependency is hardcoded to `2.3.0`, while the whole repo migrated to Kotlin **2.4.0** (`gradle/libs.versions.toml`: `kotlin = "2.4.0"`). So the codegen output is validated against a 2.3.0 compiler, not the 2.4.0 compiler the rest of the build and downstream consumers use — a version skew that predates the bump and could let 2.4.0-specific codegen issues slip through (or fail for stale reasons).

**Proposed:** Track the catalog Kotlin version instead of hardcoding: add a `kotlin-compiler-embeddable` entry to `libs.versions.toml` keyed to the existing `kotlin` version (`module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin"`) and reference it here. That keeps the compile-integration test on the same compiler as the toolchain automatically on future bumps. (Optional, same file: the `junit-jupiter-*:5.8.2` test pins on lines 27-28 are also hardcoded and stale, but that's lower-priority and separable.)

**Why:** The whole point of `CompilationIntegrationTest` is to prove generated code compiles on the project's Kotlin; pinning a different (older) compiler undermines that guarantee and is exactly the kind of skew a 2.4.0 migration should sweep up. Catalog-referencing it makes the test self-maintaining.

**Effort:** mechanical

monkopedia-coder:
Batches the four mechanical, user-approved code-health cleanups surfaced by the Kotlin 2.4.0 migration. No library runtime/behavior change; build-config + codegen only.

Closes #17 — add `-Xexpect-actual-classes` to the `jvm` target (native targets already had it) → ~20 expect/actual Beta warnings gone on the JVM compile.
Closes #19 — remove the dead OSSRH publishing block (`MAVEN_USERNAME`/`MAVEN_PASSWORD` the workflow never sets; real publish is vanniktech `publishToMavenCentral`) + the unused `oss.sonatype` snapshot dependency repo from `codegen`; migrate deprecated `-Xjvm-default=all-compatibility` → stabilized `-jvm-default=enable`.
Closes #20 — `@OptIn(ExperimentalXmlUtilApi::class)` on `Xml2Kotlin.run()` (the experimental xmlutil policy override is load-bearing) → 7 warnings gone.
Closes #21 — catalog-reference `kotlin-compiler-embeddable` (was hardcoded `2.3.0` vs the 2.4.0 toolchain) via a new `libs` entry pinned to the `kotlin` version ref.

## Verification
- [x] `codegen:compileKotlin` warning-free; JVM compile Beta-warning count: 0
- [x] `codegen:test` green (compilation test now uses embeddable 2.4.0)
- [x] `ktlintCheck` + `apiCheck` green (no API/ABI delta)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -49,6 +49,7 @@ kotlin {
         // JVM-21 bytecode and break every JVM-17 consumer.
         compilerOptions {
             jvmTarget.set(JvmTarget.JVM_17)
+            freeCompilerArgs.add("-Xexpect-actual-classes")
         }
     }
     linuxX64 {
--- a/codegen/build.gradle.kts
+++ b/codegen/build.gradle.kts
@@ -14,7 +14,6 @@ group = "com.monkopedia"
 
 repositories {
     mavenCentral()
-    maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
 }
 
 dependencies {
@@ -22,7 +21,7 @@ dependencies {
     api(libs.clikt)
     api(libs.kotlinpoet)
     testImplementation(project(":"))
-    testImplementation("org.jetbrains.kotlin:kotlin-compiler-embeddable:2.3.0")
+    testImplementation(libs.kotlin.compiler.embeddable)
     testImplementation(kotlin("test"))
     testImplementation("org.junit.jupiter:junit-jupiter-api:5.8.2")
     testImplementation("org.junit.jupiter:junit-jupiter-params:5.8.2")
@@ -55,7 +54,7 @@ tasks.test {
 kotlin {
     compilerOptions {
         jvmTarget.set(JvmTarget.JVM_1_8)
-        freeCompilerArgs.add("-Xjvm-default=all-compatibility")
+        freeCompilerArgs.add("-jvm-default=enable")
     }
 }
 java {
@@ -101,15 +100,6 @@ mavenPublishing {
                 "scm:git:ssh://github.com/Monkopedia/sdbus-kotlin.git"
             )
             url.set("https://github.com/Monkopedia/sdbus-kotlin/")
-        }
-    }
-    repositories {
-        maven(url = "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
-            name = "OSSRH"
-            credentials {
-                username = System.getenv("MAVEN_USERNAME")
-                password = System.getenv("MAVEN_PASSWORD")
-            }
         }
     }
     publishToMavenCentral(automaticRelease = true)
--- a/codegen/src/main/kotlin/Xml2Kotlin.kt
+++ b/codegen/src/main/kotlin/Xml2Kotlin.kt
@@ -80,6 +80,7 @@ import com.github.ajalt.clikt.parameters.types.file
 import java.io.File
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.decodeFromString
+import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi
 import nl.adaptivity.xmlutil.QName
 import nl.adaptivity.xmlutil.XmlReader
 import nl.adaptivity.xmlutil.serialization.DefaultXmlSerializationPolicy
@@ -107,6 +108,7 @@ class Xml2Kotlin : CliktCommand() {
         help = "Override Kotlin package for generated files"
     )
 
+    @OptIn(ExperimentalXmlUtilApi::class)
     override fun run() {
         val packageOverride = outputPackage?.trim()?.takeUnless { it.isEmpty() }
         val xml = XML {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -29,6 +29,7 @@ kotlinx-datetime = { module = "org.jetbrains.kotlinx:kotlinx-datetime", version.
 kotlinx-atomicfu = { module = "org.jetbrains.kotlinx:atomicfu", version.ref = "kotlin-atomicfu" }
 xmlutil = { module = "io.github.pdvrieze.xmlutil:serialization", version.ref = "pdvrieze-xmlutil" }
 kotlinpoet = { module = "com.squareup:kotlinpoet-jvm", version.ref = "kotlinpoet" }
+kotlin-compiler-embeddable = { module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin" }
 clikt = { module = "com.github.ajalt.clikt:clikt", version.ref = "clikt"}
 dbus-java-core = { module = "com.github.hypfvieh:dbus-java-core", version.ref = "dbus-java" }
 dbus-java-transport-junixsocket = { module = "com.github.hypfvieh:dbus-java-transport-junixsocket", version.ref = "dbus-java" }


monkopedia-reviewer (approved):
Reviewed all four mechanical cleanups. Each fix is correct and scoped exactly to what the issue describes:

**#17** — `-Xexpect-actual-classes` added to the jvm target's `compilerOptions.freeCompilerArgs` only. Matches the existing pattern on native targets.

**#19** — Dead OSSRH `repositories { maven(...) { credentials {...} } }` block removed; unused `oss.sonatype` snapshot dep repo removed. `publishToMavenCentral(automaticRelease = true)` is still present — publishing intact. `-jvm-default=enable` is the correct stable replacement for the deprecated `-Xjvm-default=all-compatibility` in Kotlin 2.4.0.

**#20** — `import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi` + `@OptIn(ExperimentalXmlUtilApi::class)` added on `run()`. Pure opt-in, zero logic change.

**#21** — `kotlin-compiler-embeddable` catalog entry added, pinned to `version.ref = "kotlin"` (resolves to 2.4.0). Hardcoded `2.3.0` replaced with `libs.kotlin.compiler.embeddable`. Version skew eliminated.

**CI:** all checks passed — `static-checks` (ktlintCheck + apiCheck: no API/ABI delta), `full-tests-x64` (codegen:test), all `pr-smoke` matrix jobs, `native-build-x64`, `arm-runtime-tests`. No public API/ABI change. Tier 1 — auto-merging.

PR #23, opened 2026-06-07
Drop vestigial 'inline' where it has no effect (#18)

### LINKED ISSUE #18: Drop vestigial 'inline' from trivial delegating functions (11 compiler warnings)
**Category:** idiom
**Where:** `src/commonMain/kotlin/com/monkopedia/sdbus/Proxy.kt` (lines 300, 303, 331, 353, 371, 390); `Error.kt:36`; `TypedMethod.kt:225, 720`; `src/jvmMain/.../Connection.jvm.kt:43`; `Message.jvm.kt:541`

**Current:** Each of these functions is marked `inline` but has no `reified` type parameter and no function-type parameter, so it just delegates / constructs an object, e.g.:
```kotlin
inline fun Proxy.getAllProperties(): AllPropertiesGetter = AllPropertiesGetter(this)
inline fun Proxy.callMethod(message: MethodCall, timeout: Duration): MethodReply =
    callMethod(message, timeout.inWholeMicroseconds.toULong())
internal inline fun sdbusRequire(condition: Boolean, msg: String, errNo: Int) { ... }
```
Kotlin 2.4.0 flags every one of these (confirmed via build):
```
w: ...Proxy.kt:371:1 Expected performance impact from inlining is insignificant. Inlining works best for functions with parameters of function types.
```
11 warnings total. Note: the sibling overloads that genuinely need it keep `inline` — e.g. `sdbusRequire(condition: () -> Boolean, ...)` (Error.kt:32, has a lambda param), and all the `reified`/`crossinline` builders in TypedMethod.kt. Only the no-benefit ones are listed above.

**Proposed:** Remove `inline` from the listed functions (leave the reified/lambda-bearing ones untouched).

**Why:** Eliminates 11 spurious warnings and clarifies intent — `inline` currently signals a performance choice that the compiler says is a no-op. **Caveat (worth a decision):** these are public top-level functions, so `inline` is part of the tracked ABI (it appears as `final inline fun` in `api/sdbus-kotlin.klib.api`). Removing it is a technically binary-incompatible change for already-compiled native consumers and will require regenerating the BCV `.api` dumps (`apiDump`). For a pre-1.0 library this is usually fine, but that's the call to make — hence `needs-decision`. (`Error.sdbusRequire` is `internal`, so it has no ABI impact.)

**Effort:** small refactor (plus an `apiDump` regen)

monkopedia-coder:
Closes #18. Removes the `inline` modifier from the 18 functions Kotlin 2.4.0 flags with "inlining is insignificant" (no reified / function-type params, so `inline` does nothing). No behavior change.

## Scope
- **Internal (no ABI impact):** the `now()` and `deserializeArrayFast()` `expect`/`actual` groups (expect + both actuals changed together), plus `ConnectionImpl`/`Error`/`TypedMethod`/`Utils` helpers.
- **Public (8 functions):** `Proxy.kt` (`callMethod`, `callMethodAsync`, `getPropertyAsync`, `setPropertyAsync`, `getAllProperties`, `getAllPropertiesAsync`) + `TypedArgumentsBuilderContext.call` / `TypedMethodBuilderContext.args`. `api/sdbus-kotlin.klib.api` is regenerated — **the only delta is `inline fun` → `fun`** (no signature added/removed/changed; see the diff).

## ABI safety (reviewed + approved with the repo owner)
This is the binary-**safe** direction (removing, not adding, `inline`): already-compiled consumers have the body inlined, and those bodies reference only **public** symbols that remain, so they keep working with no recompile; new consumers call the now-regular function. `callMethodAsync` also drops its now-pointless `noinline` on the (already non-inlinable) callback param. The owner reviewed this trade-off in-session and approved doing all sites.

## Verification
- [x] All targets compile **warning-free** (0 "inlining insignificant")
- [x] `jvmTest`, `linuxX64Test`, `:plugin:test`, `:codegen:test` green
- [x] `ktlintCheck` + `apiCheck` green (apiDump regenerated; delta is inline-modifier only)


monkopedia-reviewer (approved):
Reviewed to merge standard — clean and mergeable.

**apiDump diff:** Exactly inline-modifier removal only. The 8 public symbols that change (6 Proxy extension funs + TypedArgumentsBuilderContext.call + TypedMethodBuilderContext.args) move from `final inline fun` to `final fun` — no signature added, removed, or altered. The callMethodAsync callback drops its now-pointless `noinline` annotation, consistent with the function no longer being inline. No other ABI delta.

**expect/actual consistency:** `now()` and `deserializeArrayFast()` both have their `expect` and all `actual` declarations updated together — compilable and consistent.

**Source diff:** Purely keyword removal (`inline`/`noinline`); no logic change whatsoever. Internal helpers (sdbusRequire, checkObjectPath/InterfaceName/ServiceName/MemberName) have no ABI impact.

**Quality (axis A + B):** Correct — does exactly what it claims. Idiomatic — removal of vestigial modifiers is the right move when Kotlin itself flags them. Consistent with surrounding code patterns. No dead code introduced.

**CI:** All checks pass — static-checks (ktlintCheck + apiCheck), full-tests-x64, native-build-x64, arm-runtime-tests, and all 8 pr-smoke matrix jobs. nightly-stress skipped (expected).

**Tier note:** Classifies as public_api (tier 2 for sdbus-kotlin by default), but the human-judgment gate (ABI change decision) was pre-cleared in-session by the repo owner with explicit reasoning — removing inline is the binary-safe direction, no consumer break. Resolves as a merge.

PR #47, opened 2026-06-09
Fix #28: JVM PropertiesChanged emits current value in changed_properties

### LINKED ISSUE #28: PropertiesChanged emit parity — JVM sends empty changedProperties vs native populating values
**Decision + tests.** On the JVM backend, `emitPropertiesChangedSignal` sends an empty `changedProperties` map plus invalidated-only names, whereas the native backend populates `changedProperties` with the current values from the vtable getter. That's a real cross-backend divergence found during the blue-falcon-sdbus bring-up.

Decide whether to make JVM match native (read the getter and populate `changedProperties`) or to document the difference as intentional — then add cross-backend tests either way.

monkopedia-coder:
## What

Makes the JVM `PureJavaDbusBackend`'s `emitPropertiesChangedSignal(interfaceName, propNames)` match native behavior (Option A, per the owner's decision on #28).

Previously the JVM backend sent an **empty** `changed_properties` dict and placed every named property in `invalidated_properties`. Native's `sd_bus_emit_properties_changed_strv` instead reads each property's getter and carries the current value in `changed_properties`. The divergence meant a consumer's `PropertyGetter.changes()` never fired on the JVM backend — only `changesOrNull()` did.

## How

- Added `resolveChangedProperties(...)` which, for each named property, reads its registered vtable getter and puts the current value into `changed_properties`. Names with no readable getter (write-only / value unavailable) fall back to `invalidated_properties`, mirroring native.
- Both the local (no-connection) signal-bus path and the real dbus-java connection path now emit the populated `changed_properties` map. The real-connection path converts each `Variant` to a dbus-java `Variant` via the existing `extractVariantPayload` / `toJavaVariantValue` helpers.

This is a behavior-only change — the method signature is unchanged, so there is **no apiDump impact** (`apiCheck` passes with no `apiDump` update).

## Tests

Strengthened the cross-backend `propertiesChangedSignal_roundTripsForDeclaredProperty` test to assert the emitted signal carries the value in `changed_properties` (not `invalidated_properties`), which is what exercises `PropertyGetter.changes()` on both backends.

## Verification

- `./gradlew ktlintCheck apiCheck` — pass
- `:jvmTest` (full root suite, incl. the strengthened test) under `dbus-run-session` — pass

Fixes #28

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/CommonApiIntegrationTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/CommonApiIntegrationTest.kt
@@ -1068,12 +1068,17 @@ class CommonApiIntegrationTest {
         }
 
         try {
+            state = true
             obj.emitPropertiesChangedSignal(ids.iface, listOf(stateProperty))
 
             val (changedProperties, invalidatedProperties) = withTimeout(2_000) { seen.await() }
-            val propertySeen = changedProperties.containsKey(stateProperty) ||
-                invalidatedProperties.contains(stateProperty)
-            assertTrue(propertySeen)
+            // Matching native, the emitted signal carries the current value in
+            // changed_properties (read from the registered getter), not invalidated_properties.
+            // This is what makes a consumer's PropertyGetter.changes() fire, not just
+            // changesOrNull().
+            assertTrue(changedProperties.containsKey(stateProperty))
+            assertFalse(invalidatedProperties.contains(stateProperty))
+            assertTrue(changedProperties.getValue(stateProperty).get<Boolean>())
         } finally {
             signalRegistration.release()
             registration.release()
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/PureJavaDbusBackend.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/PureJavaDbusBackend.kt
@@ -1483,13 +1483,47 @@ private class PureJavaDbusObject(
         }.toMap()
     }
 
+    private fun readPropertyValue(property: PropertyVTableItem): Variant {
+        val getter = property.getter ?: error("no getter")
+        val reply = PropertyGetReply()
+        val value = JvmCurrentMessageContext.withMessage(reply) {
+            getter(reply)
+            reply.payload.firstOrNull()
+        }
+        return propertyToVariant(value)
+    }
+
+    // Mirrors native's sd_bus_emit_properties_changed_strv: each named property whose vtable
+    // getter can be read goes into changed_properties with its current value; names without a
+    // readable getter (write-only / value unavailable) fall back to invalidated_properties.
+    private fun resolveChangedProperties(
+        interfaceName: String,
+        propNames: List<PropertyName>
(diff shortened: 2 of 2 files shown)


monkopedia-coder:
## CI triage: both failures are non-regressions (native flake + Maven Central 403)

Investigated the two red checks. Neither is caused by this PR's change (which is JVM-only: `PureJavaDbusBackend.emitPropertiesChangedSignal` + a strengthened common test).

### `arm-runtime-tests` — known native parallel-async flake
Run `27207978810`. The whole `CommonApiIntegrationTest` suite passed on ARM, **including** the strengthened `propertiesChangedSignal_roundTripsForDeclaredProperty` ([ OK ], 1 ms). The job then aborted in the **native** runtime:

```
[ RUN ] DBusAsyncMethodsTest.handlesCorrectlyABulkOfParallelServerSideAsyncMethods
malloc(): unsorted double linked list corrupted
##[error]Process completed with exit code 134.
```

This is the documented flaky parallel server-side async stress test — native heap corruption / teardown timing under load, exit 134 (SIGABRT). This PR touches only `src/jvmMain/...` (the dbus-java backend); that code is **not compiled or executed** in the ARM native test run, so it cannot be the cause.

### `pr-smoke (reverse, true, control, timeout, native-client-timeout)` — Maven Central infra 403
Run `27207978568`. Build failed during dependency resolution, before any test ran:

```
Could not GET 'https://repo.maven.apache.org/maven2/org/jetbrains/kotlinx/atomicfu-jvm/0.32.1/atomicfu-jvm-0.32.1.pom'.
Received status code 403 from server: Forbidden
BUILD FAILED in 17s
```

A transient Maven Central 403 on the `atomicfu-jvm` POM — pure network/infra, unrelated to the diff.

### Local verification
On the PR head, `dbus-run-session -- ./gradlew :jvmTest --tests "...propertiesChangedSignal_roundTripsForDeclaredProperty"` passes, confirming the JVM behavior change (current value now carried in `changed_properties`) is sound.

### Verdict
Both failures are flake/infra, not regressions. **A re-run should clear both.** No code change warranted.

monkopedia-reviewer (approved):
Tier-1 auto-review (bug_fix, sdbus-kotlin defaults).

**Correctness** — `resolveChangedProperties` correctly mirrors native `sd_bus_emit_properties_changed_strv`: reads each named property's vtable getter via the existing `JvmCurrentMessageContext`/`PropertyGetReply` plumbing, puts the current value in `changed_properties`, and falls back to `invalidated_properties` for write-only / unreadable props (via `runCatching { }.getOrNull()`). Both the local-bus and real dbus-java connection paths are updated; the real-connection path correctly converts variants using the existing `extractVariantPayload`/`toJavaVariantValue` helpers. Leaving the no-arg `emitPropertiesChangedSignal(interfaceName)` unchanged is reasonable — native's emit-all depends on per-property `EmitsChangedSignal` flags not tracked in the JVM vtable.

**Test** — The strengthened test sets `state = true` before emitting, then asserts the value lands in `changed_properties` (not `invalidated_properties`) with the correct `Boolean` value. This directly exercises `PropertyGetter.changes()` on both backends.

**Quality** — Idiomatic Kotlin throughout; new helpers are small and focused; comment on `resolveChangedProperties` explains the native-mirror intent; naming and style match surrounding code; no style-guide violations.

**CI** — All checks green. Auto-merging.

PR #11, opened 2026-06-02
Fix JVM empty-collection arg signature + ay→UByte deserialization; pin jvmTarget=17

monkopedia-coder:
## Summary

Three JVM-backend fixes, found and validated while bringing up blue-falcon-sdbus' JVM target against a real BlueZ peripheral (BF-Test on adolin). Native is unaffected by all three.

1. **Empty-collection argument signature.** `PureJavaDbusBackend` derived each argument's D-Bus signature from its runtime *value*. An empty collection has nothing to infer from, so an empty `Map<String, Variant>` serialized as the malformed `a{}` and an empty `List` as the bare `a`. The bus daemon rejects that and drops the connection (`Underlying transport returned -1`). Now the declared signature is captured from the serializer descriptor at serialize time (`Message.declaredBodySignature`) and used for the wire body, falling back to value-inference only when no declared signature is available. This is what broke `GATT ReadValue(a{sv} options)` (empty options dict) on JVM.

2. **`ay` → `UByte` deserialization.** `coerceForDescriptor` boxed `ObjectPath`/`UnixFd`/etc. and recursed into `LIST`/`MAP`, but left the unsigned inline-class primitives as the raw signed value dbus-java returns. A `ay` (e.g. a GATT characteristic value) deserialized into a `List` holding `java.lang.Byte` where `List<UByte>` was expected, throwing `class java.lang.Byte cannot be cast to class kotlin.UByte`. Now signed primitives are boxed into the declared unsigned type (`UByte`/`UShort`/`UInt`/`ULong`).

3. **Pin `jvmTarget = 17`.** The `jvm()` target set no `jvmTarget`, so published bytecode followed the build JDK — a JDK-21 build emits JVM-21 bytecode that JVM-17 consumers can't inline. Pinned to 17 for deterministic releases. (Also gates `signAllPublications()` off for `-SNAPSHOT` so local snapshot publishing works without a GPG key.)

## Test plan

- [x] New `commonTest` round-trips for empty `a{sv}` and empty `as` method args
- [x] New `jvmTest` real-wire regression (`BecomeMonitor` with empty `as`) — fails without fix #1
- [x] New `jvmTest` `ay`→`UByte` deserialization regression — fails without fix #2
- [x] Full `jvmTest` + `linuxX64Test` green; `apiCheck` clean (all changes internal, no public API delta)
- [x] Verified a JDK-21 build now emits JVM-17 bytecode
[...]

PR #51, opened 2026-06-09
Add cross-backend D-Bus type-matrix round-trip tests

### LINKED ISSUE #25: Tests: cross-backend D-Bus type-matrix round-trips (jvm + native)
Systematically round-trip every D-Bus type through serialize → wire → deserialize on **both** backends (native sd-bus + JVM dbus-java), asserting identical results across them.

Cover: basic types (`y b n q i u x t d s o g h`), arrays incl. **empty**, dicts (`a{sv}`, `a{ss}`, deeply nested `a{oa{sa{sv}}}`), structs, variants, and nested compositions. Run both as in-process round-trips (commonTest) **and** over a real bus (the local path short-circuits, so same-process tests miss wire-serialization bugs — see the BecomeMonitor approach).

**Why this is the highest-value test for 1.0:** the empty-`a{sv}` signature corruption (#11) and the `ay`→UByte ClassCastException were JVM-only marshalling gaps that a systematic cross-backend matrix would have caught proactively, before a consumer hit them.

monkopedia-coder:
Fixes #25

Adds `TypeMatrixRoundTripTest` in `src/commonTest/kotlin/com/monkopedia/sdbus/` so the exact same assertions run against **both** backends: the native sd-bus backend (`linuxX64Test`) and the JVM dbus-java backend (`jvmTest`). Any marshalling divergence between the two surfaces as a failure on one of them.

## Approach

Each test stands up a real bus connection with a server object exposing an identity `Echo` method (declared to take and return the type under test), calls it through a proxy, and asserts the returned value equals the value sent. Values cross a **real bus** rather than short-circuiting in-process, so wire-serialization bugs (e.g. the empty-`a{sv}` corruption #11, or the `ay`->UByte ClassCastException) are exercised against the actual daemon round trip. The connection/object/proxy setup follows the existing `CommonApiIntegrationTest` pattern; the file is new and standalone to avoid conflicts with in-flight test PRs (#28).

## Coverage

- Basic types: `y b n q i u x t d s o g v`, including min/max extremes for the numerics and empty strings
- Arrays: `ai as ay ad`, empty arrays, arrays of arrays
- Dicts: `a{ss} a{ix} a{sv}`, empty dicts, and a deeply nested `a{oa{sa{sv}}}`
- Structs: `(is)`, a wide all-basic-types struct, nested structs
- Variants: wrapping scalars, arrays, and dicts
- Nested compositions: arrays of structs, dicts of arrays

## Verification

Run under `dbus-run-session` mirroring the CI full-tests job:
- `:jvmTest` — pass (34 cases)
- `:linuxX64Test` — pass (34 cases)
- `ktlintCheck` — pass
- `apiCheck` — pass (test-only change, public API unaffected)

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/TypeMatrixRoundTripTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/TypeMatrixRoundTripTest.kt
@@ -1,0 +1,354 @@
+package com.monkopedia.sdbus
+
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.serialization.Serializable
+
+/**
+ * Cross-backend D-Bus type-matrix round-trip tests (issue #25).
+ *
+ * Each test stands up a real bus connection with a server object exposing an `Echo` method that
+ * returns its argument unchanged, then calls it through a proxy and asserts the returned value
+ * equals the value sent. Because this lives in commonTest, the exact same assertions run against
+ * BOTH the native sd-bus backend (linuxX64Test) and the JVM dbus-java backend (jvmTest), so any
+ * marshalling divergence between the two backends shows up as a test failure on one of them.
+ *
+ * The values cross a real bus rather than short-circuiting in-process, so wire-serialization bugs
+ * (e.g. the empty-`a{sv}` signature corruption #11, or the `ay`->UByte ClassCastException) are
+ * exercised against the actual daemon round trip.
+ *
+ * The matrix covers the basic D-Bus types y b n q i u x t d s o g v, plus arrays (incl. empty),
+ * dicts `a{..}` (incl. empty and deeply nested), structs, variants, and nested compositions.
+ */
+class TypeMatrixRoundTripTest {
+
+    @Serializable
+    private data class SimpleStruct(val i: Int, val s: String)
+
+    @Serializable
+    private data class WideStruct(
+        val y: UByte,
+        val b: Boolean,
+        val n: Short,
+        val q: UShort,
+        val i: Int,
+        val u: UInt,
+        val x: Long,
+        val t: ULong,
+        val d: Double,
+        val s: String,
+        val o: ObjectPath,
+        val g: Signature,
+        val v: Variant
+    )
+
+    @Serializable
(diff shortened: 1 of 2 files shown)


monkopedia-coder:
Fixed the failing `arm-runtime-tests` check by relocating `TypeMatrixRoundTripTest` from package `com.monkopedia.sdbus` into `com.monkopedia.sdbus.integration` (and the matching directory), adding the required imports for the public API it references.

The ARM CI job runs tests in `com.monkopedia.sdbus.integration.*` inside `dbus-run-session` (with a private bus) and everything else bare. Because these tests stand up a real D-Bus connection, running them bare on ARM failed with `org.freedesktop.DBus.Error.AccessDenied: Failed to request bus name (Permission denied)`. Moving the class into the integration package routes it through `dbus-run-session` like the other bus-using tests.

Verified locally under `dbus-run-session`: `:jvmTest :linuxX64Test --tests "*TypeMatrixRoundTripTest*"` passes on both backends, and `ktlintCheck` is clean.

monkopedia-reviewer (approved):
Clean test-only PR. Full review across both axes:

**Axis A — Correctness:**
All 13 D-Bus basic types covered (y b n q i u x t d s o g v) with min/max extremes where applicable. Arrays (incl. empty, array-of-arrays), dicts (incl. empty, and the deeply nested `a{oa{sa{sv}}}` from issue #25), structs (simple, all-basic-types wide, nested), and variant-wrapping-containers all exercised. `roundTrip` creates real bus connections and routes through a real server object — genuine wire serialization, no in-process short-circuit. Variant fields correctly compared by inner-value extraction (not `Variant.equals`, which has no value semantics); the comment explains why. Resource cleanup in the `finally` block is complete. Package `com.monkopedia.sdbus.integration` correctly routes ARM CI through `dbus-run-session`. All CI checks green.

**Axis B — Quality:**
Mirrors `CommonApiIntegrationTest` patterns exactly: same `FixtureIds`/`uniqueFixtureIds` helper, same connection setup, same `createProxy(dontRunEventLoopThread = true)` call. Idiomatic Kotlin (inline reified generics, data classes, standard collections). Section comments and descriptive test names (`type_dbusType_roundTrips`) aid readability. The `roundTrip`/`assertRoundTrips` helper pair reduces boilerplate cleanly.

PR #54, opened 2026-06-09
Add cross-backend failure-path parity tests

### LINKED ISSUE #29: Tests: error/failure-path consistency across backends (timeouts, errors, malformed, teardown)
Assert consistent behaviour and error surfacing across both backends for the failure paths: method-call timeouts, method-error replies, invalid/malformed signatures, and transport teardown.

The transport-teardown-on-malformed-message behaviour (the original blue-falcon-sdbus failure mode) had no test coverage when it shipped; this closes that gap and pins the error model before 1.0.

monkopedia-coder:
## What

Adds `FailurePathParityTest` in `src/commonTest/kotlin/com/monkopedia/sdbus/` — a fresh, distinctly-named file (no shared test files touched). Because it lives in commonTest it compiles and runs against every target, so the **native sd-bus backend** (`linux*Test`) and the **pure-JVM backend** (`jvmTest`) both execute the same assertions; a backend divergence shows up as a per-target failure.

## Coverage (the failure paths from the issue)

- **method-call timeout** — a slow server method past the client's per-call timeout surfaces a timeout-shaped `Error` (`Timeout` / `NoReply` / "timed out"), never a hang or unrelated exception.
- **remote named D-Bus error** — a handler that throws `Error(name, message)` reaches the client with the **same name and message** on both backends. Pins the error model.
- **nonexistent method** — surfaces an `Error` carrying a name.
- **wrong-argument-type / signature mismatch** — calling an `Int` method with a `String` is rejected as an error, not delivered to the handler.
- **connection teardown** — `leaveEventLoop` + `release` completes cleanly and **time-bounded**; a call to a torn-down server surfaces an `Error`.

## Hang safety

Every teardown/timeout assertion is wrapped in `withTimeout`, so a hang in the known native-teardown path surfaces as a deterministic test failure rather than an indefinitely stuck suite. The slow-handler timeout case keeps both event loops alive until the late reply flushes — otherwise the native backend throws an uncaught `ENOTCONN` ("Failed to send reply") when a handler completes into an already-torn-down loop and crashes the test process.

## Discovered backend divergences (intentionally not asserted as parity)

While building these I found two places where the backends do **not** behave consistently today, so I scoped the tests around them rather than shipping red assertions:

1. **Lenient reply deserialization (JVM).** Deserializing an `Int` reply (`i`) into a struct type like `(ii)` throws on the native backend but the pure-JVM backend returns a value instead of erroring. I dropped that sub-case; the wrong-argument-type mismatch (which *is* consistent) covers the signature-mismatch path.
2. **Call-after-release (JVM).** After `leaveEventLoop` + `release` of a proxy connection, a subsequent call throws on native but is still serviced on the pure-JVM backend. So `proxyConnectionTeardown_completesCleanly` asserts only the robust parity property — teardown completes cleanly and time-bounded — and a code comment documents why the "call-after-release fails" assertion is omitted.

These look like real JVM-backend gaps worth a follow-up; flagging rather than fixing here since this PR is additive test-only.

## Verification

- `ktlintCommonTestSourceSetCheck` — pass.
- Under `dbus-run-session` (mirroring the full-tests-x64 CI job): `:jvmTest` and `:linuxX64Test` filtered to `FailurePathParityTest` — all pass, stable across repeated runs.
- Test-only; `apiCheck` unaffected.

Fixes #29

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/FailurePathParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/FailurePathParityTest.kt
@@ -1,0 +1,314 @@
+package com.monkopedia.sdbus
+
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+import kotlin.time.Duration.Companion.milliseconds
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+
+/**
+ * Error / failure-path consistency tests. These assert that BOTH backends (the native sd-bus
+ * backend exercised by the linux*Test targets and the pure-JVM backend exercised by jvmTest)
+ * surface failures the same way: a timeout becomes a timeout [Error], a server-thrown named
+ * D-Bus error round-trips its name + message, a wrong-argument-type / signature mismatch becomes
+ * an [Error] rather than reaching the handler, and tearing a connection down stays consistent
+ * without crashing or hanging.
+ *
+ * Because the suite is in commonTest it is compiled and run against every target, so a backend
+ * divergence shows up as a per-target test failure.
+ *
+ * NOTE on teardown: connection teardown / event-loop shutdown is the project's known native-hang
+ * area. Every teardown assertion here is wrapped in [withTimeout] so that a hang surfaces as a
+ * deterministic test failure instead of an indefinitely stuck suite.
+ */
+class FailurePathParityTest {
+
+    private data class FixtureIds(
+        val service: ServiceName,
+        val path: ObjectPath,
+        val iface: InterfaceName
+    )
+
+    private fun uniqueFixtureIds(suffix: String): FixtureIds {
+        val id = Random.nextInt(100_000, 999_999)
+        val base = "com.monkopedia.sdbus.failure.$suffix$id"
+        return FixtureIds(
+            service = ServiceName(base),
+            path = ObjectPath("/com/monkopedia/sdbus/failure/$suffix$id"),
+            iface = InterfaceName("$base.Interface")
+        )
+    }
+
+    // --- method-call timeout ---------------------------------------------------------------
+
+    // A server method that deliberately sleeps far past the client's per-call timeout must
+    // surface as a timeout Error on both backends -- not a hang, and not some other error type.
+    @Test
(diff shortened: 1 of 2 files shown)


monkopedia-coder:
Fixed the failing `arm-runtime-tests` check by moving `FailurePathParityTest` into package `com.monkopedia.sdbus.integration` (both the `package` declaration and the file path under `src/commonTest/kotlin/com/monkopedia/sdbus/integration/`), adding explicit imports for the sdbus symbols it now references across the package boundary.

The ARM job runs tests in two batches: `com.monkopedia.sdbus.integration.*` inside `dbus-run-session` (private session bus), everything else bare. Since this suite stands up real D-Bus connections, running it bare on ARM failed with `org.freedesktop.DBus.Error.AccessDenied: Failed to request bus name (Permission denied)`. Relocating it into the `.integration` package puts it in the dbus-run-session batch alongside the other bus-using integration tests.

Verified locally under `dbus-run-session`: `:jvmTest :linuxX64Test --tests "*FailurePathParityTest*"` passes on both backends, and `ktlintCheck` is clean.

monkopedia-reviewer (approved):
tier-1 auto-review: test_only, CI green, no overrides for sdbus-kotlin.

**Correctness (A):**
- Package placement correct: `com.monkopedia.sdbus.integration` ensures ARM CI runs it inside `dbus-run-session` with a bus available.
- All teardown/timeout assertions are wrapped in `withTimeout` — hangs surface as deterministic failures, not suite stalls.
- The timeout test keeps both event loops alive for `serverDelayMs + 400ms` to let the late reply flush before teardown, correctly avoiding the native ENOTCONN crash.
- Named-error test asserts exact name + message equality — pins the round-trip contract.
- `assertFailsWith<Throwable>` for the wrong-arg case is intentionally broad (documented scope-down: backends diverge on the exact type here); the assertion still guarantees the call is rejected, not silently delivered.
- Both scope-downs (lenient reply deserialization, call-after-release JVM leniency) are commented in code and documented in PR body as filed follow-ups (#56). Not shipped as red assertions — correct.

**Quality (B):**
- `uniqueFixtureIds` + random suffix prevents cross-test service-name collisions; consistent with `CommonApiIntegrationTest` pattern.
- Cleanup order (release without leaveEventLoop in non-teardown tests) matches the established pattern in `releasedVTable_rejectsFurtherMethodCalls`.
- Naming, structured concurrency, and comment density match surrounding integration tests.

CI: arm-runtime-tests, full-tests-x64, native-build-x64, all pr-smoke variants, static-checks — all pass.

PR #55, opened 2026-06-09
Tests: exhaustive empty-collection wire coverage; fix empty signal-body signature

### LINKED ISSUE #26: Tests: exhaustive empty-collection argument coverage over the wire
Exhaustive coverage of **empty** collection arguments — empty `as`, `a{sv}`, `ai`, `ay`, and empty structs — as both method-call arguments and signal payloads, over the **real wire**, on both backends.

#11 added empty `a{sv}`/`as` cases; make it complete. Guards the malformed-signature class of bug (empty collection → inferred signature `a{}`/`a` → daemon drops the connection).

monkopedia-coder:
## Summary

Exhaustive over-the-wire coverage of **empty** collection arguments on both backends, extending the two cases added in #11. New `EmptyCollectionWireTest` (commonTest, 23 tests) exercises, over a real session-bus connection:

- **Empty arrays** of each element type: `as`, `ai`, `ax`, `ad`, `ab`, `ay` (UByte), `au` (UInt), `ao` (ObjectPath), `av` (Variant)
- **Empty dicts** of assorted key/value types: `a{sv}`, `a{si}`, `a{is}`, `a{sas}`, `a{ov}`
- **Nested empties**: empty array-of-arrays (`aas`), empty `as` inside a populated struct, empty `a{sv}` inside a variant
- **Empty collections as signal bodies**, asserting both round-trip-to-empty and (where the backend preserves it) the observed wire signature

Each case is a method-call arg or signal body that round-trips through a real proxy/object pair, so the native (sd-bus) and JVM (pure-Java dbus) backends are each covered by their test target.

## Library fix (the bug this guards)

While writing the signal-body cases I hit a live instance of the `inferSignalSignature` empty-collection bug class. `PureJavaDbusObject.emitSignal` built the wire signature by inferring from the runtime values, so an empty array emitted the malformed bare `a` (dbus-java threw `Index 1 out of bounds for length 1`) and an empty dict the malformed `a{}`. This is the same class of bug #11 fixed on the method-call path, but the signal-emit path was still inferring.

The fix reuses the existing `bodySignature()` helper — which already prefers the declared signature captured at serialize time (`Message.declaredBodySignature`) and only falls back to value inference when no declared signature is available. I promoted `bodySignature` from a `PureJavaDbusProxy` member to a top-level private function so `PureJavaDbusObject.emitSignal` can share the exact same logic as the method-call path. No behavior change for the method-call path.

The JVM backend reconstructs received bodies as plain Kotlin values with no retained per-element type, so peeking a *received* empty collection can only report the bare container. The wire-signature assertion is therefore gated on a new `EmptyCollectionWireSupport` expect/actual capability flag (true on native, false on JVM), mirroring the existing `FdTestSupport` pattern. The round-trip-to-empty assertion runs unconditionally on both backends.

## Files

New test files only, plus the one `emitSignal` fix:
- `src/commonTest/.../integration/EmptyCollectionWireTest.kt` (new)
- `src/commonTest/.../integration/EmptyCollectionWireSupport.kt` + jvm/native actuals (new)
- `src/jvmMain/.../PureJavaDbusBackend.kt` — `bodySignature` promoted to top-level; `emitSignal` uses it

No shared existing test files were edited (avoids conflicts with the in-flight sibling test PRs).

## Test plan

- [x] `:jvmTest` full suite green under `dbus-run-session`
- [x] `:linuxX64Test` full suite green under `dbus-run-session` (validates the wire-signature assertions on the backend that preserves them)
- [x] `EmptyCollectionWireTest`: 23 tests, 0 failures on both backends
- [x] `ktlintCheck` clean
- [x] `apiCheck` clean (no public API change)

Fixes #26

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/EmptyCollectionWireSupport.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/EmptyCollectionWireSupport.kt
@@ -1,0 +1,16 @@
+package com.monkopedia.sdbus.integration
+
+/**
+ * Per-backend capabilities for [EmptyCollectionWireTest].
+ *
+ * The native (sd-bus) backend keeps the real wire signature on a received message, so an empty
+ * collection can be peeked back as e.g. `as` / `a{sv}`. The JVM (pure-Java dbus) backend
+ * reconstructs the body as plain Kotlin values with no retained per-element type, so peeking an
+ * empty collection can only ever report the bare container (`a` / `a{}`). The send-side
+ * declared-signature fix (so the daemon does not drop the connection) and the round-trip back to
+ * an empty collection are exercised on both backends regardless.
+ */
+internal expect object EmptyCollectionWireSupport {
+    /** True when peeking a received empty collection reports its full element signature. */
+    val preservesWireSignatureForEmptyCollections: Boolean
+}
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/EmptyCollectionWireTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/EmptyCollectionWireTest.kt
@@ -1,0 +1,463 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.Message
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.Variant
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.callMethod
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.deserialize
+import com.monkopedia.sdbus.emitSignal
+import com.monkopedia.sdbus.method
+import com.monkopedia.sdbus.signal
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard on both axes. Clean approve.

**Correctness (A):**

The fix is sound and precisely mirrors the existing method-call path:

- `PureJavaDbusObject.emitSignal` was the sole remaining site that called `inferSignalSignature` directly on each payload element. For an empty `List`/`Map`, that produces the malformed `"a"` / `"a{}"` wire signatures that dbus-java rejects with an index OOB and the daemon drops the connection.
- The fix uses `bodySignature(message.payload.toList(), message.declaredBodySignature.toString()) { ... }` — the identical call shape already used on both call sites in `callRemoteMethod` (lines 916, 1034 in the pre-merge file). `declaredBodySignature` is a non-null `StringBuilder` accumulated by the serializer; `.toString()` returning `""` is handled cleanly by `isNullOrEmpty()` inside `bodySignature`.
- Promoting `bodySignature` from a private member of `PureJavaDbusProxy` to a top-level private function is the minimum necessary refactor; the function body is unchanged.
- The empty-signal-body test cases are non-tautological: a malformed send signature drops the bus connection, which means the `decoded.await()` would never complete (2s timeout expiry / connection-drop exception surface as a test failure, not a green pass).
- `EmptyCollectionWireSupport` expect/actual mirrors the established `FdTestSupport` pattern exactly.
- No public API change; `apiCheck` confirms.

**Quality (B):**

- Test structure: `roundTripUnaryMethod` / `roundTripUnarySignal` helpers cleanly cover 23 cases without boilerplate; the `crossinline` + `reified` approach is idiomatic.
- `finally`-block cleanup in `roundTripUnarySignal` correctly uses `suspend leaveEventLoop()` within the enclosing `runBlocking { }` context; `roundTripUnaryMethod` (regular function) uses nested `runBlocking { }` — both are correct.
- Section KDoc comments explain the intent, matching the comment density of the existing integration test files.
- No style-guide violations.

CI: all required checks green (arm-runtime-tests, full-tests-x64, native-build-x64, pr-smoke ×10, static-checks).

Issue #56, opened 2026-06-09
JVM backend more lenient than native on two failure paths (deserialization + call-after-release)


monkopedia-coder:
Surfaced by the failure-path parity tests in #29 (PR #54). The pure-JVM backend (`PureJavaDbusBackend`) is more permissive than the native sd-bus backend on two failure paths. For 1.0 we want consistent cross-backend behavior (same principle as the #28 "make JVM match native" decision), so these need a direction call.

## Divergence 1 — lenient reply deserialization on JVM
Deserializing a reply whose actual signature is `i` (a single Int) into a struct type `(ii)`:
- **native**: throws (signature mismatch)
- **JVM**: returns a value (does not reject the mismatch)

The #29 test dropped this sub-case rather than ship a red test. Question: tighten the JVM deserializer to reject signature mismatches like native? (Likely yes — silent acceptance can hand consumers garbage — but it's a behavior change with possible consumer impact.)

## Divergence 2 — call-after-release serviced on JVM
After `leaveEventLoop` + `release`, a subsequent proxy call:
- **native**: throws
- **JVM**: is still serviced

The #29 teardown test was scoped down to the robust parity property (clean, time-bounded teardown) and this divergence documented in a code comment. Question: should a call after release be a hard error on JVM too?

## Recommendation
Lean toward tightening JVM to match native on both (native is the reference, consistent with #28). Both are behavior-only changes — confirm no public API/ABI impact. Each could be its own fix PR once the direction is confirmed; splitting this ticket in two is fine if preferred.

Refs: #29, PR #54.

monkopedia-coder:
**Decision (owner, 2026-06-10): tighten JVM to match native on both paths.**

Native is the reference implementation — same principle as #28 (PR #47).

1. **Strict reply deserialization**: the JVM deserializer rejects signature mismatches (e.g. actual `i` into expected `(ii)`) like native does. Silent acceptance hands consumers garbage that looks like data — the worst failure mode.
2. **Call-after-release throws**: after `leaveEventLoop` + `release`, a proxy call errors on JVM exactly as on native. A use-after-release that "works" on one backend is a portability trap.

Each lands as its own fix PR, restoring the #29 test sub-cases that were dropped/scoped down (TypeMatrix struct-mismatch case; FailurePathParity call-after-release case) as the proof.

[...]

Issue #39, opened 2026-06-09
1.0 ABI: async reply callback is a bare Function2<MethodReply, Error?, Unit>


monkopedia-coder:
Child of #30 (1.0 public-API review).

## Symbols
- `Proxy.callMethodAsync(MethodCall, kotlin.Function2<MethodReply, Error?, Unit>): PendingAsyncCall`
- `Proxy.callMethodAsync(MethodCall, kotlin.Function2<MethodReply, Error?, Unit>, ULong): PendingAsyncCall`
- the `Duration` extension `Proxy.callMethodAsync(MethodCall, noinline Function2<MethodReply, Error?, Unit>, Duration)`

## What's risky
The async-reply handler is an anonymous `Function2<MethodReply, Error?, Unit>`. This:
- Locks an **unnamed structural type** into the ABI — the parameter has no name/KDoc anchor; tooling and call sites read poorly (`{ reply, error -> }` with no hint which is which, and `error` being nullable-on-success is non-obvious).
- Couples the success/error channel into one positional lambda rather than a named contract.
- Is awkward from Java (raw `Function2`).

## Options
1. **Lock as-is.** Cheapest; relies on docs to explain the `(reply, error?)` convention.
2. **Introduce a named `fun interface` before 1.0 (recommended).** e.g. `fun interface AsyncReplyHandler { fun onReply(reply: MethodReply, error: Error?) }` (SAM-convertible, so `{ reply, error -> }` still works) — names the ABI type, documents the contract, improves Java interop.
3. **Discuss / fold into #38.** If the callback async surface is demoted to internal (see #38), this becomes moot and no named type needs to be public.

## Recommendation
Resolve jointly with #38. If any callback async API stays public, do Option 2 (named `fun interface`). If #38 demotes callbacks to internal, close this as subsumed.

Links back to #30.

monkopedia-coder:
**Decision (owner, 2026-06-10): subsumed by #38.**

#38 resolved to suspend-only: the callback `callMethodAsync` overloads and `PendingAsyncCall` leave the public surface, so there is no public callback type left to name and no `fun interface` is needed. Closing as subsumed; the internal implementation may keep whatever shape is convenient.

PR #57, opened 2026-06-10
1.0 ABI: demote callback async, sig machinery, cinterop send, createError; de-data VTable builders

### LINKED ISSUE #38: 1.0 ABI: two parallel async call APIs (callback PendingAsyncCall vs suspend MethodReply)

## Symbols
On `Proxy`:
- Callback style: `callMethodAsync(MethodCall, Function2<MethodReply, Error?, Unit>): PendingAsyncCall` (+ `ULong` overload) and the `Duration` extension.
- Suspend style: `suspend callMethodAsync(MethodCall): MethodReply` (+ `ULong` overload) and the `Duration` extension.
- Higher-level suspend extension: `suspend Proxy.callMethodAsync(InterfaceName, MemberName, builder): A`.

Plus `PendingAsyncCall : Resource` (`isPending()`, `release()`).

## What's risky
Two fully-public, parallel ways to do the same asynchronous call. The callback variant returns `PendingAsyncCall` (a cancellable handle) and uses a hand-rolled `(reply, error?) -> Unit` callback; the suspend variant is the idiomatic coroutine form. Committing both to the 1.0 ABI doubles the surface we must support indefinitely and forces every future async feature to be mirrored in both. The callback form is also the one carrying the bare `Function2` (see sibling ticket) and is largely a structured-concurrency anti-pattern in idiomatic Kotlin.

## Options
1. **Commit to both.** Keep callback + suspend; document when to use which (callback only when you need the `PendingAsyncCall` cancel handle without a coroutine scope).
2. **Make the callback surface internal/removable before 1.0 (recommended).** Keep `suspend` as the public async contract; demote the callback `callMethodAsync` + `PendingAsyncCall` to internal (the suspend impl can use it under the hood). Cancellation is then expressed through coroutine cancellation.
3. **Discuss.** If callers genuinely need fire-and-forget with an external cancel handle, keep a single, clearly-named callback entry (not an overload collision with the suspend one) rather than the current mirrored pair.

## Recommendation
Option 2 unless there's a concrete consumer that needs the non-coroutine cancel handle. The suspend form should be the blessed 1.0 API.

Links back to #30.

### LINKED ISSUE #40: 1.0 ABI: Error is the single error contract (final class Error : Exception(name, message))

## Symbol
```
final class com.monkopedia.sdbus/Error : kotlin/Exception {
    constructor(name: String, message: String = ...)
    val errorMessage: String   // note: NOT the Throwable.message
    val name: String
}
```
Also related: `createError(Int, String): Error` (factory taking an int + message), and `Error?` appearing in the async callback (#39) and `MethodCall.createErrorReply(Error)`.

## What's risky
This is the entire error contract for the library, so it's worth being deliberate rather than defaulting:
- It exposes **two** message-ish fields: the inherited `Throwable.message` and a separate `errorMessage` property. Ambiguity about which is authoritative; `name`+`errorMessage` is the D-Bus (name, message) pair but the `Exception(message=…)` base also carries one. Worth confirming `message` == `errorMessage` and documenting, or collapsing.
- `createError(Int, String)` maps an int (errno? sd_bus negative code?) to an `Error`, but the `Error` class itself has no numeric/code field — the int is lost. If error codes matter, the class can't represent them after 1.0 without an ABI break.
- `final` with a single `(name, message)` constructor: fine if intentional, but means consumers can't subclass for typed/category errors, and we can't add a `cause`/`code` later without a new constructor (additive, but the shape should be decided now).

## Options
1. **Lock as-is**, but make it a deliberate decision: document `name`+`errorMessage` as *the* contract, confirm/align `message` vs `errorMessage`, and accept that numeric codes aren't represented.
2. **Tighten before 1.0 (recommended):** decide whether a D-Bus error *code* belongs on `Error` (add it now, since `createError` already has one and it's currently dropped); reconcile `message`/`errorMessage` to one obvious accessor; confirm `final` + name-or-name-and-message is the intended permanent shape.
3. **Discuss** whether a small sealed/category hierarchy is warranted (e.g. timeout vs remote vs marshalling) — probably overkill, but should be ruled on, not defaulted.

## Recommendation
Option 2, minimally: reconcile `message`/`errorMessage`, and decide the code question now because `createError(Int,…)` currently discards the int.

Links back to #30.

### LINKED ISSUE #41: 1.0 ABI: signature/serialization machinery (SdbusSig hierarchy, Typed, *Sig vals) is fully public

## Symbols
A large internal-looking marshalling layer is public:
- `sealed class SdbusSig` (`value`, `valid`, `isTrivial`) and its subclasses: `PrimitiveSig<A, B: CVariable>`, `ListSig`, `MapSig`, `StructSig`, `object InvalidSig`.
- `PrimitiveSig`'s `copy(... NativeTypeConverter<A,B>? ...)` — leaks `com.monkopedia.sdbus.internal.NativeTypeConverter` (an `internal`-named type) and `kotlinx.cinterop.CVariable` into a public signature.
- The full set of top-level `*Sig` vals: `BoolSig, DoubleSig, IntSig, LongSig, ObjectPathSig, ShortSig, SignatureSig, StringSig, UByteSig, UIntSig, ULongSig, UShortSig, UnixFdSig, VariantSig, VoidSig`.
- `class Typed<A>` + `typed<A>()`, `signatureOf(KType)` / `signatureOf<A>()`, `val SerialDescriptor.asSignature`.
- `MapSig` exposes `contents`, `dictValue`, `t1`, `t2` — internal field names baked into ABI.

## What's risky
This is the type-signature engine that the codegen and the inline `call/args/typed` helpers sit on top of. Most users should never touch `SdbusSig`/`PrimitiveSig`/`*Sig` directly — but it's all public, so 1.0 freezes:
- a `kotlinx.cinterop.CVariable` type parameter bound (native-only concept) in a common-API class,
- a reference to `…internal.NativeTypeConverter` in a public `copy()` signature (an internal type surfacing through a public API is a smell and an ABI hazard),
- internal-looking property names (`t1`, `t2`, `dictValue`, `contents`).
The inline `reified` helpers (`typed<A>()`, `signatureOf<A>()`, `call(...)`) legitimately need *something* public to bind to, so this can't all just go internal — but the raw class hierarchy and the `NativeTypeConverter`/`CVariable` leakage almost certainly shouldn't be frozen.

## Options
1. **Lock as-is.** Accept the whole engine as public 1.0 surface.
2. **Minimize before 1.0 (recommended).** Keep only what the public `reified` helpers genuinely require; make `SdbusSig` subclasses, `PrimitiveSig`/`NativeTypeConverter`/`CVariable` exposure, and the internal-named fields `internal` or `@PublishedApi internal`. Decide which `*Sig` vals (if any) are part of the intended user API vs codegen-only.
3. **Discuss** the boundary: what does an advanced user (custom serializer) legitimately need, and expose exactly that under a deliberately-named entry point.

## Recommendation
Option 2. At minimum, get `NativeTypeConverter` and `CVariable` out of any public signature, and reconsider whether the `SdbusSig` subclass hierarchy needs to be public at all.

Links back to #30.

### LINKED ISSUE #42: 1.0 ABI: raw cinterop types (sd_bus_message, sd_bus_error, CPointer/CFunction) leaked via MethodCall.send

## Symbol
```
MethodCall.send(
    CPointer<CFunction<Function3<
        CPointer<cnames.structs.sd_bus_message>?,
        CPointer<out CPointed>?,
        CPointer<sdbus.sd_bus_error>?,
        Int>>>,
    Any?,
    ULong
): Resource
```
(the `MethodCall.send(ULong): MethodReply` overload is covered by the timeout ticket #37.)

## What's risky
This overload exposes raw libsystemd cinterop types — `cnames.structs.sd_bus_message`, `sdbus.sd_bus_error`, `kotlinx.cinterop.CPointer/CFunction/CPointed` — directly in the **public** ABI. These are native-only, generated-binding types; they cannot exist on the JVM target (so this is a native-only public method on a common-looking class) and they hard-couple the public surface to the exact shape of the sd-bus C bindings. Freezing this into 1.0 means any libsystemd binding change or any cross-platform expectation breaks the ABI.

## Options
1. **Lock as-is** (accept a native-only, cinterop-typed public method).
2. **Make it internal before 1.0 (recommended).** This is plumbing for the async-send implementation; callers use `callMethodAsync`. Mark `internal` / `@PublishedApi internal`.
3. **Discuss** if any consumer legitimately drives a raw sd_bus callback send — if so, wrap it behind a Kotlin-typed signature, not raw `CPointer<CFunction<…>>`.

## Recommendation
Option 2 — no public API should carry raw `sd_bus_*` cinterop pointers into 1.0.

Links back to #30.

### LINKED ISSUE #45: 1.0 ABI: VTableItem types are data classes — copy()/componentN()/mutable vars frozen into the ABI

## Symbols
The vtable-builder item types are all `data class`-shaped (auto `copy()`, `componentN()`, `equals/hashCode/toString`) **and** expose mutable `var`s:
- `MethodVTableItem` — 7-arg constructor, `copy(...)`, `component1..7`, and `var callbackHandler/hasNoReply/inputParamNames/inputSignature/isDeprecated/isPrivileged/outputParamNames/outputSignature`.
- `PropertyVTableItem` — `copy(...)`, `component1..5`, `var getter/setter/signature/isDeprecated/isPrivileged`.
- `SignalVTableItem` — `copy(...)`, `component1..4`, `var signature/paramNames/isDeprecated`.
- `InterfaceFlagsVTableItem` — `copy(...)`, `component1`, `var hasNoReply/isDeprecated/isPrivileged`.
- `SignalEmitter`, `TypedMethod`, `TypedArguments`, `TypedMethodCall.{Sync,Async}MethodCall` similarly expose `copy/componentN`.

## What's risky
These are DSL builder targets (`addVTable { method(...) { ... } }`). Exposing them as data classes freezes a lot of incidental surface into 1.0:
- `componentN()` destructuring of a builder type is meaningless to users but is now ABI.
- `copy()` with the full positional constructor pins the **constructor parameter order** as ABI (reordering/adding a field becomes a break).
- Mutable `var`s + data semantics means `equals/hashCode` shift under mutation (builders aren't value-equal things).
The constructors themselves take raw `Function1<MethodCall, Unit>?` / `Signature?` etc. that arguably should also be builder-internal.

## Options
1. **Lock as-is.**
2. **Reduce to a builder contract before 1.0 (recommended):** drop `data` from the builder item types (kill `copy`/`componentN`), keep only the named `var`s/methods the DSL needs, so the ABI commits to the *builder behavior* not the *constructor tuple*. Consider making the primary constructors `internal` and driving construction through the `method{}`/`prop{}`/`signal{}` DSL only.
3. **Discuss** which of these are genuinely user-facing values vs pure builders.

## Recommendation
Option 2 — `data class` on DSL builders over-commits the ABI; convert to plain classes with a curated public surface before freezing.

Links back to #30.

monkopedia-coder:
Implements the owner decisions commented on #38, #40, #41, #42, and #45 on 2026-06-10 (ABI wave 1: visibility demotions). Behavior is unchanged — this is a visibility/shape-only PR.

### #38 — suspend is the only public async contract
- The callback-style `Proxy.callMethodAsync(MethodCall, (reply, error?) -> Unit)` overloads are removed from the `Proxy` interface (interface members cannot be `internal`) and restructured as **internal extension functions** that dispatch through a new internal `CallbackAsyncProxy` interface implemented by `JvmProxy` and `ProxyImpl`.
- `PendingAsyncCall` (expect + both actuals) and the `AsyncReplyHandler` typealias are now `internal`.
- The public suspend inline `Proxy.callMethodAsync(interfaceName, methodName, builder)` extension keeps using the callback machinery under the hood via a `@PublishedApi internal` impl function — identical body, so cancellation still releases the pending call and the JVM async-thread path is preserved.
- Same-module tests keep exercising the callback path (internal is visible to them); cross_test/stress_test only use the suspend form. #39 is subsumed (no public callback type remains to name).

### #40 — `createError` internal, `Error` KDoc
- `createError(errNo, customMsg)` is `internal` (common expect + jvm/native actuals).
- `Error` KDoc now documents that `message` is derived as `"$name: $errorMessage"`. No structural change; JVM errno naming intentionally untouched (deferred to #56's parity batch).
- cross_test's native peer (separate module) was migrated off `createError`: it now constructs `Error("org.freedesktop.DBus.Error.Failed", msg)` directly. Note: this changes only that test peer's wire error *name* (previously the sd-bus errno mapping for errno 1); the asserting side matches only on `errorMessage`, so coverage is unchanged. Flagging as the one judgment call outside the decision text.

### #41 — signature machinery minimized
- `SdbusSig` stays public as an **opaque** sealed handle (it is the return type of public `signatureOf`); its `value`/`valid`/`isTrivial` members were already `@PublishedApi internal`.
- `PrimitiveSig` (native + jvm), `ListSig`, `MapSig`, `StructSig`, `InvalidSig`, all fifteen top-level `*Sig` vals, and `SerialDescriptor.asSignature` are now `internal`.
- Hard requirement verified: `com.monkopedia.sdbus.internal.NativeTypeConverter` and `kotlinx.cinterop.CVariable` appear **zero** times in the regenerated `.api` dumps.
- `typed<A>()`, `signatureOf<A>()`/`signatureOf(KType)`, and the `call`/`args` DSL are untouched.

### #42 — cinterop `send` internal
- The native `MethodCall.send(sd_bus_message_handler_t, Any?, ULong)` overload is `internal`. Its only caller (`ConnectionImpl`) is already internal.

### #45 — VTable builder items de-data'd
- `MethodVTableItem`, `PropertyVTableItem`, `SignalVTableItem`, `InterfaceFlagsVTableItem`, `SignalEmitter`, `TypedMethod`, `TypedArguments`, `TypedMethodCall.SyncMethodCall`, `TypedMethodCall.AsyncMethodCall` drop `data` — `copy()`, `componentN()`, value `equals`/`hashCode`, and `toString` synthesis are gone from the ABI.
- Primary constructors are `internal` (`@PublishedApi internal` where the public inline DSL builders — `signal {}`, `prop {}`, `interfaceFlags {}`, `emitSignal {}` — construct them). `AsyncMethodCall.withContext` now constructs explicitly instead of `copy(...)`.
- The `addVTable { method(...) { ... } }` DSL is fully source-compatible; all named vars/builder methods stay public.

## Evidence: .api dump shrink

| dump | before | after | delta |
|---|---|---|---|
| `api/sdbus-kotlin.api` | 901 | 743 | −158 |
| `api/sdbus-kotlin.klib.api` | 1007 | 838 | −169 |
| total | 1908 | 1581 | **−327** |

Only one symbol was *added*: `@PublishedApi internal callMethodAsyncImpl` (plumbing for the suspend inline extension).

## Verification
- `ktlintFormat`, `apiDump`, `ktlintCheck apiCheck` — pass
- `jvmTest` under `dbus-run-session` — 236 tests, 0 failures
- `compileKotlinLinuxX64`, `compileKotlinLinuxArm64`, `compileTestKotlinLinuxX64`, `:cross_test:compileTestKotlinLinuxX64`, `:stress_test:compileTestKotlinJvm`, `:compile_test:build` — pass
- `:codegen:test` (golden output unaffected — generated code only uses the retained DSL/suspend surface), `:plugin:test` — pass

Fixes #38, Fixes #40, Fixes #41, Fixes #42, Fixes #45. Refs #39 (closed as subsumed).


monkopedia-reviewer (approved):
Reviewed against the owner decision comments on #38/#40/#41/#42/#45 (2026-06-10): all five visibility demotions implemented as specified; behavior-only-preserving (suspend async path keeps prior callback semantics incl. release-on-cancellation); api dumps shrink by 327 lines with zero NativeTypeConverter/CVariable leakage; sole added symbol is @PublishedApi callMethodAsyncImpl; cross_test peer createError migration verified sound (asserting sides match errorMessage only). CI fully green (14/14). Auto-merge authorized by owner for the pre-approved ABI wave.

Issue #30, opened 2026-06-08
Run 1.0 public-API review pass and fan findings out into per-finding tickets


monkopedia-coder:
Run a systematic review of the public API surface ahead of committing to 1.0 stability, and **fan the findings out into individual tickets** so each can be decided on separately.

## Task (this ticket = the review pass, not the decisions)
1. Walk the full BCV `apiDump` (`api/sdbus-kotlin.klib.api` + the JVM `.api`) as the inventory — every public symbol.
2. For each issue found, **file its own ticket**, labeled `needs-decision`, in the `1.0` milestone, linking back to this issue. Group closely-related findings into one ticket where it makes sense (e.g. all timeout-representation sites together); don't split hairs.
3. Each filed ticket should state: the symbol(s), what looks wrong/risky, the options (lock-as-is / change-before-1.0 and how / discuss), and a recommendation. Enough that we can just read it and rule.
4. When done, leave a summary comment here listing the tickets filed, and close this one.

This ticket itself decides nothing — it's discovery + fan-out. The decisions happen on the child tickets.

## What to look for
Naming inconsistencies, leftover sdbus-c++-isms, ergonomics of the async (`callMethodAsync`) / flow (`changes()`/`changesOrNull()`) / `Error` shapes, raw/unnamed types baked into the ABI, and anything we'd regret committing to. 1.0 means ABI stability, so the bar is "is this the shape we want to support indefinitely."

## Seed findings (already spotted — confirm + expand, not exhaustive)
- **Timeout represented two ways.** `Proxy` interface methods take raw `kotlin.ULong` (microseconds — sdbus-c++ carryover) while the idiomatic public extensions take `kotlin.time.Duration`. Should the `ULong` overloads be public, or should the interface speak `Duration` and keep micros internal?
- **Two parallel async APIs.** A callback style returning `PendingAsyncCall` and a `suspend` style returning `MethodReply`. Commit to both, or make the callback surface internal/removable before locking?
- **Async callback is a bare `Function2<MethodReply, Error?, Unit>`** rather than a named functional interface / typealias — locks an unnamed shape into the ABI and reads poorly at call sites.
- **`Error` is a single `final class Error : Exception(name, message)`.** Likely fine, but make it a deliberate "one error type, name+message" rather than a default.

## Done-when
- The full surface has been walked.
- Every finding has a `needs-decision` child ticket in the `1.0` milestone (grouped sensibly), linked here.
[...]

PR #60, opened 2026-06-10
1.0 ABI: UnixFd.adopt factory; normalize connection-factory surface (#48, #43)

### LINKED ISSUE #43: 1.0 ABI: connection-factory surface — naming, sprawl, and missing Duration/typed params

## Symbols (top-level connection factories)
- `createBusConnection()` / `createBusConnection(BusName)`
- `createSystemBusConnection()` / `createSystemBusConnection(BusName)`
- `createSessionBusConnection()` / `createSessionBusConnection(BusName)`
- `createSessionBusConnectionWithAddress(String)`
- `createRemoteSystemBusConnection(String)`
- `createDirectBusConnection(Int)` / `createDirectBusConnection(String)`
- `createServerBus(Int)`
Plus the object factories `createObject(Connection, ObjectPath)`, `createProxy(BusName, ObjectPath, Boolean=)`, `createProxy(Connection, BusName, ObjectPath, Boolean=)`, and `Connection.withService(BusName) -> ServiceConnection`.

## What's risky / inconsistent (sdbus-c++ carryover)
- **Naming inconsistency:** most are `create…BusConnection`, but the direct-server one is `createServerBus` (no `Connection` suffix) and `createDirectBusConnection` exists alongside it — the Bus/BusConnection naming is mixed. `createSessionBusConnectionWithAddress` spells out a variant in the name while `createDirectBusConnection(String)` overloads instead — two different conventions for "from an address/path."
- **Untyped params:** `createDirectBusConnection(Int)` / `createServerBus(Int)` take a bare `Int` fd (a `UnixFd`/`Int` socket) and `String` addresses are bare `String` rather than a typed address wrapper. The `Boolean` on `createProxy` is an unnamed positional flag (light vs full proxy?) frozen into the ABI.
- **Sprawl:** 11 connection factories is a lot to commit to indefinitely; some are near-duplicates differing only by overload.

## Options
1. **Lock as-is.**
2. **Normalize before 1.0 (recommended):** settle one naming convention (`create<Scope>BusConnection`), decide overload-vs-named-variant consistently, give the `Int` fd params a typed/`UnixFd` wrapper or at least named params, and replace the unnamed `Boolean` on `createProxy` with a named enum/flag. Prune genuine duplicates.
3. **Discuss** which factories are actually needed for 1.0 vs can be dropped/added later (dropping post-1.0 is the break we want to avoid, so err toward the minimal set now).

## Recommendation
Option 2 — at minimum fix `createServerBus`→naming consistency and the unnamed `Boolean`/bare-`Int` params before freezing.

Links back to #30.

### LINKED ISSUE #48: 1.0 ABI: UnixFd's public (Int, Unit) 'adopt' constructor uses Unit as a disambiguation hack

## Symbol
```
class UnixFd(val fd: Int, adoptFd: Unit) : Resource   // public
constructor(fd: Int) : this(checkedDup(fd), Unit)      // dups
constructor(other: UnixFd) : this(checkedDup(other.fd), Unit)
```
The `(Int, Unit)` constructor is the "adopt this fd without dup'ing" path; the `Unit` second parameter exists purely to disambiguate it from `(Int)` which dups. It is **public** in the klib/JVM ABI.

## What's risky
`adoptFd: Unit` is a non-idiomatic disambiguation hack leaking into the public API. A caller sees `UnixFd(fd, Unit)` with no idea it means "take ownership, don't dup." Freezing this into 1.0 commits us to:
- an unexplained `Unit` parameter in the constructor ABI,
- the ownership semantics (dup vs adopt) being expressed by *which overload you picked* rather than anything readable.

## Options
1. **Lock as-is.**
2. **Replace before 1.0 (recommended):** make the adopt path either `internal` (if users never adopt raw fds) or a named factory: `UnixFd.adopting(fd)` / `UnixFd.adopt(fd)` vs the dup'ing primary constructor. Drop the `Unit` parameter from the public surface.
3. **Discuss** whether raw-fd adoption is even a supported public operation for 1.0.

## Recommendation
Option 2 — a named `adopt`/`adopting` factory (or internal) instead of a public `Unit`-disambiguated constructor.

Links back to #30.

monkopedia-coder:
Implements the owner decisions of 2026-06-10 for issues #48 and #43 (wave PR 4 of the 1.0 public-API review, #30).

Fixes #48, Fixes #43.

## #48 — UnixFd adopt factory

- New public companion factory `UnixFd.adopt(fd: Int): UnixFd`: takes ownership of the fd **without dup'ing** (the exact semantics of the old `(Int, Unit)` constructor).
- The `(Int, Unit)` Unit-disambiguation-hack constructor is now `internal` on the expect class and both actuals — it leaves the public klib/JVM ABI entirely.
- `UnixFd(fd)` primary constructor keeps its safe dup-by-default semantics; `UnixFd(other: UnixFd)` copy constructor unchanged.

## #43 — connection-factory normalization (all 11 factories survive)

1. **One naming convention** `create<Scope>BusConnection`: `createServerBus(fd)` → `createServerBusConnection(fd)`.
2. **Overloads over name-variants**: `createSessionBusConnectionWithAddress(String)` → `createSessionBusConnection(address: String)`. No JVM signature clash with `createSessionBusConnection(ServiceName)`: the value-class overload is mangled (`createSessionBusConnection-em-jQOc`) — verified in the .api dump.
3. **Typed fd params**: `createDirectBusConnection(fd: UnixFd)` and `createServerBusConnection(fd: UnixFd)` replace the bare-`Int` overloads. The JVM actuals keep their `@Deprecated(ERROR)` native-only markers.
4. **Negated flag fixed**: `createProxy(..., dontRunEventLoopThread: Boolean = false)` → `createProxy(..., runEventLoopThread: Boolean = true)` on both overloads, and `ServiceConnection.createProxy(objectPath, runEventLoopThread = true)` mirrors it. The flag controls whether the proxy starts the connection's I/O event loop in an internally managed thread, so the positive name says exactly what it does. All internal plumbing (`ProxyImpl`, `JvmDbusBackend`, `PureJavaDbusBackend`, stub backend) renamed and conditions inverted; **default behavior is byte-for-byte identical**.

## fd ownership semantics (preserved exactly)

The old `Int` overloads adopted the fd: sd-bus (`sd_bus_open_direct`/`sd_bus_open_server`) takes over the descriptor on success; on failure ownership stayed with the caller. The `UnixFd` versions keep that contract and make it coherent with UnixFd's RAII ownership:

- On success the connection **adopts** the descriptor: the passed `UnixFd` is detached (new internal `detach()` disarms both `release()` and the GC cleaner without closing), so there is no double-close when the UnixFd is released/collected.
- On failure (factory throws) the `UnixFd` still owns the descriptor, as before.
- Call sites choose dup-vs-adopt explicitly: `createServerBusConnection(UnixFd.adopt(rawFd))` hands off ownership; `createServerBusConnection(UnixFd(rawFd))` passes a duplicate. This is the #48/#43 pairing from the decisions.

KDoc on both factories documents all of this.

## Dependency note (flagged judgment call)

cross_test's jvmTest failed to compile calling `UnixFd.adopt` because `UnixFd.Companion`'s supertype `KSerializer` wasn't resolvable: kotlinx-serialization was only an `implementation` dependency while KSerializer types are referenced throughout the public API (`Variant`, every name-type companion, `TypedArguments`). Promoted it to `api(libs.kotlinx.serialization)` in commonMain — this is the principled fix and changes only the published POM scope, not the ABI dumps.

## API dump delta

- jvm + klib: `- UnixFd <init>(Int, Unit)`, `+ UnixFd.Companion.adopt(Int)`
- jvm + klib: `createDirectBusConnection(I)` → `(UnixFd)`, `createServerBus(I)` → `createServerBusConnection(UnixFd)`, `createSessionBusConnectionWithAddress(String)` → `createSessionBusConnection(String)`
- `createProxy`/`ServiceConnection.createProxy` signatures unchanged in the dumps (parameter rename only; Boolean stays, default flipped polarity to preserve behavior)

## Verification

- `ktlintFormat`, `apiDump` (delta reviewed against both specs), `ktlintCheck apiCheck` — pass
- `dbus-run-session jvmTest` — pass
- `dbus-run-session linuxX64Test` — pass (includes DirectConnectionFixture server/direct fd connection tests, exactly the path this PR touches)
- `compileKotlinLinuxX64 compileTestKotlinLinuxX64 :codegen:test :plugin:test :compile_test:build :cross_test:compileTestKotlinJvm :cross_test:compileTestKotlinLinuxX64 :stress_test:compileTestKotlinJvm` — pass

All call sites updated (src tests, cross_test incl. 7 positional `true` flags in NativeInteropPeerTest that the rename would otherwise have silently inverted, stress_test, fixtures); behavior 100% unchanged.


monkopedia-reviewer (approved):
Approving per owner decisions of 2026-06-10 on #48 (Option 2: public UnixFd.adopt(fd) companion factory; (Int, Unit) ctor leaves the public surface) and #43 (Option 2: normalize, keep all 11 factories — create<Scope>BusConnection naming, overloads over name-variants, UnixFd-typed fd params, positively-named event-loop flag).

Spec fidelity verified against both decision comments; API dump delta contains exactly the intended changes (klib factory count remains 11; createSessionBusConnection(String) overload doesn't clash with the mangled value-class overload em-jQOc; createProxy dump entries unchanged).

Extra-scrutiny areas:
1. Boolean inversion: traced every changed flag — expect/actual createProxy (default false→true with name negation flipped), ProxyImpl, JvmDbusBackend interface, StubJvmDbusBackend, PureJavaDbusBackend (all 'if (!dontRun)' → 'if (run)'), ServiceConnection passthrough, and every call site (commonTest x10, jvmTest mocks incl. mock-expectation booleans, cross_test, stress_test, nativeTest). All 7 positional 'true' sites in NativeInteropPeerTest correctly became runEventLoopThread = false. Branch-wide grep finds zero remaining dontRun code references (one stale COVERAGE_TODO.md doc line only). Default behavior identical.
2. fd ownership/detach: native detach() CAS-consumes the singleCall closer with -1 (guarded by 'it >= 0'), so release() and the GC cleaner become no-ops without closing — no double-close on success; on factory throw, .also{} never runs and the UnixFd retains ownership — no leak. JVM detach() (disarm + clean) is equivalent. detach() is internal and called only from the two adopting connection factories; adopt() is the only user of the now-internal (Int, Unit) ctor.
3. kotlinx-serialization implementation→api: claim verified — KSerializer is a supertype of many public companions (UnixFd, BusName, InterfaceName, MemberName, ObjectPath...) and appears in public function signatures, so api() scope is the principled fix; POM-scope-only, no .api/.klib.api dump change.
4. @Deprecated(ERROR) JVM markers: pre-existing on main (dab35d1) marking the fd factories native-only on JVM (expect/actual requires JVM actuals); kept on the renamed functions with messages updated. No deprecated aliases remain at removed names — createServerBus and createSessionBusConnectionWithAddress are fully gone from both dumps, consistent with the #43 spec.

CI: all 13 checks pass (full-tests-x64, arm-runtime-tests, native-build-x64, static-checks, 9x pr-smoke matrix); nightly-stress skipped as designed. Merging per pre-approved ABI-wave authorization.

PR #61, opened 2026-06-10
1.0 ABI: rename PropertyDelegate flow()/flowOrNull() to values()/valuesOrNull() (#46)

### LINKED ISSUE #46: 1.0 ABI: PropertyDelegate has duplicate flow accessors — changes()/changesOrNull() vs flow()/flowOrNull()

## Symbols (on `PropertyDelegate<A, B>`)
- `changes(): Flow<B>` and `flow(): Flow<B>`
- `changesOrNull(): Flow<B?>` and `flowOrNull(): Flow<B?>`
- plus `get(): B`, `getOrNull(): B?`, `await(): B` (suspend)
- `MutablePropertyDelegate` adds `set(B)` / `setValue(...)`.

## What's risky
There appear to be **two parallel names for the same thing**: `changes()`/`changesOrNull()` and `flow()`/`flowOrNull()`. If they are identical (or near-identical) this is gratuitous ABI surface and a "which do I call?" papercut frozen into 1.0; if they differ subtly (e.g. one replays current value, one doesn't) that difference is undocumented at the signature level and is a trap. Either way it should be a deliberate decision before freezing.

Also worth ruling on while here: the `…OrNull` variant pairs (`get`/`getOrNull`, `changes`/`changesOrNull`) — confirm the non-`OrNull` ones throw vs the property genuinely being absent, so the contract is intentional.

## Options
1. **Lock as-is** (keep both name families) — only if they have a documented, intended difference.
2. **Collapse before 1.0 (recommended):** pick one name family (`changes`/`changesOrNull` reads best for a property-change stream) and drop the other, or alias one as deprecated-from-day-1 (which we don't want at 1.0). Document the throwing-vs-null contract.
3. **Discuss** whether `changes()` (stream) and `flow()` are meant to differ (e.g. initial-value semantics) — if so, rename to make the difference obvious rather than two generic names.

## Recommendation
Option 2 — collapse to one name family unless there's a real semantic distinction, in which case rename for clarity.

Links back to #30.

monkopedia-coder:
Implements the owner-approved decision on #46 (2026-06-10): keep both observation semantics on `PropertyDelegate`, rename for contrast.

## Changes

- `flow(): Flow<T>` → **`values(): Flow<T>`** — semantics unchanged: `changes().onStart { emit(get()) }`, i.e. current value first, then changes.
- `flowOrNull(): Flow<T?>` → **`valuesOrNull(): Flow<T?>`** — semantics unchanged: `changesOrNull().onStart { emit(getOrNull()) }`.
- `changes()` / `changesOrNull()` keep their names — they accurately say "change events only".
- Documentation pass on the whole accessor family:
  - `changes` vs `values` distinction made explicit on all four flow methods (change events only vs current-value-first).
  - Throwing-vs-null contract documented: `get()`/`values()` throw when the property is absent; `getOrNull()` returns null on `org.freedesktop.DBus.Error.InvalidArgs` (other errors still thrown); `changes()` drops invalidation events; `changesOrNull()` emits `null` on invalidation.

**BREAKING CHANGE:** `PropertyDelegate.flow()/flowOrNull()` renamed to `values()/valuesOrNull()`; `changes()/changesOrNull()` unchanged. Behavior is 100% identical — rename + docs only.

## API dump delta

Exactly four lines in each of `sdbus-kotlin.api` / `sdbus-kotlin.klib.api`: `flow`/`flowOrNull` removed, `values`/`valuesOrNull` added. Nothing else.

## Call sites

No call sites existed outside the definitions — no tests, samples (incl. bluez-scan), or generated code use `flow()`/`flowOrNull()`.

## Verification (JAVA_HOME=java-21-openjdk)

- `ktlintFormat` + `apiDump` (delta as above)
- `ktlintCheck apiCheck` — pass
- `dbus-run-session -- ./gradlew jvmTest` — pass
- `compileKotlinLinuxX64 compileTestKotlinLinuxX64 :codegen:test :plugin:test :compile_test:build :cross_test:compileTestKotlinJvm :stress_test:compileTestKotlinJvm` — pass

Fixes #46.

### DIFF
--- a/api/sdbus-kotlin.api
+++ b/api/sdbus-kotlin.api
@@ -425,8 +425,6 @@ public class com/monkopedia/sdbus/PropertyDelegate : kotlin/properties/ReadOnlyP
 	public final fun await (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun changes ()Lkotlinx/coroutines/flow/Flow;
 	public final fun changesOrNull ()Lkotlinx/coroutines/flow/Flow;
-	public final fun flow ()Lkotlinx/coroutines/flow/Flow;
-	public final fun flowOrNull ()Lkotlinx/coroutines/flow/Flow;
 	public final fun get ()Ljava/lang/Object;
 	public final fun getInterfaceName-DgtVVXE ()Ljava/lang/String;
 	protected final fun getModule ()Lkotlinx/serialization/modules/SerializersModule;
@@ -437,6 +435,8 @@ public class com/monkopedia/sdbus/PropertyDelegate : kotlin/properties/ReadOnlyP
 	protected final fun getSignature ()Lcom/monkopedia/sdbus/SdbusSig;
 	protected final fun getType ()Lkotlinx/serialization/KSerializer;
 	public fun getValue (Ljava/lang/Object;Lkotlin/reflect/KProperty;)Ljava/lang/Object;
+	public final fun values ()Lkotlinx/coroutines/flow/Flow;
+	public final fun valuesOrNull ()Lkotlinx/coroutines/flow/Flow;
 }
 
 public final class com/monkopedia/sdbus/PropertyGetReply : com/monkopedia/sdbus/Message {
--- a/api/sdbus-kotlin.klib.api
+++ b/api/sdbus-kotlin.klib.api
@@ -659,10 +659,10 @@ open class <#A: kotlin/Any?, #B: kotlin/Any> com.monkopedia.sdbus/PropertyDelega
 
     final fun changes(): kotlinx.coroutines.flow/Flow<#B> // com.monkopedia.sdbus/PropertyDelegate.changes|changes(){}[0]
     final fun changesOrNull(): kotlinx.coroutines.flow/Flow<#B?> // com.monkopedia.sdbus/PropertyDelegate.changesOrNull|changesOrNull(){}[0]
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Approving per the owner decision on #46 (2026-06-10): keep both observation semantics, rename flow()/flowOrNull() to values()/valuesOrNull() for contrast with changes()/changesOrNull(), and document the accessor family.

Verification:
- Pure rename: values()/valuesOrNull() bodies are byte-identical to the old flow()/flowOrNull() (changes().onStart { emit(get()) } / changesOrNull().onStart { emit(getOrNull()) }). changes()/changesOrNull() untouched.
- API dump delta is exactly the expected 4 lines in each of sdbus-kotlin.api and sdbus-kotlin.klib.api (flow/flowOrNull removed, values/valuesOrNull added; nothing else).
- KDoc claims checked against the code: getOrNull() returns null only on org.freedesktop.DBus.Error.InvalidArgs and rethrows everything else (matches the catch block); changes() drops invalidation-only signals via mapNotNull on changedProperties; changesOrNull() emits null on invalidation; values() throws on collection start if the property is absent (initial get()). All docs match actual semantics.
- No stale flow()/flowOrNull() references anywhere in src, api, codegen, or samples.
- CI: all checks green (static-checks, native-build-x64, full-tests-x64, arm-runtime-tests, all 9 pr-smoke matrix jobs; nightly-stress skipped as expected). No flake re-runs needed.

Owner pre-authorized auto-merge-on-green for this ABI wave; merging.

PR #59, opened 2026-06-10
1.0 ABI: idiomize Message accessors — typed val properties, named peekType result (#44)

monkopedia-coder:
PR 3 of the 1.0 ABI wave. Implements the owner decision commented on #44 (2026-06-10): Option 2 — full idiomization of the `Message` accessors.

Fixes #44.

## What changed

On `sealed class Message` (expect + jvm/native actuals), all pure `getX()` accessors become `val` properties, with name-shaped returns using the library's existing value classes:

| Before | After |
|---|---|
| `getDestination(): String?` | `val destination: BusName?` |
| `getSender(): String?` | `val sender: BusName?` |
| `getInterfaceName(): String?` | `val interfaceName: InterfaceName?` |
| `getMemberName(): String?` | `val memberName: MemberName?` |
| `getPath(): String?` | `val path: ObjectPath?` |
| `getSELinuxContext(): String` | `val seLinuxContext: String` |
| `getCredsPid(): Int` | `val credsPid: Int` |
| `getCredsUid()/getCredsEuid()/getCredsGid()/getCredsEgid(): UInt` | `val credsUid/credsEuid/credsGid/credsEgid: UInt` |
| `getCredsSupplementaryGids(): List<UInt>` | `val credsSupplementaryGids: List<UInt>` |
| `peekType(): Pair<Char?, String?>` | `fun peekType(): PeekedType` |

- The raw string is wrapped at the accessor boundary; `null` stays `null`. Creds accessors keep primitive types per the decision (no uid/gid value classes).
- `PeekedType` is a new small named result type (`val type: Char?`, `val contents: String?`) replacing the anonymous `Pair` — per the #45 precedent it is a plain class (not a data class): it is a transient peek result, not a value type, so no `copy`/`componentN`/structural equality is exposed in the ABI. `peekType()` stays a function since it performs a wire-cursor peek, not a stable attribute read.
- `BREAKING CHANGE`: source-breaking for any consumer of these accessors; behavior is 100% unchanged (accessor shape only).

## Scope

- Message hierarchy only (`Message` + its subclasses, which inherit these accessors — none duplicated them). Connection/Proxy/StandardInterfaces `getX` methods are untouched; those belong to #49's PR.
- `Signal.setDestination(String)` left alone (mutator, not a pure accessor; #49 / later wave can address it if desired).
- All call sites updated across commonMain/jvmMain/nativeMain, common/jvm/native tests, and cross_test. compile_test, stress_test, samples, codegen, plugin had no usages.
[...]

Issue #44, opened 2026-06-09
1.0 ABI: Message accessors are Java-style getX() returning raw nullable String, not typed value classes


monkopedia-coder:
Child of #30 (1.0 public-API review).

## Symbols (on `sealed class Message`)
- `getDestination(): String?`, `getInterfaceName(): String?`, `getMemberName(): String?`, `getPath(): String?`, `getSender(): String?`, `getSELinuxContext(): String`
- `getCredsUid()/getCredsEuid()/getCredsGid()/getCredsEgid(): UInt`, `getCredsPid(): Int`, `getCredsSupplementaryGids(): List<UInt>`
- `peekType(): Pair<Char?, String?>`
- Also `Connection`/`Object`/`Proxy` use `getUniqueName()`, `getMethodCallTimeout()`, `getMachineId()`, `getAll()` etc. as `getX()` methods rather than Kotlin properties.

## What's risky (sdbus-c++ / C++-getter carryover)
- **Non-idiomatic `getX()` methods** where Kotlin would use `val`s — `currentlyProcessedMessage`, `objectPath`, `connection` *are* properties on the interfaces, but `getDestination()`/`getInterfaceName()`/`getMemberName()` etc. are methods, so the surface is inconsistent (some concepts are properties, some are getter-methods).
- **Raw `String?` instead of the library's own value classes.** The library has `InterfaceName`, `MemberName`, `ObjectPath`, `BusName` value classes, yet `Message.getInterfaceName(): String?`, `getMemberName(): String?`, `getPath(): String?`, `getDestination(): String?`, `getSender(): String?` return bare strings. Freezing the untyped form into 1.0 throws away the type-safety the value classes exist to provide.
- `peekType(): Pair<Char?, String?>` exposes an anonymous `Pair` of raw `Char?`/`String?` — an unnamed/structural return baked into the ABI.

## Options
1. **Lock as-is.**
2. **Idiomize before 1.0 (recommended):** convert getter-methods to `val` properties where they're pure accessors; return the existing value classes (`InterfaceName?`, `MemberName?`, `ObjectPath?`, `BusName?`) instead of `String?`; replace `peekType`'s anonymous `Pair` with a small named type.
3. **Discuss** the creds accessors (`UInt`/`List<UInt>`) — likely fine, but confirm the grouping/naming.

## Recommendation
Option 2 for the name/path accessors at least — returning raw `String?` where typed value classes exist is the clearest "we'll regret freezing this" case.

[...]

PR #62, opened 2026-06-10
1.0 ABI: naming sweep — startEventLoop/stopEventLoop, properties over getters, INTERFACE_NAME dedupe (#49)

### LINKED ISSUE #49: 1.0 ABI: naming/sdbus-c++-isms grab-bag (addObjectManager, enterEventLoopAsync, getX setX, INTERFACE_NAME duplication)
Child of #30 (1.0 public-API review). Catch-all for smaller naming/consistency calls found in the walk; each is individually small but collectively worth ruling on before freezing.

## Findings

1. **`addObjectManager` overloaded across two meanings.** `Connection.addObjectManager(ObjectPath): Resource` (register a manager at a path) vs `Object.addObjectManager(): Resource` (no-arg, on the object's own path). Same name, two receivers, subtly different contracts — confirm this is intended or rename one.

2. **Event-loop method naming (sdbus-c++ carryover).** `Connection.enterEventLoopAsync()` (not suspend, not returning a handle) alongside `suspend leaveEventLoop()`. `enterEventLoopAsync` reads like it should be suspend or return something; the enter/leave asymmetry (one plain, one suspend) is a C++-loop idiom. Confirm the intended Kotlin shape.

3. **`getX()`/`setX()` methods vs properties, broadly.** `Connection.getUniqueName()`, `getMethodCallTimeout()`/`setMethodCallTimeout()`, `PeerProxy.getMachineId()`, `PropertiesProxy.getAll()`, `AsyncAllPropertiesGetter.getResult()`, `AsyncPropertySetter.getResult()` are getter-methods where several siblings are `val` properties. (Message accessors are tracked separately in the Message ticket.) Decide a consistent property-vs-method convention.

4. **`INTERFACE_NAME` / `DBUS_PROPERTIES_INTERFACE_NAME` duplicated across companions.** `PeerProxy.Companion.INTERFACE_NAME`, `PropertiesProxy.Companion.INTERFACE_NAME`, `ObjectManagerProxy.Companion.INTERFACE_NAME`, and `DBUS_PROPERTIES_INTERFACE_NAME` exists both top-level and on `AllPropertiesGetter`/`AsyncAllPropertiesGetter`/`AsyncPropertyGetter`/`AsyncPropertySetter` companions. Multiple public sources of the same constant — pick canonical homes.

5. **`onPropertiesChanged` on `PropertiesProxy` is public** (looks like a callback hook meant to be overridden/internal) — confirm it belongs on the public proxy contract.

6. **`registerPropertiesProxy()` / `registerSignalHandler(...)` / `registerProperty(...)`** — `register*` naming is fine but mixes "returns Resource" (signal handler) with "returns Unit/item"; confirm the convention.

## Options (per finding)
Lock-as-is / rename-before-1.0 / discuss. None of these are deep; they're "is this the name/shape we want forever" calls.

## Recommendation
Sweep these together in one naming pass: settle property-vs-`getX`, resolve the `addObjectManager` double-meaning and the `enterEventLoopAsync` shape, and dedupe the `INTERFACE_NAME` constants to canonical locations. Confirm `onPropertiesChanged` visibility.

Links back to #30.

monkopedia-coder:
Implements issue #49 (owner decision 2026-06-10): the full naming/consistency sweep, six findings in one coordinated PR. Behavior is 100% unchanged; all call sites in src, tests, cross_test, stress_test, compile_test updated (no references existed in samples, codegen templates, plugin, or docs).

## Per finding

**1. `addObjectManager` pair — KEPT both.** `Connection.addObjectManager(ObjectPath)` and `Object.addObjectManager()` are unchanged; each KDoc now cross-references the other (`@see` + a sentence explaining which to use when).

**2. Event loop rename.** `Connection.enterEventLoopAsync()` → `startEventLoop()`, `suspend leaveEventLoop()` → `suspend stopEventLoop()`. Pure rename, no shape change. KDoc rewritten: the old "The same as enterEventLoop, except that it doesn't block" text is gone; the new docs describe start (spawns the internally managed loop thread) and stop (exits the loop, frees the thread) honestly.
*Judgment call:* there is **no** public blocking `enterEventLoop()` on the surface — the only other variant is the internal loop body inside `ConnectionImpl`'s event-thread machinery (`internal suspend fun enterEventLoop()`), which is not public API and was left alone. So no `runEventLoop()` was needed. The internal JVM backend interface (`JvmDbusConnection`) loop methods were renamed in lockstep to avoid a confusing internal/public name mismatch.

**3. Properties over getters — classification of every getX/setX on the surface:**

| Member | Classification | Result |
|---|---|---|
| `Connection.getUniqueName(): BusName` | pure local accessor (sd_bus cached unique name) | → `val uniqueName: BusName` |
| `Connection.getMethodCallTimeout()/setMethodCallTimeout(Duration)` | pure local accessor (bus-local setting) | → `var methodCallTimeout: Duration` |
| `PeerProxy.getMachineId()` | **remote D-Bus call** | stays method |
| `PropertiesProxy.getAll()/getAllAsync()/setAsync()` | remote calls | stay methods |
| `AsyncAllPropertiesGetter.getResult()`, `AsyncPropertySetter.getResult()`, `AsyncPropertyGetter.get()` | remote calls | stay methods |
| `ObjectManagerProxy.getManagedObjects()` | remote call (GetManagedObjects) | stays method |
| `PropertyDelegate.get()/getOrNull()`, `MutablePropertyDelegate.set()` | remote Get/Set | stay methods |
| `Proxy.getProperty/setProperty/getAllProperties/...Async` extensions | remote calls / builder factories | stay |
| `Flags.set(flag, value)/test(flag)` | keyed mutator/query, not an accessor | stay |
| everything else (`proxy`, `connection`, `objectPath`, `currentlyProcessedMessage`, `destination`, creds, `dontExpectReply`, `fd`, ...) | already `val`/`var` | unchanged |

*Judgment call:* `getUniqueName()` already returned `BusName`, so `val uniqueName: BusName` needed no type change. Note JVM binary signatures for the new properties are identical to the old methods (`getUniqueName-AwvMx1g`, `get/setMethodCallTimeout`) — the JVM `.api` diff for finding 3 is empty by construction; the klib dump shows the property conversion.

**4. `INTERFACE_NAME` dedupe.** `PeerProxy`, `PropertiesProxy`, `ObjectManagerProxy` each keep `INTERFACE_NAME` on their own companion. The five duplicated `DBUS_PROPERTIES_INTERFACE_NAME` copies (top-level in PropertyGetter.kt + `AllPropertiesGetter`/`AsyncAllPropertiesGetter`/`AsyncPropertyGetter`/`AsyncPropertySetter` companions) are all removed; internal usages now reference the canonical home.
*Judgment call:* canonical = **`PropertiesProxy.Companion.INTERFACE_NAME`** (not the top-level val). It matches the established per-proxy pattern from the first sentence of the decision and reads best (`PropertiesProxy.INTERFACE_NAME`); since this wave is breaking anyway, "minimize breakage" did not outweigh consistency. The four now-empty builder companions were deleted entirely.

**5. `onPropertiesChanged` internalized.** Kotlin forbids `internal` members on a public interface, so literal internalization was impossible. The hook is removed from the `PropertiesProxy` contract and its functionality moved into a callback parameter: `registerPropertiesProxy(onPropertiesChanged: (InterfaceName, Map<PropertyName, Variant>, List<PropertyName>) -> Unit)`. Behavior is identical (same signal subscription, same dispatch).
*Judgment calls:* (a) codegen does **not** reference it — the `onPropertiesChanged`/`PropertiesProxy` hits in codegen test resources are coincidentally-named generated types from XML fixtures, so no `@PublishedApi` was needed; (b) the callback parameter has **no default**: a no-arg overload would silently stop delivering updates to existing subclasses that override the removed hook — better a loud compile error than a silent behavior change. The nativeTest fixture (`IntegrationTestsProxy`/`TestProxy`) was migrated to the callback and keeps its own open hook locally.

**6. `register*` naming kept, convention documented.** KDoc on `Proxy.registerSignalHandler`, `PropertiesProxy.registerPropertiesProxy`, and `registerProperty` now states the convention: `register*` returns a `Resource` when the registration must be explicitly released, and `Unit` (or the registered item) otherwise.

**From #44's review (same sweep scope): `Signal.setDestination`.** Kept as a method — `override var destination` is impossible (`Message.destination` is `val destination: BusName?` from #59; a var/val + nullability clash). For type consistency it now takes `BusName` instead of `String`: `fun setDestination(destination: BusName)`. Zero call sites existed outside the expect/actual declarations. KDoc notes the value is observable via `Message.destination`.

## API dump delta
- JVM `.api`: `enterEventLoopAsync`/`leaveEventLoop` → `startEventLoop`/`stopEventLoop`; `PropertiesProxy.onPropertiesChanged` removed; `registerPropertiesProxy()` → `registerPropertiesProxy(Function3)`; five `DBUS_PROPERTIES_INTERFACE_NAME` declarations (4 companions + `PropertyGetterKt` top-level) removed; `Signal.setDestination(String)` → mangled `setDestination-em-jQOc(String)` (BusName value class). Property conversions are JVM-binary-identical.
- klib `.api`: same, plus `getUniqueName()` → `val uniqueName`, `get/setMethodCallTimeout` → `var methodCallTimeout`, `setDestination(kotlin/String)` → `setDestination(BusName)`.

## Verification
- `./gradlew ktlintFormat` + `apiDump` (diff reviewed against spec), `ktlintCheck apiCheck` ✅
- `dbus-run-session -- ./gradlew jvmTest` ✅
- `dbus-run-session -- ./gradlew linuxX64Test` ✅ (root: 19 suites, 0 failures; cross_test native suites also green — event-loop rename exercised on native lifecycle)
- `compileKotlinLinuxX64 compileTestKotlinLinuxX64 :codegen:test :plugin:test :compile_test:build :cross_test:compileTestKotlinJvm :cross_test:compileTestKotlinLinuxX64 :stress_test:compileTestKotlinJvm` ✅

BREAKING CHANGE: enterEventLoopAsync/leaveEventLoop renamed startEventLoop/stopEventLoop; pure-accessor getX/setX methods become properties; duplicate DBUS_PROPERTIES_INTERFACE_NAME constants removed; onPropertiesChanged internalized.

Owner decision 2026-06-10. Fixes #49.


monkopedia-reviewer (approved):
Approving per the owner decision on #49 (2026-06-10), which pre-approved this six-finding naming sweep with auto-merge-on-green.

Reviewed the full diff against the spec, with extra scrutiny on the five flagged areas:

1. **registerPropertiesProxy redesign (finding 5)**: Faithful to the intent — Kotlin forbids internal members on public interfaces, so removing the open onPropertiesChanged hook and making it a required callback parameter of registerPropertiesProxy removes the hook from the public contract while preserving identical signal subscription/dispatch. Verified all callers: the only in-repo caller is the nativeTest fixture (IntegrationTestsProxy), which wires the callback to a local open hook (overridden by TestProxy) — behavior unchanged. Codegen does NOT reference the hook: the onPropertiesChanged/PropertiesProxy hits in codegen test resources are coincidentally-named generated types from an XML 'Properties' interface fixture (a suspend signal *emitter* on an adaptor, not the library hook). Requiring the parameter (no default) is the right call — a defaulted no-op would compile silently for subclasses that overrode the removed hook and drop their updates.

2. **Property conversions (finding 3)**: getUniqueName→val uniqueName, get/setMethodCallTimeout→var methodCallTimeout. Verified in the JVM .api diff that getUniqueName-AwvMx1g / getMethodCallTimeout-UwyO8pc / setMethodCallTimeout-LRDsOJo lines are untouched context — JVM binary signatures identical as claimed; only the klib dump shows the val/var conversion. Classification table is honest: both converted members are local sd_bus accessors (sd_bus_get_unique_name, sd_bus_get/set_method_call_timeout); everything doing remote I/O (getMachineId, getAll/getResult builders, PropertyDelegate get/set, getManagedObjects) stays a method.

3. **Event-loop rename (finding 2)**: startEventLoop/stopEventLoop everywhere — grep of the PR head finds zero stale enterEventLoopAsync/leaveEventLoop references in code, KDoc, README, or codegen. The only remaining enterEventLoop is the internal loop body inside ConnectionImpl's event-thread machinery (non-public, disclosed in the PR body). Internal JvmDbusConnection renamed in lockstep.

4. **INTERFACE_NAME dedupe (finding 4)**: canonical home is PropertiesProxy.Companion.INTERFACE_NAME; the top-level val and all four builder-companion duplicates are gone from both dumps, the four now-empty companions deleted, and all internal usages (ConvenienceApiClasses, PropertyGetter, PropertyDelegate) reference the canonical constant. PeerProxy/ObjectManagerProxy keep their own per the spec.

5. **Signal.setDestination(BusName)**: zero call sites outside the expect/actual declarations (verified by grep). No clash with Message.destination from #59 — it remains val destination: BusName? on Message, and setDestination is a plain function whose JVM name mangles to setDestination-em-jQOc, so no accidental accessor pairing.

Findings 1 (keep both addObjectManager, cross-referencing KDoc) and 6 (register* convention documented) are implemented exactly as specified. API dump deltas match the PR body claims.

CI: all checks green (static-checks, native-build-x64, full-tests-x64, arm-runtime-tests, all 10 pr-smoke matrix legs pass; nightly-stress skipped as usual) — no flake re-runs needed. Merging per the pre-approved auto-merge-on-green authorization.

PR #63, opened 2026-06-10
JVM parity: strict reply deserialization, call-after-release throws, errno error names (#56)

### LINKED ISSUE #56: JVM backend more lenient than native on two failure paths (deserialization + call-after-release)

monkopedia-coder:
Implements the owner decision on #56 (2026-06-10): tighten the pure-JVM backend to match the native sd-bus backend on three failure paths. Native is the reference implementation. **Behavior-only — no public API/ABI change; the `api/*.api` dumps are untouched (`apiDump` verified as a no-op).**

## Fix 1 — strict reply deserialization (`src/jvmMain/.../Message.jvm.kt`)

The JVM deserializer only enforced signature compatibility for basic expected types, so a reply whose actual signature was `i` silently "deserialized" into an expected struct `(ii)` (the erased generic cast can't catch it). Enforcement now also covers container-shaped expected types (`a`, `(`, `v`), compared by top-level type class only — value-based inference cannot reconstruct nested signatures (an empty list infers as `a`, dict/variant element types come from the first runtime value), so element types are deliberately not over-enforced.

The mismatch now throws `Error("System.Error.ENXIO", ...)` — the exact name native produces (`sd_bus_message_enter_container` returns `-ENXIO` on a struct/body mismatch; verified empirically against libsystemd).

**Restored test** (the sub-case dropped in PR #54): `TypeMatrixRoundTripTest.mismatchedReply_intIntoStruct_throwsOnBothBackends` — server replies `i`, client deserializes `(ii)`, asserts `assertFailsWith<Error>` **and** the shared `System.Error.ENXIO` name on both backends.

## Fix 2 — call-after-release throws (`src/jvmMain/.../PureJavaDbusBackend.kt`)

After `stopEventLoop` + `release()`, the JVM proxy kept servicing calls through the in-process static-dispatch short circuit. `PureJavaDbusConnection` now tracks a released flag (making `release()` idempotent too, like native), and `PureJavaDbusProxy` guards both `callMethod` paths with `require(!released) { "Connection has already been released" }` — the **exact exception type and message** native raises (`ConnectionImpl.checkNotReleased`), plus an `ENOTCONN` `Error` for a connection that dropped without an explicit release (mirroring `sd_bus_call` on a closed bus).

**Restored test** (the sub-case scoped down in PR #54 with a code comment documenting the divergence): `FailurePathParityTest.proxyConnectionTeardown_completesCleanly_andCallAfterReleaseFails` — asserts `assertFailsWith<IllegalArgumentException>` with message `"Connection has already been released"` on both backends. (Running it against native first is what pinned the exact contract — native throws `IllegalArgumentException`, not `Error`, on this path.)

## Fix 3 — errno error-name parity (`src/jvmMain/.../Error.jvm.kt`, from #40)

JVM `createError(errNo, msg)` fabricated `Error("jvm.errno.$errNo", msg)`. It now mirrors native's `sd_bus_error_set_errno` mapping, with every table entry verified empirically against libsystemd on this machine:

- EINVAL → `org.freedesktop.DBus.Error.InvalidArgs`, EACCES/EPERM → `...AccessDenied`, ENOMEM → `...NoMemory`, ETIMEDOUT → `...Timeout`, ENOENT → `...FileNotFound`, ECONNRESET → `...Disconnected`, EIO/EEXIST/EBADMSG/ESRCH/EADDRINUSE/ENOBUFS/EOPNOTSUPP → their well-known names
- named-but-unmapped errnos → `System.Error.<NAME>` (ENXIO, EBUSY, ENODATA, ENOTCONN, EBADR, EHOSTUNREACH)
- unknown values → `org.freedesktop.DBus.Error.Failed` with `Unknown error N`
- negated errnos take the absolute value (so the ubiquitous `createError(-1, ...)` behaves as EPERM, exactly as it does on native)
- message format matches native: `"$customMsg (<strerror text>)"`

**New test**: `JvmCreateErrorParityTest` (jvmTest, same module so it can call the internal `createError`) pins 14 representative outputs to the native-format expectations.

## Verification

- `ktlintFormat` clean; **`apiDump` is a no-op** (`git diff -- api/` empty); `ktlintCheck apiCheck` pass
- `dbus-run-session -- ./gradlew jvmTest` — pass (incl. all 3 new/restored tests)
- `dbus-run-session -- ./gradlew linuxX64Test` — pass (restored parity tests green on the native backend too; **5 consecutive clean runs** of both suites)
- `compileKotlinLinuxX64`, `:codegen:test`, `:plugin:test` — pass
- New bus-using tests live in `com.monkopedia.sdbus.integration` for ARM CI

Fixes #56

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
@@ -228,12 +228,14 @@ class FailurePathParityTest {
     // is time-bounded so a hang in the known native-teardown path surfaces as a deterministic
     // test failure rather than an indefinitely stuck suite.
     //
-    // NOTE: this intentionally does NOT assert that a *subsequent* call on the released proxy
-    // throws. The two backends diverge there (the native backend rejects the call; the pure-JVM
-    // backend currently still services it), so a "call-after-release fails" assertion is not a
-    // parity property today -- see the PR description.
+    // This also asserts the sub-case that was scoped down when this suite landed (PR #54) and
+    // restored by the #56 decision: a *subsequent* call on the released proxy connection must
+    // throw on BOTH backends, never be silently serviced. The native backend guards every
+    // post-release operation with require(!released) { "Connection has already been released" }
+    // (ConnectionImpl.checkNotReleased); the JVM backend now guards its in-process dispatch
+    // short circuit with the same exception type and message.
     @Test
-    fun proxyConnectionTeardown_completesCleanly() = runBlocking {
+    fun proxyConnectionTeardown_completesCleanly_andCallAfterReleaseFails() = runBlocking {
         val ids = uniqueFixtureIds("teardownClean")
         val serverConnection = createBusConnection(ids.service)
         val proxyConnection = createBusConnection()
@@ -262,8 +264,19 @@ class FailurePathParityTest {
             withTimeout(5_000) {
                 proxyConnection.stopEventLoop()
             }
-            proxy.release()
             proxyConnection.release()
+
+            // A call after the proxy's connection has been released must throw on both
+            // backends, not be serviced. The per-call timeout bounds the failure mode where
+            // a regressed backend would instead try to perform the call.
+            val thrown = assertFailsWith<IllegalArgumentException> {
+                proxy.callMethod<Int>(ids.iface, MethodName("Echo")) {
+                    call(6)
+                    timeout = 2_000.milliseconds
+                }
+            }
+            assertEquals("Connection has already been released", thrown.message)
+            proxy.release()
         } finally {
             withTimeout(5_000) {
                 serverConnection.stopEventLoop()
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/TypeMatrixRoundTripTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/TypeMatrixRoundTripTest.kt
@@ -1,5 +1,6 @@
 package com.monkopedia.sdbus.integration
 
+import com.monkopedia.sdbus.Error
 import com.monkopedia.sdbus.InterfaceName
 import com.monkopedia.sdbus.MethodName
 import com.monkopedia.sdbus.ObjectPath
@@ -15,6 +16,7 @@ import com.monkopedia.sdbus.method
 import kotlin.random.Random
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
 import kotlin.test.assertTrue
 import kotlinx.serialization.Serializable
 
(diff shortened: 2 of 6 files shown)


monkopedia-reviewer (approved):
Approved as the final PR of the pre-approved ABI wave, implementing the owner decisions on #56 and #40 (2026-06-10) exactly: three behavior-only JVM-parity fixes with native as the reference, no public API/ABI change.

Scrutiny performed:
- **No ABI change**: changed-files list contains no api/*.api; diff vs merge base of api/ is empty; static-checks (ktlintCheck + apiCheck) green.
- **Fix 1 (strict reply deserialization)**: enforcement extends to top-level type class for 'a'/'('/'v' only, gated on actualSig != null. Verified it cannot reject previously-valid replies: in-process struct replies are data-class instances (inference returns null, enforcement skipped), primitive arrays infer null, empty list/dict infer 'a'/'a{}' which the lenient startsWith check accepts, and the real-dbus-java struct-into-data-class path was already non-functional on main (unchecked erased cast). The lenient-nested rationale is sound given value-based inference. Legitimate container paths covered by 35 TypeMatrixRoundTripTest + 23 EmptyCollectionWireTest cases on both backends, all green.
- **Fix 2 (call-after-release)**: confirmed against nativeMain ConnectionImpl.checkNotReleased — require(!released) { "Connection has already been released" } — exact exception type (IllegalArgumentException) and message match. Guard placed at top of both PureJavaDbusProxy.callMethod overloads; release() is idempotent via compareAndSet with no new throw on first or repeated release. The secondary ENOTCONN path only fires when AbstractConnection.isConnected is false (dropped/disconnected bus), mirroring sd_bus_call on a closed bus; cannot fire during normal operation.
- **Fix 3 (errno names)**: independently verified ALL 20 table entries plus abs(-1)/abs(-22), unknown-errno (4095 -> org.freedesktop.DBus.Error.Failed / 'Unknown error 4095'), and errno-0 against this machine's libsystemd sd_bus_error_set_errno — every name and strerror text matches the PR table exactly (EINVAL->InvalidArgs, EPERM/EACCES->AccessDenied, ENOMEM->NoMemory, ETIMEDOUT->Timeout, ENOENT->FileNotFound, System.Error.* fallbacks, etc.). Message format matches native createError ('customMsg (strerror)'). No test or source anywhere referenced the old jvm.errno.* names.
- **Restored tests**: both in com.monkopedia.sdbus.integration (ARM CI constraint), commonTest so they assert on BOTH backends, and they genuinely restore the #29/#54-dropped sub-cases (TypeMatrix int-into-struct mismatch asserting System.Error.ENXIO; FailurePathParity call-after-release asserting the exact native IllegalArgumentException contract).

CI: all checks green on first run, no flake reruns needed (arm-runtime-tests, full-tests-x64, native-build-x64, static-checks, all 9 pr-smoke matrix legs pass; nightly-stress skipped as expected). The new failure-path tests passed on both backends in CI.

Faithful to the owner decisions on #56/#40; no deviation found. Auto-merging on green per owner authorization.

Issue #74, opened 2026-06-11
JVM backend cannot deserialize multi-out (grouped) method replies from a real remote peer


monkopedia-coder:
Surfaced by the #34 Secret Service independent-peer suite, which gates the affected sub-cases behind a `peerGroupedReturnSupported` expect/actual flag rather than ship red tests. Found 2026-06-11. Sibling of #71 / #72 (same family: only manifests against a REAL remote peer; in-process own-server calls short-circuit through `JvmStaticDispatch` and never hit the wire deserialization path).

## Symptom

Against a real remote peer (python-dbusmock), **any method with 2+ out-args consumed as a grouped return** (`isGroupedReturn = true`, i.e. exactly what the `:codegen`-generated proxies emit for multi-out D-Bus methods) fails on the JVM backend with a signature-mismatch error, e.g. from the Secret Service shapes:

- `OpenSession(sv → vo)`: `System.Error.ENXIO: Message.deserialize failed: signature mismatch expected=(vo) actual=vs`
- `SearchItems(a{ss} → ao ao)`: `... expected=(aoao) actual=ao`

Native is unaffected (all grouped returns in the same suite round-trip fine).

## Mechanism

`Message.deserialize` on JVM (`src/jvmMain/kotlin/com/monkopedia/sdbus/Message.jvm.kt`) consumes exactly **one** payload value (`nextDeserializedValue`) and validates it by inferring a signature from the runtime value. For a grouped return the expected signature is the degrouped *tuple* of all out-args (e.g. `(vo)`), but only the first reply argument is ever read — so the inferred signature is just that first argument's (`vs` = a variant wrapping a string, `ao` = the first array) and validation correctly rejects the mismatch. The remaining reply arguments are never consumed; the structural decoding that `KSerializer.maybeDegrouped(...)` implies never happens on the JVM value-based path.

## Impact

Every generated proxy method for a D-Bus method with more than one out-arg (a very common real-world shape: Secret Service `OpenSession`/`SearchItems`/`Unlock`/`Lock`/`CreateItem`, BlueZ pairing APIs, etc.) is unusable on the JVM backend against a real service.

## Where to look

- `Message.jvm.kt` `deserialize` — needs a path that consumes *all* remaining reply args when the (degrouped) serializer's descriptor is a structure spanning multiple top-level values, rather than one `nextDeserializedValue()`.
[...]

Issue #72, opened 2026-06-11
JVM backend discards foreign D-Bus error names: every remote error reply surfaces as org.freedesktop.DBus.Error.AccessDenied


monkopedia-coder:
Found by the issue #36 dbusmock independent-peer suite (`DbusmockForeignErrorTest` in `:cross_test`).

## Symptom

On the JVM backend, when a **remote** (out-of-process) peer replies to a method call with a D-Bus error, the resulting `com.monkopedia.sdbus.Error` always has:

- `name = "org.freedesktop.DBus.Error.AccessDenied"`
- `errorMessage` prefixed with `"Operation not permitted"` — the real error name/message from the wire is at best buried in the appended custom text.

The actual error name sent by the peer (`org.freedesktop.DBus.Error.NotSupported`, `org.freedesktop.DBus.Error.UnknownMethod`, arbitrary custom names like `com.example.SomethingWentWrong`, dbusmock's `<iface>.UnknownProperty`, …) is discarded. The native sd-bus backend preserves name and message verbatim (verified by the same commonTest suite passing ungated on `linuxX64Test`).

## Root cause

`PureJavaDbusBackend` (jvmMain) converts remote error replies via the errno-style `createError(-1, message)`:

```kotlin
if (rawReply is org.freedesktop.dbus.messages.Error) {
    val remoteError = runCatching {
        rawReply.throwException()
        null
    }.exceptionOrNull()
    throw createError(-1, remoteError?.message ?: "Remote method call failed")
}
```

`createError` maps the errno through the EPERM entry of the errno table in `Error.jvm.kt`, producing `AccessDenied` regardless of the wire error. The wire error name is available on the dbus-java `Error` message (its header `ERROR_NAME` field / `getName()`); it should be propagated into `Error(name, errorMessage)` directly instead of going through the errno mapping. The same applies to:

- the async call path (`callMethodAsync` — same conversion),
- property `Get`/`Set`/`GetAll` errors from remote peers,
- timeouts vs. error replies (only genuine local failures should use `createError`).

Own-server JVM tests miss this because in-process calls short-circuit through `JvmStaticDispatch`, which rethrows the original `Error` object without ever serializing it to the wire — this is exactly the class of gap the independent-peer suite (#32/#35/#36) exists to catch, and the foreign-error half of the #63 parity work that hadn't been validated against a non-sdbus peer before.

## Where it's gated

[...]

Issue #71, opened 2026-06-11
JVM backend cannot marshal @Serializable structs to/from a real remote peer (both directions)


monkopedia-coder:
Surfaced by the #35 independent-peer type-matrix suite (PR #70), which gated the affected sub-cases out behind `peerStructMarshallingSupported` rather than ship red tests. Found 2026-06-11.

## Symptom
Against a REAL remote peer (python-dbusmock), custom `@Serializable` struct types fail on the JVM backend in **both directions**:
- **Send**: dbus-java throws "Trying to marshall to unconvertible type" — `Message.serialize` appends the raw Kotlin object and `toJavaSignalValue()` has no struct decomposition path.
- **Receive**: signature validation rejects with e.g. `expected=(is) actual=ai` — value-based signature inference treats a decoded struct as an array.

Native is unaffected (structs round-trip fine, verified in the same suite).

## Why it was invisible
All prior JVM struct coverage ran against an in-process own-server, where `JvmStaticDispatch` short-circuits the wire marshalling layer entirely. Cross-runtime tests (cross_test) exercise JVM↔native but apparently not custom-struct shapes in the failing direction. The independent-peer suite is the first thing to push real struct bytes through dbus-java.

## Where to look
- `src/jvmMain/.../Message.jvm.kt` serialize path + `toJavaSignalValue()` (struct decomposition missing → dbus-java needs structs as `org.freedesktop.dbus.Struct` subclasses or positional Object[] with explicit signature)
- The JVM signature-inference path that yields `ai` for a decoded struct (relates to the #63 strict-deserialization machinery — the strictness is correct, the inference feeding it is wrong for structs)
- Re-enable the gated sub-cases in `cross_test/.../DbusmockTypeMatrixTest.kt` (flip `peerStructMarshallingSupported` for JVM) as the acceptance test.

## Gate note
Candidate addition to the 0.5.0 release gate (a real wire-protocol bug found by gate work; fix is behavior-only, no API change). Owner can veto inclusion.

monkopedia-coder:
[...]

PR #70, opened 2026-06-11
dbusmock coverage: independent-peer type-matrix & marshalling edge cases

monkopedia-coder:
Adds `DbusmockTypeMatrixTest` to `:cross_test` commonTest — the independent-peer counterpart to the own-server type/marshalling tests (#25/#26/#27). Every case drives our client against a generic python-dbusmock object (a foreign Python/GDBus encoder/decoder), so symmetric serializer bugs (encode-wrong + decode-wrong in matching ways) cannot hide the way they can in own-server round-trips. Builds on the #32 harness and follows its gating pattern exactly: clean SKIP when python3-dbusmock is absent (e.g. ARM CI), real runs in `full-tests-x64`.

Fixes #35.

## Coverage (12 tests, ~80 sub-case assertions, identical on both backends)

- **Basic types with extremes**: y/b/n/q/i/u/x/t (MIN/MAX incl. `ULong.MAX_VALUE` across the signed boundary), d with NaN/±Inf/-0.0/1e±300, strings (empty/unicode/multiline/quotes), object paths, signatures incl. the empty signature
- **Variants**: primitives, ULong.MAX, arrays, dicts, object paths, structs (gated, see below)
- **Arrays**: ai/as/ay/ad/at/ab/ao/aai/av — all including empty
- **Dicts**: a{ss}/a{ix}/a{yu}/a{sv}/a{ias} incl. empty, plus the GetManagedObjects-shaped `a{oa{sa{sv}}}` with empty inner dict
- **Structs**: (is), wide all-basic-type struct `(ybnqiuxtdsogv)`, nested `((is)ai)`, `a(is)` incl. empty — plus receive-direction cases where the *peer* constructs the struct
- **Large payloads**: 128 KiB `ay`, 2000-element `as`, 64 KiB string
- **Argument shapes**: multi-arg `iis`, mixed unsigned `tu`, no-arg, no-return, container-in/container-out
- **Async**: suspend `callMethodAsync` round-trips
- **Properties**: complex-typed property reads + `GetAll` via `PropertiesProxy.INTERFACE_NAME`
- **Unix-fd passing** (`h`): functional round-trip on native — a pipe read-end is echoed by dbusmock and the returned fd is proven to be a live duplicate by writing through the original write end. Skipped on JVM, where test code cannot surface a raw pipe fd int (java.io.FileDescriptor is sealed behind JPMS).

[...]

PR #75, opened 2026-06-11
Add dbusmock Secret Service (org.freedesktop.secrets) round-trip suite

### LINKED ISSUE #34: dbusmock coverage: Secret Service (org.freedesktop.secrets) round-trips
Coverage against the **Secret Service** API (`org.freedesktop.secrets`) via the dbusmock `secrets` template — a second, well-specified real-world D-Bus interface family distinct from BlueZ.

Exercise collections/items/secrets round-trips, session setup, property reads, and the `a{sv}`/struct shapes the spec uses — on both backends. Validates our ObjectManager/Properties/method-call handling against a different independent peer and a different type profile than BlueZ.

Blocked on #32 (dbusmock harness).

monkopedia-coder:
Adds the Secret Service (`org.freedesktop.secrets`) independent-peer coverage suite (issue #34) — the second real-world interface family after BlueZ, with a distinct type profile: `(oayays)` secret structs, `a{sv}` creation properties, `a{ss}` attributes, `ao` object trees, and multi-out methods.

Fixes #34.

## What's covered (`cross_test/.../DbusmockSecretServiceTest.kt`, commonTest — same assertions on native sd-bus and JVM dbus-java)

| Flow | native | JVM |
|---|---|---|
| `OpenSession` (plain) `(v,o)` reply + session `Close` lifecycle | ✅ | gated (#74) |
| Collection discovery: `Collections` property, `ReadAlias("default")`, `Label`/`Locked`/`Created`/`Modified` | ✅ | ✅ |
| `CreateItem` → `GetSecret`/`SetSecret` → `GetSecrets` `(oayays)` round-trips (secret re-targeted at the requesting session) | ✅ | gated (#71 + #74) |
| Item `Label`/`Attributes`/`Locked` reads + `Label` write via `Properties.Set` | ✅ | ✅ |
| `SearchItems`: collection-level (`ao`) | ✅ | ✅ |
| `SearchItems`: service-level (`ao` unlocked, `ao` locked split) | ✅ | gated (#74) |
| `ItemCreated`/`ItemDeleted` signals + `Items` property convergence + post-delete failure | ✅ | ✅ |
| Locked-collection error paths (`org.freedesktop.Secret.Error.IsLocked`) | ✅ | ✅ (name assertion gated, #72) |
| `Unlock` → use → `Lock` round-trip, cascading `Locked` to items | ✅ | gated (#74) |

## How

- The installed python-dbusmock (0.38.1) ships **no secrets template**, so the `org.freedesktop.Secret.*` object tree is scripted onto a generic mock via the `org.freedesktop.DBus.Mock` control interface, following the merged sibling pattern (PR #70/#73). All control-plane calls use string/dict arguments only, so the same setup runs from the JVM backend; object creation happens inside scripted python (`self.AddObject`).
- Client side goes through **hand-written typed proxies in the exact style `:codegen` emits** (PropertyDelegate-backed properties, suspend `callMethodAsync`, `isGroupedReturn` for multi-out) — validating the API the way a real adopter consumes it.
- Shared fixture (`DbusmockPeerFixture.kt`) gains optional `mockPath`/`mockInterface` overrides so the peer sits at the realistic `/org/freedesktop/secrets` path with main interface `org.freedesktop.Secret.Service`; bus names stay unique per test, existing call sites unchanged.

## New library bug found (gated, not fixed here)

**Issue #74 — JVM backend cannot deserialize multi-out (grouped) method replies from a real remote peer.** `Message.deserialize` on JVM consumes one payload value and validates it against the degrouped tuple signature, e.g. `OpenSession` fails `expected=(vo) actual=vs`, service `SearchItems` fails `expected=(aoao) actual=ao`. Every codegen-generated proxy method with 2+ out-args is affected against a real service; own-server tests miss it because JvmStaticDispatch bypasses wire deserialization. Gated behind a new `peerGroupedReturnSupported` expect/actual (sibling of the #71/#72 flags) with KDoc pointing at the issue; flip the JVM actual to `true` when fixed to enforce parity.

Known bugs #71 (struct marshalling) and #72 (foreign error names) are gated with the existing flags and referenced inline — not re-tripped.

## Verification

- `ktlintFormat` clean, `apiDump` no-op, `ktlintCheck apiCheck` pass (test-only change, no src/main edits)
- `dbus-run-session -- ./gradlew :cross_test:jvmTest :cross_test:linuxX64Test` (with `DBUSMOCK_PYTHON` pointing at the dbusmock venv): **51/51 pass on each backend** (45 existing + 6 new); native runs every sub-case ungated, JVM prints the documented SKIP lines
- `compileKotlinLinuxX64 :cross_test:compileTestKotlinLinuxX64` pass
- Suite skips cleanly when python3-dbusmock is unavailable (ARM CI)

### DIFF
--- a/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockPeerFixture.kt
+++ b/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockPeerFixture.kt
@@ -143,17 +143,25 @@ internal suspend fun <T> pumpUntilSubscribed(received: ReceiveChannel<T>, poke: 
  * @param suffix Distinguishes the peer's bus name / object path between tests.
  * @param objectManager When `true`, the peer also implements
  *   `org.freedesktop.DBus.ObjectManager` at [DbusmockPeer.objectPath] (dbusmock `-m`).
+ * @param mockPath Overrides the mock's root object path (defaults to a unique per-test path).
+ *   Object paths are scoped to the peer's connection, so a fixed well-known path (e.g.
+ *   `/org/freedesktop/secrets`) cannot collide between tests — the bus name stays unique.
+ * @param mockInterface Overrides the mock's main interface name (defaults to the unique bus
+ *   name), for mocking well-known interfaces like `org.freedesktop.Secret.Service`.
  */
 internal fun withDbusmockPeer(
     suffix: String,
     objectManager: Boolean = false,
+    mockPath: String? = null,
+    mockInterface: String? = null,
     block: suspend DbusmockPeer.() -> Unit
 ) = runBlocking {
     val id = Random.nextInt(100_000, 999_999)
     val busName = "com.monkopedia.sdbus.dbusmock.$suffix$id"
-    val objectPath = "/com/monkopedia/sdbus/dbusmock/$suffix$id"
+    val objectPath = mockPath ?: "/com/monkopedia/sdbus/dbusmock/$suffix$id"
+    val interfaceName = mockInterface ?: busName
 
-    val handle = launchDbusmock(busName, objectPath, busName, objectManager)
+    val handle = launchDbusmock(busName, objectPath, interfaceName, objectManager)
     if (handle == null) {
         println(
             "[withDbusmockPeer] SKIP: python-dbusmock unavailable. " +
@@ -173,7 +181,7 @@ internal fun withDbusmockPeer(
         proxy,
         ServiceName(busName),
         ObjectPath(objectPath),
-        InterfaceName(busName)
+        InterfaceName(interfaceName)
     )
 
     try {
--- a/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockSecretServiceTest.kt
+++ b/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockSecretServiceTest.kt
@@ -1,0 +1,966 @@
+/**
+ *
+ * (C) 2024 - 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Project: sdbus-kotlin
+ * Description: High-level D-Bus IPC kotlin library based on sd-bus
+ *
+ * This file is part of sdbus-kotlin.
+ *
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard — test-only (tier 1), approving.

**Coverage vs #34:** comprehensive. OpenSession session lifecycle, collection discovery (Collections property + ReadAlias), CreateItem/GetSecret/SetSecret/GetSecrets (oayays) struct round-trips, item properties via the Properties interface (incl. writable Label through Properties.Set), collection- and service-level SearchItems, ItemCreated/ItemDeleted signals with event-driven waits, IsLocked error paths, and the Unlock/Lock round-trip with cascading item lock state. The installed dbusmock ships no secrets template, so scripting the org.freedesktop.Secret.* tree onto a generic mock via the DBus.Mock control interface (string/dict args only, JVM-drivable) is the right call and follows the merged #70/#73 pattern. Client side via hand-written typed proxies in codegen style validates the real-adopter shape, including isGroupedReturn multi-out consumption.

**Known-bug gating:** clean. #71 (peerStructMarshallingSupported) and #72 (peerErrorNameMappingSupported) reuse the existing flags with inline issue references; the newly found JVM grouped-return deserialization gap is properly ticketed as #74 with mechanism, impact, code pointers, and an acceptance path (flip the new peerGroupedReturnSupported JVM actual to true), and every gated sub-case prints a documented SKIP rather than shipping red. assertFailsWith still runs unconditionally on JVM where only the error-name assertion is gated — good parity discipline.

**Mechanics:** all four touched files are cross_test test sources; no src/main changes, api/ dumps untouched. The DbusmockPeerFixture extension (optional mockPath/mockInterface, defaulting to the previous behavior) leaves existing call sites unchanged; the connection-scoped-object-path rationale for the fixed /org/freedesktop/secrets path is correct. expect/actual covers both cross_test targets (jvm, linuxX64). Eager onSignal registration matches the DbusmockObjectManagerTest sibling (pumpUntilSubscribed is only needed for lazy Flow listeners), and same-connection message ordering makes the signal waits race-free.

CI: all checks green on first run (no flake re-runs needed).

PR #78, opened 2026-06-11
JVM wire fixes: struct marshalling, foreign error names, grouped replies (#71, #72, #74)

### LINKED ISSUE #71: JVM backend cannot marshal @Serializable structs to/from a real remote peer (both directions)

### LINKED ISSUE #72: JVM backend discards foreign D-Bus error names: every remote error reply surfaces as org.freedesktop.DBus.Error.AccessDenied

### LINKED ISSUE #74: JVM backend cannot deserialize multi-out (grouped) method replies from a real remote peer

monkopedia-coder:
Fixes the three JVM backend wire-protocol bugs found by the independent-peer (dbusmock) suites, with native sd-bus behavior as the reference. All three only manifest against a real remote peer: in-process own-server calls short-circuit through `JvmStaticDispatch` and never hit the wire conversion paths.

## #71 — struct marshalling to/from real peers

**Approach:** a new `JvmValueCodec.jvm.kt` introduces a `JvmValueEncoder`/`JvmValueDecoder` pair — the JVM counterpart of the native `MessageEncoder`/`MessageDecoder` — that encodes `@Serializable` values into / decodes them from the JVM backend's payload value tree. A D-Bus struct is represented by a new `Message.JvmStructPayload(signature, fields)` (mirroring how `JvmVariantPayload` represents variants).

- **Send path:** `Message.serialize` decomposes struct-containing types into wire-shaped values; `toJavaSignalValue` converts `JvmStructPayload` to the positional `Object[]` dbus-java's `Message.appendOne` STRUCT case expects (previously: "Trying to marshall to unconvertible type").
- **Receive path:** a new signature-aware `fromJavaWireValue` conversion uses the wire signature (the only way to tell a struct apart from an array once dbus-java has extracted it to `Object[]`/`List`) to rebuild `JvmStructPayload` from reply bodies, signal bodies, and variant contents. The #63 strict-signature checks are preserved — mismatched payloads still fail with the same ENXIO signature-mismatch error (previously structs inferred as arrays and were *correctly* rejected: `expected=(is) actual=ai`).

## #72 — foreign error names discarded

Remote error replies now construct `Error(wireName, wireMessage)` verbatim from the dbus-java `Error` message's ERROR_NAME header (`getName()`) and first body argument — matching sd-bus, which copies both straight into `sd_bus_error`. Previously everything was routed through the errno-based `createError(-1, msg)` and surfaced as `org.freedesktop.DBus.Error.AccessDenied`. Sync, async (delegates to the sync path and rethrows the `Error` unchanged), and remote property Get/Set/GetAll calls all flow through the same conversion. **Local errno-derived `createError` behavior is unchanged** — `JvmCreateErrorParityTest` (14 tests) pins it and passes.

## #74 — grouped multi-out replies

`Message.deserialize` previously consumed exactly one payload value, so a degrouped tuple expectation like `(vo)` was validated against just the first reply arg (`vs`) and rejected. It now routes struct-signature types through structured decoding (`deserializeStructured` → `JvmValueDecoder`), which consumes one payload value per out-arg through the common degrouping machinery in `DegroupingReturns.kt` (a degrouped serializer bypasses the outer `beginStructure`, pulling fields directly from the message payload). Single-struct replies and struct-containing containers decode through the same path; raw in-process values fall through to the legacy coercion path.

## Inherited work note

This PR completes work started by a prior interrupted agent run. The inherited `JvmValueCodec` approach was reviewed and kept as-is — it compiled cleanly, mirrors the native encoder/decoder architecture, and covers all conversion paths (method replies, signals, variants, property values, sync/async). Verification below was run fresh.

## Acceptance evidence

All three JVM gate flags flipped to `true` in `cross_test/src/jvmTest/.../DbusmockFdSupport.jvm.kt`, un-gating:

- `DbusmockTypeMatrixTest` struct round-trip sub-cases (12 tests, #71)
- `DbusmockForeignErrorTest` (5 tests, #72) + `DbusmockBluezTest` `org.bluez.Error.*` name assertions (7 tests)
- `DbusmockSecretServiceTest` (6 tests; gates on all three flags — `(vo)`, `(aoao)`, `(aoo)` grouped replies, struct secrets, error names)

Full suite, fresh runs (`dbus-run-session`, JAVA_HOME=java-21-openjdk):

| Suite | Result |
|---|---|
| `jvmTest` | 251 passed, 0 failed, 0 skipped |
| `linuxX64Test` | 268 passed, 0 failed, 0 skipped |
| `:cross_test:jvmTest` | 58 passed (dbusmock 0.38.1 **and** CI's 0.31.1) |
| `:cross_test:linuxX64Test` | 58 passed (both dbusmock versions) |
| `:codegen:test`, `:plugin:test`, `compileKotlinLinuxX64`, `:stress_test:compileTestKotlinJvm` | green |
| `ktlintCheck`, `apiCheck` | green; `apiDump` no-op (no public API change) |

#63 parity preserved: strict deserialization checks and `JvmCreateErrorParityTest` pass unchanged.

Fixes #71. Fixes #72. Fixes #74.

### DIFF
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/DbusmockFdSupport.jvm.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/DbusmockFdSupport.jvm.kt
@@ -21,10 +21,10 @@
 package com.monkopedia.sdbus.integration
 
 /**
- * See the expect declaration: the JVM backend cannot yet marshal custom @Serializable struct
- * values to or from remote peers, so the remote-struct sub-cases are skipped on this backend.
+ * Struct marshalling to/from remote peers is supported since the issue #71 fix (structs are
+ * decomposed into wire-shaped values via JvmValueCodec on the way to/from dbus-java).
  */
-internal actual val peerStructMarshallingSupported: Boolean = false
+internal actual val peerStructMarshallingSupported: Boolean = true
 
 /**
  * The JVM has no portable way to surface a freshly created pipe as a raw integer file
@@ -44,14 +44,14 @@ internal actual fun closeTestFd(fd: Int) {
 }
 
 /**
- * See the expect declaration (DbusmockForeignErrorTest.kt): the JVM backend discards foreign
- * error names (issue #72). Flip to `true` when that bug is fixed.
+ * Foreign error names/messages are preserved verbatim since the issue #72 fix (remote error
+ * replies construct Error(wireName, wireMessage) instead of going through the errno mapping).
  */
-internal actual val peerErrorNameMappingSupported: Boolean = false
+internal actual val peerErrorNameMappingSupported: Boolean = true
 
 /**
- * See the expect declaration (DbusmockSecretServiceTest.kt): the JVM backend cannot
- * deserialize multi-out (grouped) method replies from a real remote peer (issue #74). Flip
- * to `true` when that bug is fixed.
+ * Multi-out (grouped) replies from remote peers are supported since the issue #74 fix
+ * (structured deserialization consumes one reply value per out-arg via the common
+ * degrouping machinery).
  */
-internal actual val peerGroupedReturnSupported: Boolean = false
+internal actual val peerGroupedReturnSupported: Boolean = true
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/JvmValueCodec.jvm.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/JvmValueCodec.jvm.kt
@@ -1,0 +1,387 @@
+package com.monkopedia.sdbus
+
+import kotlinx.serialization.DeserializationStrategy
+import kotlinx.serialization.SerializationStrategy
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.StructureKind
+import kotlinx.serialization.encoding.AbstractDecoder
+import kotlinx.serialization.encoding.AbstractEncoder
+import kotlinx.serialization.encoding.CompositeDecoder
+import kotlinx.serialization.encoding.CompositeEncoder
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.modules.SerializersModule
+
+// JVM counterpart of the native MessageEncoder/MessageDecoder pair (nativeMain/internal):
+// instead of writing into an sd-bus message, values are encoded into/decoded from the JVM
+// backend's payload value tree, where a D-Bus struct is represented by
+// [Message.JvmStructPayload] (mirroring how [Message.JvmVariantPayload] represents variants).
+// This is what allows @Serializable struct types to survive the trip through the dbus-java
+// wire layer, which needs structs decomposed into positional values (issue #71), and what
+// allows multi-out (grouped) replies to consume one payload value per out-arg (issue #74,
+// via the common degrouping machinery in DegroupingReturns.kt, which bypasses
+// beginStructure/endStructure on this encoder/decoder).
+
+private val unsignedNames = listOf(
+    "UInt",
+    "ULong",
+    "UByte",
+    "UShort"
+)
+
+/**
+ * Encodes [arg] into the JVM payload value tree using its serializer: primitives and the
+ * sdbus value types stay as themselves, lists/maps become List/Map, and struct-kind classes
+ * become [Message.JvmStructPayload] holding their decomposed field values. A degrouped
+ * serializer (multi-out reply) produces one top-level value per field.
+ */
+internal fun <T> encodeToJvmValues(
+    serializer: SerializationStrategy<T>,
+    module: SerializersModule,
+    arg: T
+): List<Any?> {
+    val values = mutableListOf<Any?>()
+    JvmValueEncoder(module, values::add).encodeSerializableValue(serializer, arg)
+    return values
+}
+
+internal class JvmValueEncoder(
+    override val serializersModule: SerializersModule,
+    private val sink: (Any?) -> Unit,
+    private val onEnd: () -> Unit = {}
+) : AbstractEncoder() {
+
+    private var lastInline: SerialDescriptor? = null
+    private val isUnsigned: Boolean
+        get() = (lastInline?.serialName?.removePrefix("kotlin.") in unsignedNames).also {
+            lastInline = null
+        }
+
+    override fun encodeNotNullMark() {
+        error("Nullables are not allowed for serialization")
+    }
+
+    override fun encodeNull() {
+        error("Nullables are not allowed for serialization")
+    }
+
+    override fun encodeInline(descriptor: SerialDescriptor): Encoder {
+        lastInline = descriptor
+        return this
+    }
+
+    override fun beginStructure(descriptor: SerialDescriptor): CompositeEncoder =
+        when (descriptor.kind) {
+            StructureKind.LIST -> {
+                val items = mutableListOf<Any?>()
+                JvmValueEncoder(serializersModule, items::add) { sink(items.toList()) }
+            }
+
+            StructureKind.MAP -> {
+                val entries = mutableListOf<Any?>()
+                JvmValueEncoder(serializersModule, entries::add) {
+                    val map = LinkedHashMap<Any?, Any?>()
+                    var index = 0
+                    while (index + 1 < entries.size) {
+                        map[entries[index]] = entries[index + 1]
+                        index += 2
+                    }
+                    sink(map)
+                }
+            }
+
+            StructureKind.CLASS,
+            StructureKind.OBJECT -> {
+                val signature = descriptor.asSignature.value
+                val fields = mutableListOf<Any?>()
+                JvmValueEncoder(serializersModule, fields::add) {
+                    sink(Message.JvmStructPayload(signature, fields.toList()))
+                }
+            }
+
+            else -> error("Unsupported structure kind ${descriptor.kind} for serialization")
+        }
+
+    override fun endStructure(descriptor: SerialDescriptor) = onEnd()
+
+    override fun encodeBoolean(value: Boolean) {
+        lastInline = null
+        sink(value)
+    }
+
+    // D-Bus has no signed byte; BYTE-kind values travel as 'y' (see TypeTraits signatureOf),
+    // matching the native encoder which always appends bytes as UByte.
+    override fun encodeByte(value: Byte) {
+        lastInline = null
+        sink(value.toUByte())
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (approved):
Reviewed to full tier-1 standard (with extra rigor given +605/−30 through the JVM backend's core wire path). Classification: bug_fix (behavior-only, no API change) → tier 1 per policy defaults for sdbus-kotlin.

**Correctness findings per scrutiny area:**

- **JvmValueCodec round-trip fidelity:** the encoder/decoder pair mirrors the native MessageEncoder/MessageDecoder exactly — Byte/Char encode as UByte ('y'), Float widens to Double ('d'), enums encode as Int index (matching TypeTraits' ENUM→IntSig), unsigned types tracked via the same encodeInline/lastInline pattern as nativeMain. ObjectPath/Signature inline strings, Variant/UnixFd/Unit special cases, nested structs, structs-in-variants, a(is) arrays, and dicts all traced through both directions. Decoder tolerates dbus-java's signed-primitive widening (Number-based slot coercion), consistent with the legacy coerceForDescriptor behavior.
- **Wire signature is the sole struct discriminator:** fromJavaWireValue only manufactures JvmStructPayload when the wire signature says "(...)"; the value-shape fallback never invents a struct, and JvmValueDecoder's CLASS branch requires JvmStructPayload — so an ai array can never decode as a struct nor vice versa. Verified the #56/#63 strictness test (mismatchedReply_intIntoStruct: i into (ii)) still ENXIOs through the new path: singleOk=false (not a struct payload), groupedOk=false (1 value < 2 element sigs).
- **#63 strictness preserved:** legacy deserialize path untouched; the structured path enforces the same isSignatureCompatible/shouldEnforceSignature checks *before* consuming values and produces the identical "System.Error.ENXIO: ... signature mismatch expected=X actual=Y" shape. JvmCreateErrorParityTest and the strict-deserialization tests are untouched and pass on CI. Local createError paths unchanged — only remote Error replies get verbatim wire names (toSdbusError), with safe handling of empty-body errors (runCatching + firstOrNull + orEmpty) and blank/missing ERROR_NAME (falls back to createError(-1)). Timeouts remain local createError, per the issue spec.
- **In-process dispatch:** JvmStaticDispatch arg-count matching still lines up (one struct arg = one JvmStructPayload = one top-level type), serialize/deserialize are symmetric so JvmStructPayload round-trips in-process, and raw handler results (GetManagedObjects/Properties dispatch) fall through to the legacy path via the firstSig==null / containsJvmStructPayload guards. Empty a(is) collections keep the declared-signature send path (#11/#26) and the legacy receive path.
- **#74 grouped replies:** degrouped serializers bypass beginStructure and pull one payload value per out-arg; the size>1 signature degrouping in serialize handles nested cases like ((is)u)→(is)u correctly; (vo)/(aoao)/(aoo) shapes traced end-to-end.
- **Scope:** 4 files, jvmMain + cross_test gates only — no api/ dumps touched (apiCheck green), no nativeMain changes.

**Acceptance evidence:** all three gate flags flipped to true; the previously-gated sub-cases in DbusmockTypeMatrixTest (struct receive/round-trip/variant-wrapped), DbusmockForeignErrorTest, DbusmockBluezTest error-name assertions, and DbusmockSecretServiceTest now run un-gated on JVM — and full-tests-x64 (CI's dbusmock 0.31.1) passes, alongside arm-runtime-tests, native-build-x64, static-checks, and all 9 pr-smoke matrix jobs.

Minor non-blocking notes: groupedOk's remaining.size >= elementSigs.size is lenient on extra reply args (consistent with sd-bus reading only what it needs), and the multi-value mismatch message now joins all inferred sigs — neither is asserted on anywhere.

Approving and merging.

Issue #66, opened 2026-06-11
Docs: JVM↔native backend parity matrix


monkopedia-coder:
Part of the 1.0 docs overhaul (owner-approved 2026-06-11).

Write the backend contract document: a table of what works identically on the JVM and native backends and where they differ, so adopters get a contract instead of surprises. Suggested home: docs/BACKENDS.md, linked prominently from the README.

Rows to cover (verify each against current main + tests — do not write from memory):
- Method calls (sync/suspend), timeouts, error propagation and error-name mapping (post-#63: errno→D-Bus names match; strict deserialization matches; call-after-release throws match)
- Properties: get/set, PropertiesChanged emission incl. emitPropertiesChangedSignal getter-resolution parity (#28/PR #47)
- Signals: emission, subscription, signalFlow
- ObjectManager: GetManagedObjects, InterfacesAdded/Removed
- Unix FD passing (UnixFd, dup vs adopt)
- Connection types: which of the 11 factories work on which backend (the fd-based ones are native-only — the JVM actuals throw @Deprecated(ERROR)-marked stubs; document this)
- Event loop semantics (startEventLoop/stopEventLoop on each backend)
- Known JVM limitations (see JvmSignalPropertyUnsupportedTest for documented gaps)

Sources of truth: src/commonTest integration suites (they run on both backends — anything asserted there is parity), the #56/#63 parity work, JvmSignalPropertyUnsupportedTest, and the per-backend actuals. Constraints: docs-only, apiDump no-op.

PR #85, opened 2026-06-11
JVM: apply connection methodCallTimeout; throw on unreachable bus instead of stub fallback (#80, #81)

### LINKED ISSUE #80: JVM: Connection.methodCallTimeout is stored but never applied to calls
Found by #66's parity-matrix verification (PR #79, documented in docs/BACKENDS.md). Native wires methodCallTimeout to sd_bus_set_method_call_timeout so it applies to calls with no explicit timeout; the JVM backend stores the value but no call path reads it — setting it silently does nothing. Fix direction (native = reference): apply the stored default in the JVM call paths when no per-call timeout is given. Behavior-only, no API change. Acceptance: a cross-backend test setting methodCallTimeout and asserting a short-timeout failure on both backends.

### LINKED ISSUE #81: JVM: silent in-process stub fallback when the bus is unreachable (native throws)
Found by #66's parity-matrix verification (PR #79, documented in docs/BACKENDS.md). When the session/system bus is unreachable, JVM connection factories silently fall back to StubJvmDbusBackend (unique name :jvm-stub) — the connection appears to succeed against a fake in-process bus, where native throws Error. This is the worst kind of silent divergence: code 'works' on JVM in CI containers with no bus and fails in production semantics. Fix direction (native = reference): real connection factories throw when the bus is unreachable; whatever tests rely on the stub get an explicit internal opt-in instead of a silent fallback. Behavior-only, no public API change expected — verify the stub's existing test usages (mockk suites) and preserve them via the explicit path.

monkopedia-coder:
Fixes #80. Fixes #81.

Two owner-approved JVM behavior fixes from the #66 parity-matrix verification (native = reference). Behavior-only: no public API change — `apiDump` is a no-op.

## #80 — `Connection.methodCallTimeout` now applied (JVM)

- `PureJavaDbusConnection` stored the `Duration` but no call path read it. `PureJavaDbusProxy` now receives a `defaultTimeoutMicros` provider from its connection, and `callMethod(message, timeout)` resolves the effective timeout exactly the way `sd_bus_call` resolves a 0 timeout via `sd_bus_get_method_call_timeout`:
  - explicit per-call timeout always wins;
  - the connection default only fills the unset (`0`/`Duration.ZERO` sentinel) case — `callMethod(message)` (no timeout arg) funnels through the same resolution;
  - if neither is set, the effective timeout stays 0 and the pre-existing default behavior (dbus-java's own reply timeout) is unchanged, mirroring native where an unset (ZERO) `methodCallTimeout` falls back to sd-bus's built-in default.
- All call paths funnel through the two sync `callMethod` overloads (callback-async, suspend, static-dispatch and remote-wire paths), so the default applies uniformly.

**Acceptance test (commonTest, runs on BOTH backends):** `FailurePathParityTest.connectionMethodCallTimeout_appliesToCallsWithoutExplicitTimeout` — reuses the slow-method fixture pattern: sets `proxyConnection.methodCallTimeout = 100.milliseconds`, calls a 600 ms-slow method WITHOUT a per-call timeout via the raw `callMethod(message)` overload, asserts a timeout-shaped `Error`; then asserts an explicit generous per-call timeout overrides the connection default (call succeeds).

## #81 — throw on unreachable bus instead of silent stub fallback (JVM)

- `PureJavaDbusBackend` no longer takes a fallback backend. When `tryCreateConnection` fails (or the request is unsatisfiable), `createConnection` throws `com.monkopedia.sdbus.Error` with a "Failed to open bus…" message, mirroring native's `openBus` ("Failed to open bus", errno from `sd_bus_open_*`). The exact D-Bus error *name* is not pinned across backends because dbus-java exposes no errno — the `Error` type and the throw are the contract.
- `JvmDbusBackendProvider.backend` is now `PureJavaDbusBackend()` with no implicit stub anywhere.

**Acceptance test (jvmTest):** `JvmUnreachableBusTest` — `createSessionBusConnection(address)` and `createDirectBusConnection(address)` pointed at a nonexistent unix socket path throw `Error` (never return a `:jvm-stub` connection). Plus `PureJavaDbusBackendTest.createConnection_throwsWhenEndpointMissing` for the unsatisfiable-request paths.

### How stub users were migrated

- `StubJvmDbusBackend` **stays** as an explicit internal opt-in for unit tests: tests construct it directly (`StubJvmDbusBackendTest` already does — unchanged) or mock `JvmDbusBackendProvider` (the mockk suites `JvmFactoryMockkTest`/`JvmConnectionMockkTest`/`JvmProxyObjectMockkTest` already do — unchanged). Its KDoc now documents that it is never selected implicitly.
- `PureJavaDbusBackendTest`: constructor lost the fallback param; `createConnection_usesFallbackWhenEndpointMissing` became `createConnection_throwsWhenEndpointMissing`.
- `JvmRealBusIntegrationTest`: the 7 `":jvm-stub"` unique-name skip gates (which used the silent fallback as a "no bus in this environment" signal) are migrated to a `connectOrNull` helper that treats the thrown `Error` as the skip signal. CI behavior is unchanged — `full-tests-x64` runs under `dbus-run-session`, so these tests run for real there.
- `JvmScaffoldTest.createBusConnection_hasStubUniqueName` renamed to `createBusConnection_hasUniqueName` (it asserts against the real bus under CI's `dbus-run-session`).

### BACKENDS.md row updates

- Summary matrix: `Connection.methodCallTimeout` row is now ✅/✅ ("JVM fixed in #80", pinned by the new commonTest); bus-unreachable row is now "✅ throws `Error`" on both ("JVM fixed in #81", stub = explicit internal test opt-in only).
- The two "Known divergence" sections are replaced with parity statements + proven-by test references; "JVM limitations at a glance" items 3/4 removed and recorded as closed historical gaps.

### Verification (all with `JAVA_HOME=/usr/lib/jvm/java-21-openjdk`)

- `./gradlew ktlintFormat` → `./gradlew apiDump` no-op (`git diff -- api/` empty) → `./gradlew ktlintCheck apiCheck` ✅
- `dbus-run-session -- ./gradlew jvmTest linuxX64Test :cross_test:jvmTest :cross_test:linuxX64Test` ✅ (FailurePathParityTest 7/7 on both backends incl. the new test; `JvmUnreachableBusTest` 2/2; `JvmCreateErrorParityTest`, strict-deserialization and dbusmock suites all green — no #63/#78 regressions)
- `./gradlew compileKotlinLinuxX64 :codegen:test :plugin:test :stress_test:compileTestKotlinJvm` ✅

### DIFF
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -24,7 +24,7 @@ Every row below is pinned by tests on current `main`, not by intention:
 | Struct (`@Serializable`) marshalling to/from remote peers | ✅ | ✅ | JVM fixed in #71 |
 | Multi-out (grouped) replies (`isGroupedReturn`) | ✅ | ✅ | JVM fixed in #74 |
 | Per-call timeout (`timeout = …` in the invoker block) | ✅ | ✅ | Timeout *error name* may differ, see below |
-| Connection-level `Connection.methodCallTimeout` | ⚠️ stored but not applied | ✅ applied | Known divergence, see below |
+| Connection-level `Connection.methodCallTimeout` | ✅ | ✅ | Applied as the default for calls without an explicit timeout; JVM fixed in #80 |
 | Error propagation: named D-Bus errors round-trip name + message verbatim | ✅ | ✅ | Incl. foreign (non-sdbus) peers; JVM fixed in #72 |
 | errno → D-Bus error-name mapping for locally created errors | ✅ | ✅ | JVM pinned to native output |
 | Call-after-release fails with `IllegalArgumentException("Connection has already been released")` | ✅ | ✅ | |
@@ -35,7 +35,7 @@ Every row below is pinned by tests on current `main`, not by intention:
 | `UnixFd` type semantics (dup constructor vs `adopt`, `release` closes) | ✅* | ✅ | *JVM dup needs junixsocket native support |
 | Unix FD passing over the wire | ⚠️ untested | ✅ | JVM has the conversion path but no independent-peer test |
 | Connection factories (11 total) | 9 of 11 | 11 of 11 | fd-based factories are native-only |
-| Behavior when the bus is unreachable | ⚠️ silent in-process stub | throws `Error` | Known divergence, see below |
+| Behavior when the bus is unreachable | ✅ throws `Error` | ✅ throws `Error` | JVM fixed in #81; the in-process stub backend is an explicit internal test opt-in only |
 | Event loop (`startEventLoop`/`stopEventLoop`) | no-op (always running) | required for dispatch | Same calling pattern works on both |
 | Strict deserialization (signature mismatch rejected) | ✅ | ✅ | Same `System.Error.ENXIO` error |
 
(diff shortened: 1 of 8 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard — approving as tier 1 (bug_fix, behavior-only, no API change; api/ untouched).

**#80 (methodCallTimeout applied):** Verified the resolution exactly mirrors native: every JVM proxy call path (callback-async, both suspend variants, the no-timeout sync overload) funnels through the two sync callMethod overloads, where effectiveTimeout = explicit-nonzero ?: connection default ?: 0 (pre-existing dbus-java default) — the same way native's ProxyImpl.callMethod(message) sends usec=0 and sd_bus_call resolves it via sd_bus_get_method_call_timeout. The 0uL sentinel can't collide with a legitimate explicit value: Duration.ZERO converts to 0 and means "connection default" on BOTH backends, consistent with the library's existing convention. Unset JVM default (Duration.ZERO stored) keeps effectiveTimeout 0 → dbus-java's own default, mirroring native's sd-bus 25s default.

**#81 (throw on unreachable bus):** PureJavaDbusBackend no longer takes a fallback; all failure paths throw com.monkopedia.sdbus.Error ("Failed to open bus…"). Grepped the PR head: StubJvmDbusBackend is referenced only by its own definition/KDoc and explicit test construction or provider mocks — no silent path remains reachable from public factories. JvmRealBusIntegrationTest's connectOrNull gate skips gracefully in no-bus environments (Error resolves to the package-local sdbus Error there, correct).

**Tests in the right places:** the new cross-backend timeout test lives in com.monkopedia.sdbus.integration (commonTest), so ARM runs it inside dbus-run-session; JvmUnreachableBusTest needs no bus (explicit nonexistent socket paths). arm-runtime-tests only executes the native binary and this PR touches no native code, so the bare non-integration ARM pass is unaffected — confirmed green. 100ms-vs-600ms margins reuse the accepted slow-method fixture pattern; flake risk low.

**BACKENDS.md** parity rows and proven-by references match the actual new tests. No #63/#78 regressions (JvmCreateErrorParityTest et al. untouched, full-tests-x64 green).

CI: all 13 checks pass on first run (arm-runtime-tests, full-tests-x64, static-checks, native-build-x64, all 9 pr-smoke legs; nightly-stress skipped by design).

Issue #69, opened 2026-06-11
Investigate: DBusAsyncMethodsTest native parallel-async malloc corruption (ARM flake — possible real race)


monkopedia-coder:
Owner-approved investigation (2026-06-11). INVESTIGATION ONLY — no production code changes, no PR; findings land as a structured comment on this issue and any fix gets its own ticket.

**Symptom**: DBusAsyncMethodsTest's parallel-async stress case intermittently dies with malloc corruption / SIGABRT on the arm-runtime-tests CI job (native target). Historically treated as a known flake and re-run. The signature (heap corruption under concurrent async calls in Kotlin/Native + sd-bus) is consistent with a real race — double-free or use-after-free in the async message/slot/callback lifecycle — which would be a production crash bug and a 1.0 blocker.

**Questions to answer**:
1. Does it reproduce on x64 under repetition (loop the test N times; CI only shows it on ARM — is that load/timing or arch-specific)?
2. With a sanitizer or valgrind/memcheck on the native test binary, what is the actual invalid access? (Note Kotlin/Native sanitizer support is limited; valgrind on the linuxX64 test binary is the pragmatic path. Document tool limitations encountered.)
3. Which lifecycle is involved — PendingAsyncCall/slot release vs reply-callback concurrency vs sd_bus_call_async message ownership? The async plumbing was restructured (visibility only, behavior preserved) in PR #57; confirm the corruption predates the wave (it does per CI history — verify by checking old run logs).
4. Verdict: real library bug (file fix ticket with root cause) / test-harness artifact (fix the test) / cannot reproduce (document the attempts + recommend keeping the re-run policy with a tracking note).

**Repro**: `dbus-run-session -- env JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew linuxX64Test --tests '*DBusAsyncMethods*'` looped; the failing case is the bulk-parallel-invocations stress. CI references: arm-build-test.yaml arm-runtime-tests job history.

monkopedia-coder:
[...]

PR #86, opened 2026-06-11
Native: take the SdBus lock in sd_bus_message_new_method_return (#84)

### LINKED ISSUE #84: Native: sd_bus_message_new_method_return bypasses the SdBus connection lock — heap corruption under server-side async load (root cause of #69 malloc flake)
Root cause identified by the #69 investigation (see the findings comment there for full evidence).

**Bug**: `SdBus.sd_bus_message_new_method_return()` is the only bus-touching wrapper in `src/nativeMain/kotlin/com/monkopedia/sdbus/internal/SdBus.kt` (lines 142-145) that does **not** take the per-connection `lock.withLock { }`. Every sibling (`sd_bus_message_new`, `sd_bus_message_new_method_error`, etc.) does.

**Why it corrupts the heap**: sd-bus refcounts are non-atomic (verified by disassembling the bundled libsystemd 257.2: `sd_bus_ref` is a plain `addl $0x1,(%rbx)`, no lock prefix). `sd_bus_message_new_method_return` internally calls `bus_message_new` → `sd_bus_ref(bus)`, i.e. an unsynchronized `bus->n_ref++`.

The only caller is `MethodCall.createReply()` (`MethodCall.native.kt:94`), which server-side async (`acall`) method handlers execute on `Dispatchers.Default` worker threads (`TypedMethod.kt:145` → `MethodBuilder.kt:94`) **while the connection's event-loop thread is concurrently inside locked `sd_bus_process` performing thousands of bus ref/unref operations** on the same bus. A lost `n_ref` update undercounts the bus refcount → premature `bus_free()` at connection teardown while live messages still hold logical refs → later unref writes freed memory → `malloc(): unsorted double linked list corrupted` / SIGABRT 134 detected at an arbitrary later allocation (which is why CI aborts land in unrelated tests 0.3–1.5 s after the bulk-parallel async tests).

Sync `call {}` methods are unaffected (handler runs on the event-loop thread under the lock); error replies are unaffected (`createErrorReply` → locked `sd_bus_message_new_method_error`). Same bug class as e2eeb3e ("Serialize async slot unref through SdBus lock", Feb 2026), which fixed a raw `sd_bus_slot_unref` in the same async path but missed this hole inside SdBus.kt itself.

**Fix**: wrap the call in `lock.withLock { }`:

```kotlin
override fun sd_bus_message_new_method_return(
    call: CValuesRef<sd_bus_message>?,
    m: CValuesRef<CPointerVar<sd_bus_message>>?
): Int = lock.withLock {
    return sdbus.sd_bus_message_new_method_return(call, m)
}
```

**Additionally (recommended, same PR)**: audit pass asserting every `ISdBus` override that can run concurrently with a live event loop takes the lock — the remaining unlocked ones (`sd_bus_flush`, `sd_bus_get_current_message`, the open/new/close constructors) were reviewed in #69 and are safe today (handshake-time / read-only / teardown-after-loop-exit), but a comment in SdBus.kt documenting the invariant would prevent the next regression of this class.

**Verification path**: existing tests cover behavior; the race itself was never reproduced locally on x64 in ~450 hardened iterations (window is too small under x86-TSO — it fires ~6% of ARM CI runs). Verification = green CI + the disappearance of the arm-runtime-tests exit-134 flake over subsequent runs. There is no practical regression test for the race itself; correctness is by inspection (lock discipline restored).

Closes the open question in #69; that issue's re-run-on-flake policy can be retired once this lands and ARM CI stays clean for a while.

monkopedia-coder:
Fixes #84. Root cause of the #69 malloc/SIGABRT-134 CI flake (full analysis in the #69 investigation findings).

## The fix

`SdBus.sd_bus_message_new_method_return()` was the **only** bus-mutating libsystemd wrapper in `src/nativeMain/kotlin/com/monkopedia/sdbus/internal/SdBus.kt` that did not take the per-connection `lock.withLock { }`. Its sole caller is `MethodCall.createReply()` (`MethodCall.native.kt:94`), which server-side async (`acall`) handlers run on `Dispatchers.Default` workers while the event-loop thread is inside locked `sd_bus_process`. The call internally does `sd_bus_ref(bus)` — a plain non-atomic `n_ref++` in libsystemd — so a lost update undercounts the bus refcount → premature `bus_free()` at teardown → later unrefs write freed memory → heap corruption detected at an arbitrary later allocation (which is why CI aborts landed in unrelated tests right after the bulk-parallel async tests).

The change wraps the call in `lock.withLock { }` exactly like every sibling wrapper, and adds a class-level KDoc on `SdBus` documenting the invariant (every libsystemd entry point that can touch live bus state must hold the lock, citing #69/#84) plus the short list of provably-safe exceptions, so the next change to this file doesn't reintroduce the bug class (as happened after e2eeb3e's Feb-2026 sweep).

## Audit of all ISdBus entry points

Independently re-audited every function in `SdBus.kt` (the only production `ISdBus` implementation; the only other implementor is the test-only `SdBusMock`). Result: **`sd_bus_message_new_method_return` was the only unsafe unlocked entry point**, confirming the #69 finding. The remaining unlocked functions cannot race a live event loop:

- **Bus constructors** (`sd_bus_new`, `sd_bus_open`, `sd_bus_open_system`, `sd_bus_open_user`, `sd_bus_open_user_with_address`, `sd_bus_open_direct` ×2, `sd_bus_open_server`, `sd_bus_open_system_remote`): the bus pointer is thread-local until `openBus`/`openPseudoBus` returns and the event loop starts.
- **`sd_bus_flush` (standalone)**: only called from `finishHandshake()` (`ConnectionImpl.kt:170`), before the bus is shared. The post-send flushes inside `sd_bus_send`/`sd_bus_call_async` already run under the lock.
- **`sd_bus_flush_close_unref` / `sd_bus_close_unref`**: bus `Reference` destructors; `ConnectionImpl.release()` tears down the event loop before `bus.release()`.
- **`sd_bus_get_current_message`**: read-only; only called from `currentlyProcessedMessage`, which is only valid on the event-loop thread inside a handler callback (already under the lock held across `sd_bus_process`).

Also re-verified that all bus-touching calls outside `SdBus.kt` (`ConnectionImpl`, `Message.native`, `MethodCall.native`, `Signal.native`, `MethodReply.native`) go through the `ISdBus` field (i.e. the locked wrappers); the remaining direct cinterop calls are message-local serialization (`sd_bus_message_append/read/...`) or pure validators (`sd_bus_*_is_valid`) that don't touch bus state.

## Deadlock analysis

No deadlock is possible from adding the lock:

- The lock is `kotlinx.atomicfu.locks.ReentrantLock` — **reentrant** (owner-thread + reenter count) on Kotlin/Native.
- Sync `call {}` handlers run inside libsystemd's dispatch from `sd_bus_process`, i.e. on the event-loop thread **while it already holds the lock**; their `createReply()` now re-enters the lock on the same thread, which is exactly the pattern the sync **error** path has used all along (`createErrorReply` → already-locked `sd_bus_message_new_method_error`) — proven in production.
- Async handlers call `createReply()` from a worker thread that holds no lock; it simply contends briefly with the event loop. This is the only lock in play on these paths (no ordering hazard), and the subsequent `reply.send()` → locked `sd_bus_send` already does the same.

## Verification

- `./gradlew ktlintFormat` (no changes), `./gradlew apiDump` (no-op — behavior-only change), `./gradlew ktlintCheck apiCheck` green.
- `dbus-run-session -- ./gradlew linuxX64Test jvmTest :cross_test:linuxX64Test` — all green (the bulk-parallel async tests are the stress test).
- **Local soak**: looped `build/bin/linuxX64/debugTest/test.kexe` filtered to `DBusAsyncMethodsTest.* : CommonApiIntegrationTest.*` (34 tests/iteration) under `dbus-run-session` with `MALLOC_CHECK_=3` and randomized `MALLOC_PERTURB_` per iteration: **36/36 iterations passed, 0 failures** (35-iteration loop + 1 timed warm-up). As noted in #84, the race was never reproducible on x64 (window too small under x86-TSO; it fired ~6% of ARM CI runs), so the soak validates nothing-broke; correctness of the fix rests on the restored lock discipline. The real signal is the arm-runtime-tests exit-134 flake disappearing over subsequent CI runs, after which the #69 re-run-on-flake policy can be retired.

No jvmMain or docs/BACKENDS.md changes (deliberately avoided — sibling PR territory).

### DIFF
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/SdBus.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/SdBus.kt
@@ -57,6 +57,47 @@ import sdbus.sd_bus_message_handler_t
 import sdbus.sd_bus_vtable
 import sdbus.sd_id128_t
 
+/**
+ * Thin wrapper around libsystemd's sd-bus that serializes all bus access through a
+ * per-instance [ReentrantLock].
+ *
+ * ## Locking invariant (#69, #84)
+ *
+ * sd-bus is single-threaded by design: its reference counts (e.g. `bus->n_ref`,
+ * bumped by `sd_bus_ref` from nearly every message/slot operation) are plain
+ * non-atomic increments. sdbus-kotlin runs the connection's event loop on one
+ * thread while user code (proxies, async server handlers on worker dispatchers)
+ * calls into the same bus concurrently, so EVERY libsystemd entry point that can
+ * touch live bus state MUST hold [lock] (`lock.withLock { }`).
+ *
+ * A single unlocked entry point is enough to corrupt the heap: a lost refcount
+ * update undercounts `n_ref`, the bus is freed prematurely at teardown while live
+ * messages still hold logical references, and later unrefs write to freed memory
+ * ("malloc(): unsorted double linked list corrupted" under server-side async load
+ * — the #69 flake, root-caused in #84 to exactly one wrapper missing the lock).
+ *
+ * The lock is reentrant on purpose: message handlers are dispatched by libsystemd
+ * from inside locked `sd_bus_process`, and synchronous handlers create/send their
+ * replies on that same thread, re-entering wrappers like
+ * [sd_bus_message_new_method_return] while the event-loop thread already holds
+ * the lock.
+ *
+ * The only functions allowed to skip the lock are those that provably cannot race
+ * a live event loop:
+ * - bus constructors ([sd_bus_new], the `sd_bus_open*` family): the bus pointer
+ *   is thread-local until the connection is published and the loop starts;
+ * - [sd_bus_flush] standalone: only used during the connection handshake, before
+ *   the bus is shared (the post-send flushes in [sd_bus_send]/[sd_bus_call_async]
+ *   already run under the lock);
+ * - [sd_bus_flush_close_unref]/[sd_bus_close_unref]: teardown destructors that run
+ *   after the event loop has exited;
+ * - [sd_bus_get_current_message]: read-only, only valid on the event-loop thread
+ *   inside a handler callback (which already holds the lock).
+ *
+ * If you add a new wrapper, take the lock unless it falls into one of the
+ * categories above — when in doubt, lock (it is reentrant and cheap compared to a
+ * Feb-2026-style heap-corruption hunt).
+ */
 internal class SdBus : ISdBus {
     private val lock = ReentrantLock()
     private val useAnonymousDirectAuth by lazy {
@@ -142,7 +183,9 @@ internal class SdBus : ISdBus {
     override fun sd_bus_message_new_method_return(
         call: CValuesRef<sd_bus_message>?,
         m: CValuesRef<CPointerVar<sd_bus_message>>?
-    ): Int = sdbus.sd_bus_message_new_method_return(call, m)
+    ): Int = lock.withLock {
+        return sdbus.sd_bus_message_new_method_return(call, m)
+    }
 
     override fun sd_bus_message_new_method_error(
         call: CValuesRef<sd_bus_message>?,


monkopedia-reviewer (approved):
Approving after an independent verification of both the fix and its safety claims.

**Fix matches the #69 root-cause analysis.** The diff wraps `sd_bus_message_new_method_return` in `lock.withLock { }` exactly like every sibling wrapper (same `= lock.withLock { return ... }` idiom), plus the class-level KDoc documenting the locking invariant. Nothing else behaviorally changed; `apiDump` untouched as expected for a nativeMain-internal change.

**Independent audit of SdBus.kt (read top to bottom on the PR branch).** Confirmed `sd_bus_message_new_method_return` was the only unsafe unlocked entry point. The remaining unlocked functions check out against their actual callers:
- `sd_bus_new` / `sd_bus_open*` family (incl. `open_system_remote`, both `open_direct`, `open_server`): constructor-time, bus pointer not yet shared (verified `openBus`/`openPseudoBus`).
- standalone `sd_bus_flush`: only caller is `finishHandshake` (ConnectionImpl.kt:170), invoked from `openBus` before the connection/loop exists; the post-send flushes in `sd_bus_send`/`sd_bus_call_async` are inside the lock.
- `sd_bus_flush_close_unref`/`sd_bus_close_unref`: bus `Reference` destructors; `ConnectionImpl.release()` waits for the event-loop Job to complete before `bus.release()`, and its fail-safe leaks rather than races if the loop never exits.
- `sd_bus_get_current_message`: read-only, only from `currentlyProcessedMessage` (handler context, lock already held across `sd_bus_process`).
Also confirmed `Message`'s ref/unref (ctor + cleaner, Message.native.kt:140/147) go through the locked ISdBus wrappers — important since `sd_bus_message_unref` can chain to a bus unref — and the remaining raw cinterop calls in Message/MethodCall are message-local serialization or pure validators.

**Deadlock verification.**
- Lock is `kotlinx.atomicfu.locks.ReentrantLock` (atomicfu 0.32.1), reentrant on Kotlin/Native; one instance per `SdBus`, one `SdBus` per connection.
- Same-instance check: the vtable callback builds `MethodCall(msg, vtable.obj.connection.getSdBusInterface())` — the same `sdbus` field the EventLoopThread uses for `sd_bus_process`, so sync handlers' `createReply()` re-enters the same lock on the same thread.
- Empirical reentrancy proof: the sync success path has ALWAYS re-entered this lock — `SyncMethodCall.invoke` runs inline on the event-loop thread and its `onResult` calls `reply.send()` → locked `sd_bus_send` (and the error path uses locked `..._method_error`). Adding the lock to `createReply` introduces no new pattern.
- No ordering inversion: the async worker holds no other lock at `createReply` (plain `CoroutineScope.launch` → handler → createReply → send), and `send()` immediately after already acquires this same lock, so no new lock-pair ordering is introduced anywhere.

**Soak evidence**: 36/36 hardened local x64 iterations is informative but not proof (race never reproduced on x64, as the PR honestly notes); correctness rests on the restored lock discipline, which I verified by inspection. The real confirmation will be arm-runtime-tests staying clean over future runs — this PR's own ARM run passed.

**CI**: all 13 checks pass (nightly-stress skipped as designed), including arm-runtime-tests — no exit-134.

Classification: bug_fix → tier 1 (sdbus-kotlin uses defaults). Merging.

Issue #67, opened 2026-06-11
Docs/sample: server-side sample — export an object, emit signals, properties


monkopedia-coder:
Part of the 1.0 docs overhaul (owner-approved 2026-06-11).

The only sample (samples/bluez-scan) is client-side. Add the missing half: a server/adaptor sample that exports an object on the session bus, with:
- An interface defined via D-Bus introspection XML and generated through the codegen plugin (generateAdapters=true) — same flow as bluez-scan uses for proxies
- A method, a read/write property (with PropertiesChanged emission), and a signal the service emits periodically
- addVTable usage via the DSL, addObjectManager, startEventLoop lifecycle, clean shutdown
- Runnable on BOTH targets (jvm + linuxX64), mirroring bluez-scan's structure (own gradle build, README with run instructions incl. dbus-run-session usage so it works without a desktop session)
- A tiny client mode (or busctl instructions) to poke it, so a reader can see round-trips

Use the current post-freeze API only. Constraints: new sample directory + README; no library code changes; apiDump no-op.

PR #88, opened 2026-06-12
Docs/sample: server-side sample (samples/demo-service)

monkopedia-coder:
## What this adds

The missing **server/adaptor** half of the samples. `bluez-scan` is a pure client; `samples/demo-service` exports an object on the session bus and shows the adaptor side: a method, a read/write property, and a periodic signal, with the full connection lifecycle.

The interface is defined in [`src/dbusMain/DemoService.xml`](https://github.com/Monkopedia/sdbus-kotlin/blob/docs/issue-67-server-sample-v2/samples/demo-service/src/dbusMain/DemoService.xml) and generated through the codegen plugin (`generateAdapters = true`, `generateProxies = true`):

- method `Greet(s) -> s`
- read/write property `Prefix` that emits `PropertiesChanged` when written
- signal `Tick(t)` emitted once per second

`Main.kt` wires `createBusConnection(ServiceName)` -> `startEventLoop()` -> `addObjectManager()` -> `createObject()` -> register vtable -> emit ticks -> clean shutdown (`stopEventLoop()` + resource releases). Modes: default (run until Ctrl-C), `<seconds>` (run then shut down), and `client` (poke a running service).

### One deliberate deviation from the generated `register()`
The sample subclasses the generated `DemoService1Adaptor` but **overrides `register()`** to bind the property with the explicit `prop { withGetter { … }; withSetter { … } }` DSL instead of the generated `with(::prefix)` property-reference binding. This is necessary: the generated binding gave the setter a place to emit `PropertiesChanged`, and — heads-up for a separate library issue — `PropertyVTableItem.with(KProperty0)` installs `getter = { receiver.get() }`, but `PropertyGetCallback` is `(PropertyGetReply) -> Unit`, so the value is **discarded and never serialized into the reply**. A remote property `Get` against an adaptor wired via `with` therefore hangs on native (methods are unaffected). The sample sidesteps this with the working `withGetter`/`withSetter` DSL (which the spec explicitly allows) and does **not** touch library code. Worth filing/fixing separately.

## Dependency mechanism: composite build (local, not the published artifact)

The current API (`startEventLoop`, `withGetter`/`withSetter`, typed property flows, `Duration` timeouts) does not exist in the published 0.4.5 jar. So, mirroring `bluez-scan`, [`settings.gradle.kts`](https://github.com/Monkopedia/sdbus-kotlin/blob/docs/issue-67-server-sample-v2/samples/demo-service/settings.gradle.kts) uses `includeBuild("../..")`. Gradle substitutes `com.monkopedia:sdbus-kotlin` with the local source tree, so the `0.4.5` coordinate in `build.gradle.kts` is just a placeholder and the sample always compiles against the in-repo (post-0.5.0) library. No `publishToMavenLocal` needed.

## Build + run proof

Both targets build (`JAVA_HOME=/usr/lib/jvm/java-21-openjdk`): `linkReleaseExecutableNative` and `compileKotlinJvm` succeed.

**Native server, driven by `busctl` under `dbus-run-session`:**
```
===== introspect =====
NAME    TYPE     SIGNATURE RESULT/VALUE FLAGS
.Greet  method   s         s            -
.Prefix property s         "Hello"      emits-change writable
.Tick   signal   t         -            -
===== GetManagedObjects =====
… "/com/monkopedia/demo/service" … "com.monkopedia.demo.DemoService1" 1 "Prefix" s "Hello"
===== call Greet world =====       -> s "Hello, world!"
===== get-property Prefix =====    -> s "Hello"
--- monitor while set-property Prefix=Howdy ---
  Member=Tick               UINT64 4;
  Member=PropertiesChanged  STRING "Prefix"; STRING "Howdy";
  Member=Tick               UINT64 5/6/7;
===== get-property Prefix (after set) =====  -> s "Howdy"
===== call Greet world (new prefix) =====    -> s "Howdy, world!"
Shutting down ...   (clean exit)
```

**JVM client (`./gradlew runJvm --args=client`) against the native server — cross-target round trip:**
```
Prefix is currently: "Hello"
greet("world") -> "Hello, world!"
Setting Prefix to "Hi" (server should emit PropertiesChanged) ...
greet("world") -> "Hi, world!"
tick -> 5
tick -> 6
tick -> 7
```
Native `client` against the native server behaves identically.

## Target notes (documented in the README)

- **linuxX64 (native):** full sd-bus server — external processes can introspect, call, get/set properties, and receive signals.
- **JVM (dbus-java):** runs as a first-class client (proven above) and its server mode starts/exports/emits/shuts down cleanly, but the dbus-java backend routes exported objects through an in-process dispatch table, so serving arbitrary method calls to *separate* processes is limited. For a production D-Bus service, run native.

## Verification
- `./gradlew apiDump` from repo root → `git diff --quiet -- api/` clean (sample touches no library API).
- `./gradlew ktlintCheck` passes. The sample is a standalone build (no library files changed); all changes live under `samples/demo-service/`.


monkopedia-reviewer (approved):
Approving — docs_only / tier 1.

Scope: all 9 changed files live under samples/demo-service/ (README, build/settings gradle, wrapper, Main.kt, DemoService.xml). No file under src/, codegen/, plugin/, or api/*.api is touched — the prior attempt's stray edit to PropertyVTableItem.kt is NOT present. apiDump is a no-op as required.

Coverage vs #67: XML interface + codegen generateAdapters/generateProxies; method Greet(s)->s; read/write property Prefix emitting PropertiesChanged via emitPropertiesChangedSignal; periodic Tick(t) signal; full lifecycle createBusConnection(name) -> startEventLoop -> addObjectManager -> createObject -> addVTable -> emit -> stopEventLoop + release() clean shutdown. Both jvm (runJvm) and linuxX64 (runReleaseExecutableNative) targets. README documents dbus-run-session + busctl introspect/call/get-property/set-property/monitor/GetManagedObjects, plus a client mode.

API correctness: verified every call against current source — method{inputParamNames/outputParamNames; acall}, signal{with<ULong>(name)}, prop{withGetter/withSetter<String>}, emitPropertiesChangedSignal(iface, List<PropertyName>), Connection/Object : Resource (release valid), addObjectManager(path): Resource, withService().createProxy(). All post-0.5.0.

Dependency mechanism: composite build via includeBuild("../..") in settings.gradle.kts substitutes the local source tree; the 0.4.5 coordinate is a documented placeholder. Sound and clearly documented. Standalone build — NOT wired into root settings.gradle.kts, so root CI doesn't build it (mirrors bluez-scan).

Build+run proof: PR body shows busctl introspect/call/get-set/monitor output against the native server and a cross-target JVM-client round trip.

Nice catch flagging the PropertyVTableItem.with(KProperty0) getter-discard bug as a separate library issue without touching library code.

CI green: full-tests-x64, native-build-x64, static-checks, all pr-smoke matrix pass; arm-runtime-tests clean (57s, no malloc abort); nightly-stress skipped.

PR #91, opened 2026-06-12
Correct BACKENDS.md JVM server-side rows (#90)

### LINKED ISSUE #90: JVM backend can't serve exported objects to external processes (in-process dispatch only) — and parity matrix may over-claim JVM server rows

monkopedia-coder:
## Why

Investigating #90 (owner decision 2026-06-12: fix the JVM cross-process export gap before 0.5.0). This PR lands **only the docs correction** so the contract doc is honest at the freeze; the code fix is deferred pending an owner scoping decision (see the investigation comment on #90). This PR does **not** close #90.

## The gap (confirmed)

The JVM (dbus-java) backend's server side routes incoming method/property calls through an **in-process dispatch table** (`JvmStaticDispatch` / `LocalManagedObjectsRegistry` in `PureJavaDbusBackend.kt`). A JVM-hosted service claims its bus name, exports, emits signals to the real bus, and shuts down cleanly — and in-process JVM clients reach it — but a method/property call from a **separate process** (busctl, another client) gets `UnknownObject`. Native is a full server backend.

The BACKENDS.md matrix marked the server-relevant rows (Properties `Get`/`Set`/`GetAll`, ObjectManager `GetManagedObjects`) as full parity, but the evidence was the dbusmock suites — which exercise us **as a client** against a foreign server — plus JVM in-process own-server round-trips. JVM *serving* those to an external peer was never verified and partly doesn't work.

## What changed (docs only)

- Matrix: annotated the Properties / ObjectManager rows as **client-verified**; added a dedicated **"Serving exported objects to external processes"** row (JVM `⚠️ in-process only`, native `✅`).
- New **"Server-side object export (JVM limitation)"** section: explains the gap, why it's not a one-line fix (dbus-java's `exportObject` dispatches by reflecting a statically-typed `DBusInterface`, vs sdbus-kotlin's dynamic vtable + its own codec; dbus-java exposes no raw incoming-method-call hook), and recommends the native backend for cross-process services. Notes that signal emission (incl. `PropertiesChanged`, `InterfacesAdded`/`Removed`) **does** reach the real bus.
- Serving caveats added to the Properties / ObjectManager prose and to the in-process dispatch note.

No code or API changes.

### DIFF
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -28,10 +28,11 @@ Every row below is pinned by tests on current `main`, not by intention:
 | Error propagation: named D-Bus errors round-trip name + message verbatim | ✅ | ✅ | Incl. foreign (non-sdbus) peers; JVM fixed in #72 |
 | errno → D-Bus error-name mapping for locally created errors | ✅ | ✅ | JVM pinned to native output |
 | Call-after-release fails with `IllegalArgumentException("Connection has already been released")` | ✅ | ✅ | |
-| Properties: `Get`/`Set`/`GetAll`, property delegates (`values()`/`changes()`) | ✅ | ✅ | |
-| `PropertiesChanged` emission incl. getter resolution | ✅ | ✅ | JVM fixed in #28 |
-| Signals: emission, subscription, `signalFlow` | ✅ | ✅ | |
-| ObjectManager: `GetManagedObjects`, `InterfacesAdded`/`Removed`, `ObjectManagerProxy` flows | ✅ | ✅ | |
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Docs-only correction of the BACKENDS.md JVM server-side rows (#90). Verified: scopes the previously-over-claiming Properties/ObjectManager rows as client-verified, adds an honest 'Serving exported objects to external processes' row (JVM in-process-only / native full) and a dedicated explanatory section. Accurate interim contract state while the #90 approach-A cross-process-serve fix (owner-gated for 0.5.0) is implemented — that fix will re-flip these rows to parity with a cross-process acceptance test. CI green.

PR #92, opened 2026-06-12
WIP: Export JVM server objects via dbus-java for cross-process reachability (approach A, partial) (#90)

### LINKED ISSUE #90: JVM backend can't serve exported objects to external processes (in-process dispatch only) — and parity matrix may over-claim JVM server rows

monkopedia-coder:
**Status: PARTIAL / WIP** — a working, tested subset of issue #90 approach A. Does **not** close #90 (struct & multi-out method signatures remain in-process only; see "What remains").

## Problem
dbus-java only dispatches incoming method/property/ObjectManager calls to objects registered through its native `AbstractConnection.exportObject(path, DBusInterface)` — a reflection-driven path requiring a statically-typed `DBusInterface`. sdbus-kotlin's JVM vtable is dynamic, so nothing was registered with dbus-java and external callers got `UnknownObject`. (Signal emission already reached the bus and is untouched.)

## Mechanism (approach A)
New `SdbusObjectExporter` (jvmMain) registers each exported object with dbus-java alongside the existing in-process `JvmStaticDispatch` table:

- For each D-Bus interface, ByteBuddy synthesizes at runtime a `DBusInterface` subtype (annotated `@DBusInterfaceName`) whose methods (annotated `@DBusMemberName`) carry Java parameter/return types that dbus-java's `Marshalling.getDBusType` turns back into the vtable's **exact** wire signatures.
- A `java.lang.reflect.Proxy` implementing those interfaces (plus `Properties`/`ObjectManager` when served) is exported. Its `InvocationHandler` converts dbus-java's deserialized Java values into the sdbus value model, routes through the **same** `JvmStaticDispatch` handlers the vtable already populates, and converts the reply back.
- **dbus-java owns the wire codec** — there is no second serializer to keep in lockstep with the #78 codec. The only conversion is at the Java-value boundary, reusing the already-tested `toJavaSignalValue`/`fromJavaWireValue` helpers from the client path.

The bridge is strictly additive and best-effort: the in-process short-circuit is unchanged, and any method whose signature it cannot yet map is left un-exported rather than failing `addVTable`.

## Signature mapping
D-Bus single-complete-types map to the Java types dbus-java's `getJavaType`/`getDBusType` round-trip identically: `b/y/n/q/i/u/x/t/d → Boolean/Byte/Short/UInt16/Integer/UInt32/Long/UInt64/Double`, `s → String`, `o → DBusPath`, `h → FileDescriptor`, `v → Variant`, `a… → List<…>`, `a{…} → Map<…>` (built as ByteBuddy parameterized types). Object-path replies are emitted as `DBusPath` and containers recurse with their element/entry signatures so the wire signature — not the value shape — drives the Java type.

## New dependency (flag for owner awareness)
`net.bytebuddy:byte-buddy:1.18.2` (~3MB, widely used) added as a **jvmMain `implementation`** dependency — a dependency-graph change to the published `-jvm` artifact. dbus-java-core does not bring a bytecode lib transitively, so this is genuinely new on the main classpath (it was previously only a transitive `testImplementation` of mockk). No public API change; `apiDump` is a no-op.

## Acceptance test (fails-before / passes-after)
`JvmCrossProcessExportTest` drives a JVM-exported object with **busctl** — a genuinely external process. (Two sdbus-kotlin connections in the *same* JVM short-circuit through the global `JvmStaticDispatch` single-candidate fallback and never hit the wire, so they cannot detect the gap; busctl has no such entry and must go through dbus-java export.) It asserts a method call, `Properties.Get`, and `ObjectManager.GetManagedObjects` all return real values.

- **Before** the bridge: every busctl op fails with `… is not an object provided by this process` (`UnknownObject`) — verified by disabling export.
- **After**: all three pass.

## BACKENDS.md
Server-side rows re-flipped to reflect cross-process serve for the common surface (busctl-verified), with the struct/multi-out gaps documented honestly.

## What remains (tracked in #90; in-process only for now)
- Methods whose signatures contain a **struct** `(...)` — needs runtime generation of concrete dbus-java `Struct` subclasses (positioned fields + assigning constructor); dbus-java's `getDBusType(DBusStructType)` does **not** round-trip, so a generated `Struct` class is required. (Struct-*valued* properties / `GetManagedObjects` entries already work — they travel inside a `Variant`.)
- Methods returning **multiple out-values** — needs runtime `Tuple` subtype generation for the parameterized return type dbus-java requires.
- `g` (signature) arguments.

## Verification
`ktlintFormat` → `apiDump` (no-op) → `ktlintCheck apiCheck` clean; `jvmTest`, `linuxX64Test`, `cross_test:jvmTest`, `cross_test:linuxX64Test` all green under `dbus-run-session`; `compileKotlinLinuxX64`, `codegen:test`, `plugin:test` green. Nothing regressed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -108,6 +108,10 @@ kotlin {
                 implementation(libs.kotlinx.datetime)
                 implementation(libs.dbus.java.core)
                 runtimeOnly(libs.dbus.java.transport.junixsocket)
+                // Runtime synthesis of typed DBusInterface adapters so JVM-hosted server
+                // objects are reachable cross-process via dbus-java's native object export
+                // (issue #90). New jvmMain runtime dependency on the published -jvm artifact.
+                implementation(libs.byte.buddy)
                 implementation(kotlin("stdlib"))
             }
         }
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -28,11 +28,11 @@ Every row below is pinned by tests on current `main`, not by intention:
 | Error propagation: named D-Bus errors round-trip name + message verbatim | ✅ | ✅ | Incl. foreign (non-sdbus) peers; JVM fixed in #72 |
 | errno → D-Bus error-name mapping for locally created errors | ✅ | ✅ | JVM pinned to native output |
 | Call-after-release fails with `IllegalArgumentException("Connection has already been released")` | ✅ | ✅ | |
-| Properties: `Get`/`Set`/`GetAll`, property delegates (`values()`/`changes()`) — **as a client** | ✅ | ✅ | Client-verified; JVM *serving* properties to an external peer is in-process only — see "Server-side object export" |
+| Properties: `Get`/`Set`/`GetAll`, property delegates (`values()`/`changes()`) — **client + server** | ✅ | ✅ | Client-verified; JVM now also *serves* `Get`/`Set`/`GetAll` to external peers via dbus-java export (#90), busctl-verified cross-process — see "Server-side object export" |
 | `PropertiesChanged` emission incl. getter resolution | ✅ | ✅ | JVM fixed in #28; JVM emits to the real bus, but external reception is not test-verified |
 | Signals: emission, subscription, `signalFlow` | ✅ | ✅ | JVM emission goes to the real bus; subscription/`signalFlow` are cross-process-verified as a client |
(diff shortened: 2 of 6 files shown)


monkopedia-coder:
Closing the ByteBuddy approach. Owner decided (2026-06-12) to OWN the JVM connection instead (epic #) rather than ship a runtime-type-generation stopgap we'd delete in weeks. The feasibility spike retired all risks (incl. external serve + fd-passing) and showed owning the connection kills the whole dbus-java impedance-bug class. This WIP branch stays as a reference; not merging.

Issue #93, opened 2026-06-12
Epic: own the JVM D-Bus connection (replace dbus-java transport with junixsocket + our own marshaller/dispatch)


monkopedia-coder:
Owner-decided 2026-06-12. Replace the JVM backend's reliance on dbus-java's **connection** with a self-owned D-Bus connection layer — raw junixsocket + SASL EXTERNAL + D-Bus wire framing + our own marshaller + read/dispatch loop — mirroring how the native backend owns sd-bus. **Part of the 0.5.0 release** (0.5.0 waits for this; no ByteBuddy stopgap).

## Why
Nearly every JVM bug this cycle (#71/#72/#74 wire marshalling, #80/#81 behavior, #90 serving) traced to wrapping dbus-java's static-typed, reflection-driven model. The feasibility spike (#90 comments) proved owning the connection is feasible and cleanly bounded (~3–4 engineer-weeks), with ALL four risk prototypes passing against a real bus: connect+SASL+Hello, hand-framed call round-trip, **external busctl reaching a served method** (impossible through dbus-java), and **SCM_RIGHTS fd-passing** (also closes #83's worry). Crux finding: `Message.jvm.kt` delegates all byte marshalling to dbus-java today, so owning the connection requires a byte-level marshaller — but that turns the entire #71/#72/#74/#78/#80/#81 bug corpus into our own spec-level unit tests, and lets us drop BOTH dbus-java and bytebuddy.

## Outcome
- Public sdbus-kotlin API unchanged (internal backend swap behind `JvmDbusConnection`/`JvmDbusProxy`).
- JVM client + serve on ONE owned connection / ONE unique name (native model).
- dbus-java dependency removed; no bytebuddy.
- Closes #90 (cross-process serving) and #83 (JVM fd wire-test gap).

## Phases (each a reviewed PR)
1. **Byte-level D-Bus marshaller** (pure Kotlin, ~4–5d) — serialize/deserialize the existing JVM value model ↔ wire bytes; alignment, all basic types, a/()/a{}/v, both endiannesses. Bug corpus (#71/#72/#74/#80/#81) + dbusmock type-matrix as spec tests. Standalone, no transport.
2. **Transport + SASL EXTERNAL + framing + bootstrap** (~3d) — junixsocket connection, address parsing, auth, message header read/write, Hello/RequestName/AddMatch.
[...]

Issue #27, opened 2026-06-08
Tests: unsigned-type (UByte/UShort/UInt/ULong) round-trips, both backends


monkopedia-coder:
Round-trip coverage for the unsigned inline-class types — `UByte`/`UShort`/`UInt`/`ULong` — as scalars, list elements, dict keys/values, and inside variants, on both backends.

The unsigned-as-inline-class mapping has produced `ClassCastException`s twice: the native `ListEncoder` boxing (#8) and the JVM `ay`→`UByte` decode. Lock the behaviour with systematic coverage.

PR #94, opened 2026-06-12
Phase 1 (#93): pure-Kotlin D-Bus byte marshaller

monkopedia-coder:
## Summary

Phase 1 of epic #93 (own the JVM D-Bus connection): a **pure-Kotlin, byte-level D-Bus marshaller** for the JVM backend. It converts between the existing JVM value model in `Message.jvm.kt` and D-Bus wire bytes, driven by a type signature. Standalone and independently testable — **no transport, no connection changes, no dispatch** in this phase, and **not yet wired into** `Message.jvm.kt`'s send/receive (that's phase 2/3). dbus-java stays a dependency and is used here purely as a test oracle.

New file: `src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusMarshaller.kt`.

## What the marshaller covers
- **Signature tokenizer**: recursive-descent, handles nested `a`, `()`, `a{}`, `v` (e.g. `(ii)`, `a{sv}`, `aai`, `a{oa{sa{sv}}}`).
- **All basic types**: `y b n q i u x t d s o g h`.
- **Containers**: arrays `a`, structs `()`, dict-entries `a{}`, variants `v`.
- **Alignment**: faithful natural alignment counted from the start of the message body, NUL padding; array length is the **content byte-length** (not element count) with element-alignment padding inserted **even for empty arrays**.
- **Endianness**: BOTH little- and big-endian **read and write**.
- Operates on the **same value model** `Message.jvm.kt` already uses (boxed `Boolean`/`Short`/`Int`/`Long`/`Double`, `UByte`/`UShort`/`UInt`/`ULong`, `String`, `List` for arrays, `Map` for dicts, `JvmStructPayload` for structs, `JvmVariantPayload` for variants), so it slots into the connection later without a value-model change.

## Value-model extension
**None needed.** The marshaller consumes/produces the existing value model unchanged. Variants are handled as `JvmVariantPayload`; the `Variant` ⇄ `JvmVariantPayload` conversion already exists in the connection layer and will bridge in a later phase (flagged in the file KDoc).

## Test coverage
Three angles, 45 test cases, all green:

1. **Differential vs dbus-java** (`DBusMarshallerDifferentialTest`) — the oracle. For a broad matrix (basic types, strings, arrays incl. empty, dicts, structs incl. nested, variants, `a{sv}`, multi-top-level) it marshals with **our** marshaller and with **dbus-java's internal `Message.append` body marshaller** (reached by reflection on a bare `Message`; the `big` field selects endianness) and asserts **byte-identical** output in **both endiannesses**. It also **cross-demarshals**: dbus-java's reference bytes → our reader → re-marshal must reproduce the reference.

2. **Spec canaries** (`DBusMarshallerSpecTest`) — hand-computed expected byte sequences from the D-Bus spec (int32 LE/BE, string, signature, `ay`/`ai`, struct `(is)`, variant, `at` alignment padding, empty `a{sv}` padding, multi-out `isi`), so correctness is anchored to the **spec, not dbus-java** (recall #71/#72/#74 were dbus-java quirks).

3. **Bug corpus as regression tests** — encoded as spec-correct assertions:
   - **#71** structs both directions: `(is)`, `((is)ai)`, `a(is)`, wide struct round-trips.
   - **#74** multiple top-level values / multi-out: `isi`, `iis`, `tu` byte-identical + round-trip.
   - **#26/#11** empty collections: `as`/`ai`/`ay`/`a{sv}`/`a{ss}`/`a(is)` — incl. the empty-`a{sv}` element-alignment padding.
   - **#27** unsigned round-trips incl. `ULong.MAX` across the signed-Long boundary, and all unsigned widths min/max.
   - dbusmock type-matrix shapes (`a{oa{sa{sv}}}`, `av`, `aai`, variant-wrapped struct, wide struct).
   - Round-trip property over the whole matrix in **both endiannesses**, plus a **hand-crafted big-endian** body to exercise the `'B'` read branch (not just mirror our own writer).

Where our output would differ from a known dbus-java bug case, the spec canaries assert **our** (spec-correct) bytes.

## Verification
- `./gradlew apiDump` → **no-op** (marshaller is `internal`, no public API change).
- `./gradlew ktlintCheck apiCheck` → pass.
- `dbus-run-session -- ./gradlew :jvmTest` → green (full suite; nothing regressed).
- `./gradlew compileKotlinLinuxX64` → pass (multiplatform build intact).

## Remaining for later phases (not this PR)
Wiring the marshaller into `Message.jvm.kt` send/receive (phase 2/3), transport/SASL/framing (phase 2), `h`/fd index ↔ out-of-band fd array (phase 5). The `h` type is marshalled here as a plain `uint32` and is excluded from the dbus-java differential (dbus-java writes an fd index for `FileDescriptor`); it's covered by round-trip + the format is identical to `u`.

Refs #93.

### DIFF
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusMarshaller.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusMarshaller.kt
@@ -1,0 +1,600 @@
+/**
+ *
+ * (C) 2024 - 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Project: sdbus-kotlin
+ * Description: High-level D-Bus IPC kotlin library based on sd-bus
+ *
+ * This file is part of sdbus-kotlin.
+ *
+ * sdbus-kotlin is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * sdbus-kotlin is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with
+ * sdbus-kotlin. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.monkopedia.sdbus.internal.jvmdbus
+
+import com.monkopedia.sdbus.Message
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.Signature
+import com.monkopedia.sdbus.UnixFd
+
+/**
+ * Pure-Kotlin byte-level D-Bus marshaller for the JVM backend (epic #93, phase 1).
+ *
+ * Converts BETWEEN the JVM value model used by `Message.jvm.kt` (boxed Boolean/Short/Int/Long/
+ * Double, [UByte]/[UShort]/[UInt]/[ULong], String, [ObjectPath]/[Signature]/[UnixFd], [List] for
+ * arrays, [Map] for dictionaries, [Message.JvmStructPayload] for structs, and
+ * [Message.JvmVariantPayload] for variants) and D-Bus wire bytes, driven by a D-Bus type
+ * signature.
+ *
+ * Implements the D-Bus marshalling rules faithfully: per-type natural alignment counted from the
+ * start of the message body, all basic types, arrays/structs/dict-entries/variants, and BOTH
+ * little- and big-endian read+write. This phase is standalone (no transport, no connection wiring)
+ * and exercised by differential tests against dbus-java, hand-computed spec canaries, and the
+ * #71/#74/#26/#11/#27 bug corpus.
+ *
+ * NOTE: variants are consumed/produced as [Message.JvmVariantPayload]; the
+ * `com.monkopedia.sdbus.Variant` <-> [Message.JvmVariantPayload] conversion already exists in the
+ * connection layer and will bridge in a later phase.
+ */
+internal enum class Endian(val code: Byte) {
+    LITTLE('l'.code.toByte()),
+    BIG('B'.code.toByte());
+
+    companion object {
+        fun fromCode(code: Byte): Endian = when (code.toInt().toChar()) {
+            'l' -> LITTLE
+            'B' -> BIG
+            else -> throw DBusMarshallingException("Unknown endianness flag: $code")
+        }
+    }
+}
+
+internal class DBusMarshallingException(message: String) : RuntimeException(message)
+
+/**
+ * The parsed model of a single complete D-Bus type, carrying its natural [alignment] and the
+ * signature fragment it [code]s back to.
+ */
+internal sealed class DBusType {
+    abstract val alignment: Int
+    abstract val code: String
+
+    data class Basic(val type: Char) : DBusType() {
+        override val alignment: Int get() = alignmentOfBasic(type)
+        override val code: String get() = type.toString()
+    }
+
+    data class ArrayType(val element: DBusType) : DBusType() {
+        override val alignment: Int get() = 4
+        override val code: String get() = "a${element.code}"
+    }
+
+    data class StructType(val fields: List<DBusType>) : DBusType() {
+        override val alignment: Int get() = 8
+        override val code: String get() = "(${fields.joinToString("") { it.code }})"
+    }
+
+    data class DictEntryType(val key: DBusType, val value: DBusType) : DBusType() {
+        override val alignment: Int get() = 8
+        override val code: String get() = "{${key.code}${value.code}}"
+    }
+
+    data object VariantType : DBusType() {
+        override val alignment: Int get() = 1
+        override val code: String get() = "v"
+    }
+
+    companion object {
+        fun alignmentOfBasic(type: Char): Int = when (type) {
+            'y', 'g' -> 1
+            'n', 'q' -> 2
+            'b', 'i', 'u', 'h', 's', 'o' -> 4
+            'x', 't', 'd' -> 8
+            else -> throw DBusMarshallingException("Unknown basic type '$type'")
+        }
+    }
+}
+
+/**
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Approving — Phase 1 marshaller (epic #93). Reviewed with full rigor; the byte-level correctness is sound.

**Alignment** — verified body-relative (writer counts from size==0 = body start; reader from the passed offset, which in D-Bus is always 8-aligned so body-relative == message-relative). Every type's natural alignment is correct: y/g=1, n/q=2, b/i/u/h/s/o=4, x/t/d=8, struct()=8, dict-entry=8, array=4, variant=1. Pad bytes are NUL.

**Array length semantics** — length is the CONTENT byte-length (patched as size-contentStart AFTER the pre-element alignment padding), not element count, and the element-alignment pad is written even for empty arrays. The empty-collection trap is handled and spec-anchored: empty a{sv} = 00000000 00000000 (4-byte len 0 + pad to the 8-byte dict-entry boundary), empty at likewise reserves the pad; emptyDict_asv_hasElementAlignmentPadding asserts this against hand-computed bytes.

**Strings** — s/o: uint32 length excluding NUL + UTF-8 + NUL; g: uint8 length + NUL. Verified against canary_string/canary_signature.

**Endianness** — no hardcoded LE anywhere; putRaw/putUInt32At/readRaw all branch on the endian param for both read and write. BE is genuinely exercised: canary_int32_bigEndian, canary_double_bigEndian (hand-computed), and bigEndianRead_handCrafted feeds a BE body the writer never produced through the reader's 'B' branch.

**Variants** — signature (g, 1-byte len) then value per that single complete type, with a parseOne completeness check; nested variants (av, variant-of-struct) round-trip.

**Differential oracle is sound** — drives dbus-java's actual internal Message.append body marshaller via reflection, both endiannesses, and correctly handles the endianness-source finding: it sets the private  field directly (documented that updateEndianess only writes the header flag byte, not the marshalling direction). Cross-demarshal (dbus-java bytes -> our reader -> re-marshal == reference) included.

**Bug-corpus = spec-correct, not dbus-java-match (the crucial check)** — every bug-corpus class has at least one hand-computed spec assertion in DBusMarshallerSpecTest, so there's no false-confidence/re-baking risk: #71 structs (canary_struct_is), #74 multi-out (canary_multipleTopLevel_struct74 'isi' streaming alignment), #26/#11 empty collections (emptyArray_as/ai/ay + emptyDict_asv), #27 unsigned incl. ULong.MAX (canary_uint64_max = 8×ff, canary_uint32). The differential's agreement with dbus-java on these is consistent because dbus-java's raw body marshaller is itself spec-correct — those bugs lived in the higher-level reflection/wrapping layer, not append() — so the differential is supplementary, not the anchor. Round-trip property holds across the matrix in both endiannesses.

**Scope** — internal-only (Endian/DBusType/DBusMarshaller all ), apiDump no-op (no api/ files in diff), Message.jvm.kt send/receive untouched (correct for phase 1; existing JvmStructPayload/JvmVariantPayload reused with no value-model change), nothing regressed. CI fully green (full-tests-x64, static-checks, native-build, arm-runtime-tests — the #86 fix held, no abort — all pr-smoke pass).

Minor (non-blocking, phase-5 territory): 'h' has no dedicated round-trip test despite the PR body's 'covered by round-trip' wording, but it shares the exact putAligned(4)/readRaw(4) path as the heavily-tested 'u'/'i', and real fd-array handling is phase 5. Not worth gating.

Solid foundation — every byte checked out.

monkopedia-reviewer (commented):
Review rationale (the prior approval body was mangled by shell backtick-expansion; this is the clean version).

Phase 1 marshaller (epic #93) — reviewed with full rigor; byte-level correctness is sound.

**Alignment** — body-relative (writer counts from size==0 = body start; reader from the passed offset, which in D-Bus is always 8-aligned so body-relative == message-relative). Per-type natural alignment all correct: y/g=1, n/q=2, b/i/u/h/s/o=4, x/t/d=8, struct=8, dict-entry=8, array=4, variant=1. Pad bytes are NUL.

**Array length** — CONTENT byte-length (patched as size-contentStart AFTER the pre-element alignment pad), not element count; element-alignment pad written even for empty arrays. Empty-collection trap handled and spec-anchored: empty a{sv} = 00000000 00000000, empty at reserves its pad; emptyDict_asv_hasElementAlignmentPadding asserts hand-computed bytes.

**Strings** — s/o: uint32 length excluding NUL + UTF-8 + NUL; g: uint8 length + NUL. Verified via canary_string / canary_signature.

**Endianness** — no hardcoded LE; putRaw/putUInt32At/readRaw all branch on the endian param for read AND write. BE genuinely exercised: canary_int32_bigEndian, canary_double_bigEndian (hand-computed), and bigEndianRead_handCrafted feeds a BE body the writer never produced through the reader's BE branch.

**Variants** — signature (g, 1-byte len) then value per a single complete type, with a parseOne completeness check; nested variants round-trip.

**Differential oracle is sound** — drives dbus-java's actual internal Message.append body marshaller via reflection, both endiannesses, and handles the endianness-source finding correctly: sets the private big field directly (documented that updateEndianess only writes the header flag byte, not marshalling direction). Cross-demarshal (dbus-java bytes -> our reader -> re-marshal == reference) included.

**Bug-corpus = spec-correct, not dbus-java-match (the crucial check)** — every bug-corpus class has a hand-computed spec assertion in DBusMarshallerSpecTest, so no false-confidence/re-baking risk: #71 structs (canary_struct_is), #74 multi-out (canary_multipleTopLevel_struct74 isi streaming alignment), #26/#11 empty collections (emptyArray_as/ai/ay + emptyDict_asv), #27 unsigned incl. ULong.MAX (canary_uint64_max = 8x ff, canary_uint32). The differential's agreement with dbus-java on these is consistent because dbus-java's raw body marshaller is itself spec-correct — those bugs lived in the higher-level reflection/wrapping layer, not append() — so the differential is supplementary, not the anchor. Round-trip property holds across the matrix in both endiannesses.

**Scope** — internal-only, apiDump no-op (no api/ files in diff), Message.jvm.kt send/receive untouched (correct for phase 1; existing JvmStructPayload/JvmVariantPayload reused, no value-model change), nothing regressed. CI fully green.

Minor (non-blocking, phase-5 territory): 'h' has no dedicated round-trip test despite the PR body's wording, but shares the exact putAligned(4)/readRaw(4) path as the heavily-tested u/i, and real fd-array handling is phase 5.

PR #95, opened 2026-06-12
Phase 2 (#93): owned D-Bus connection — junixsocket transport, SASL EXTERNAL, framing, bootstrap

monkopedia-coder:
## Phase 2 of epic #93 — owned D-Bus connection (JVM)

Builds on Phase 1's pure-Kotlin marshaller (now on main) to deliver the low-level, self-owned D-Bus connection for the JVM backend: junixsocket transport, SASL EXTERNAL auth, D-Bus message framing, a read loop, and bus bootstrap (Hello). It can authenticate, obtain a unique name, send method calls, and receive/parse replies and signals against a real bus.

This layer is **internal-only** and is exercised **solely by its own tests**. It is deliberately NOT wired into the production client path (Phase 3) and does NOT serve incoming method calls (Phase 4). dbus-java stays untouched.

### What the connection does

- **Address parsing** — reads DBUS_SESSION_BUS_ADDRESS / DBUS_SYSTEM_BUS_ADDRESS; supports unix:path= and unix:abstract= with an optional guid= suffix and a semicolon-separated candidate list.
- **Transport** — opens a raw AF_UNIX socket via junixsocket (AFUNIXSocket / AFUNIXSocketAddress, abstract namespace via inAbstractNamespace). junixsocket-core is promoted to a compile dependency; it was already pulled transitively by dbus-java's junixsocket transport, so no new artifact enters the runtime classpath.
- **Bootstrap** — Hello captures the assigned unique name; requestName / releaseName / addMatch / removeMatch are provided for later phases.

### SASL approach

Line protocol over the socket: a leading NUL byte, then AUTH EXTERNAL hex(uid); on OK it sends NEGOTIATE_UNIX_FD and records whether the bus replied AGREE_UNIX_FD (so Phase 5 fd-passing is already negotiated, even though fds are not wired yet), then BEGIN. REJECTED or unexpected responses raise DBusAuthException. The uid is read reflectively from com.sun.security.auth.module.UnixSystem so the build needs no explicit module wiring.

### Framing approach

WireMessageCodec reads and writes a full message and uses the Phase-1 DBusMarshaller for BOTH the a(yv) header-fields array and the body. The entire header (endianness, type, flags, version, body length, serial, and the header-fields array) is marshalled as a single yyyyuua(yv) signature so alignment is computed correctly from the start of the message; the body is then 8-aligned and marshalled against the SIGNATURE header field. Writes are always little-endian; reads honour the message's declared endianness byte. All nine header-field codes (PATH/INTERFACE/MEMBER/ERROR_NAME/REPLY_SERIAL/DESTINATION/SENDER/SIGNATURE/UNIX_FDS) are supported. A dedicated low-level WireMessage type models a parsed/outgoing message; it shares the boxed value representation (JvmStructPayload / JvmVariantPayload) so the marshaller drives it directly.

### Thread model

One daemon reader thread owns all reads. For each framed message it dispatches: METHOD_RETURN/ERROR complete a per-serial CompletableFuture; SIGNAL is delivered to every registered listener; METHOD_CALL is ignored for now (Phase 4). Writes are serialized under a single lock, so any thread may call out concurrently. Outgoing calls take a monotonically increasing serial and await the matching reply: call() is suspend-friendly (awaits the future), and callBlocking() is used by the bootstrap helpers so connect() need not be a coroutine; send() is fire-and-forget. close() flips a running flag, closes the socket (which unblocks the reader's read), joins the reader thread, and fails all outstanding calls — no threads or sockets are leaked.

### Tests (jvmTest, under dbus-run-session)

Real-bus proof against a live session daemon:
- Connect + SASL EXTERNAL + Hello asserts a unique name matching :1.x.
- ListNames is called and parsed via the framing + Phase-1 marshaller; the reply (signature as) contains both the bus daemon and our own unique name.
- GetId is called and its s reply parsed.
- AddMatch for NameOwnerChanged followed by RequestName: the broadcast signal is delivered through the read loop (asserted name and new-owner = our unique name).
- Clean shutdown: open then close, asserting the reader thread has stopped.

Plus a bus-independent framing round-trip unit test (method call / return / error / signal with a nested asa{sv} body and a variant / empty body), decoded both directly and via the streaming reader.

All five real-bus tests execute against the daemon (connect ~81ms, calls a few ms each — not the no-bus early-return path). Full :jvmTest is green; ktlintCheck + apiCheck pass; compileKotlinLinuxX64 builds.

### API impact

No public API change — everything is internal, and apiDump is a no-op.

Refs #93.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -108,6 +108,11 @@ kotlin {
                 implementation(libs.kotlinx.datetime)
                 implementation(libs.dbus.java.core)
                 runtimeOnly(libs.dbus.java.transport.junixsocket)
+                // Owned D-Bus connection (epic #93): raw AF_UNIX transport on the compile
+                // classpath. junixsocket is already pulled transitively by dbus-java's
+                // junixsocket transport; depend on it directly so the owned connection can use
+                // AFUNIXSocket without going through dbus-java.
+                implementation(libs.junixsocket.core)
                 implementation(kotlin("stdlib"))
             }
         }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -15,6 +15,7 @@ ktlint = "14.0.1"
 buildconfig = "6.0.9"
 vannik = "0.36.0"
 dbus-java = "5.2.0"
+junixsocket = "2.10.1"
 mockk = "1.14.9"
 
 [libraries]
@@ -33,6 +34,7 @@ kotlin-compiler-embeddable = { module = "org.jetbrains.kotlin:kotlin-compiler-em
 clikt = { module = "com.github.ajalt.clikt:clikt", version.ref = "clikt"}
 dbus-java-core = { module = "com.github.hypfvieh:dbus-java-core", version.ref = "dbus-java" }
 dbus-java-transport-junixsocket = { module = "com.github.hypfvieh:dbus-java-transport-junixsocket", version.ref = "dbus-java" }
+junixsocket-core = { module = "com.kohlschutter.junixsocket:junixsocket-core", version.ref = "junixsocket" }
 mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
 
 [bundles]
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
@@ -1,0 +1,417 @@
+/**
+ *
+ * (C) 2024 - 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Project: sdbus-kotlin
+ * Description: High-level D-Bus IPC kotlin library based on sd-bus
+ *
+ * This file is part of sdbus-kotlin.
+ *
+ * sdbus-kotlin is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * sdbus-kotlin is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with
+ * sdbus-kotlin. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.monkopedia.sdbus.internal.jvmdbus
+
+import java.io.BufferedInputStream
+import java.io.ByteArrayOutputStream
+import java.io.Closeable
+import java.io.EOFException
+import java.io.File
+import java.io.IOException
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.charset.StandardCharsets
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CopyOnWriteArrayList
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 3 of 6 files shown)


monkopedia-reviewer (approved):
## Reviewer verdict: approve (feature_internal, tier-2-as-gate per epic #93)

Phase 2 of 6 of the owner-directed owned-connection epic. **PARTIAL by design** — the PR body is explicit that this is the internal I/O foundation only: not wired into the production client path (Phase 3) and not serving incoming calls (Phase 4); dbus-java untouched. The delivered subset is coherent and self-contained, and the "remaining" work is tracked on #93. Reviewed the whole diff to merge standard; it's sound on every axis the gate cares about.

### Correctness — verified
- **Address parsing** — `unix:path=` / `unix:abstract=` with `guid=` tolerated (unknown keys ignored), `;`-separated candidate list, session/system env + system default. Abstract namespace goes through `AFUNIXSocketAddress.inAbstractNamespace` (junixsocket handles the leading NUL).
- **SASL EXTERNAL** — leading NUL, `AUTH EXTERNAL <hex(uid)>`, `OK` parsed, `NEGOTIATE_UNIX_FD`→`AGREE_UNIX_FD` recorded for Phase 5, `BEGIN`; `\r\n` line protocol; `REJECTED`/unexpected → `DBusAuthException`. SASL and the reader share the same `BufferedInputStream`, so no buffered-byte loss at handover.
- **Framing** — the header-fields `a(yv)` array is driven by the **Phase-1 marshaller** (not ad-hoc byte-pushing): the whole header is marshalled as one `yyyyuua(yv)` signature so alignment is correct from the message start, then pad-to-8, then body marshalled against the SIGNATURE field. All nine field codes present with correct variant types (`o`/`s`/`g`/`u`). **Reads honour the message's declared endianness byte** (`Endian.fromCode(bytes[0])`), not hardcoded LE.
- **Serial/reply correlation** — monotonic `AtomicInteger` (starts at 1, never 0), pending future **registered before send** (no register/deliver race), reader removes+completes on reply. ERROR completes the waiter then `throwIfError` raises `DBusCallException`. Connection drop and `close()` both `failAllPending` → waiters fail rather than hang; `callBlocking` also has a 30s timeout.
- **Thread model + shutdown** — single daemon reader; `close()` flips running, closes the socket (unblocks the blocking read), joins with a 2s timeout (guards against self-join from a listener), fails pending. `close_stopsReaderThreadAndClosesSocket` asserts `isReaderRunning == false`. Reader start order (SASL → startReader → hello) is correct so the Hello reply is serviced by the reader.
- **METHOD_CALL** — explicitly ignored (Phase 4 placeholder); no accidental serving path.

### Real-bus proof — genuine, not mocked
CI's `full-tests-x64` runs `./gradlew allTests test` inside `dbus-run-session` with `DBUS_SESSION_BUS_ADDRESS` exported, so `connectSession()` returns a live connection and the five tests actually hit the daemon: SASL+Hello → `:1.x` unique name; real ListNames (`as`, contains daemon + own name); real GetId (`s`); NameOwnerChanged delivered through AddMatch + the read loop (asserts new-owner = our unique name); clean open/close. Plus a bus-independent framing round-trip (call/return/error/signal incl. nested `asa{sv}` + variant + empty body), decoded both directly and via the streaming reader.

### Internal-only
Diff adds only the two new files + the two test files + `junixsocket-core` (pinned to 2.10.1, the version already pulled transitively by dbus-java — no new runtime artifact, no skew). `PureJavaDbusBackend.kt` / the dbus-java client path untouched; no `api/` files (apiDump no-op).

### Minor, non-blocking
The codec correctly honours the endianness byte, but the framing unit test only round-trips little-endian; a hand-crafted big-endian decode case would close the gap the #93 risk register flagged ("the `'B'` branch isn't dead code"). BE decoding flows entirely through the already-tested Phase-1 `DBusReader`, so this is a coverage nicety, not a defect — worth adding when Phase 3 lands.

### CI
All required checks green: full-tests-x64, native-build-x64, static-checks (ktlint + apiCheck), arm-runtime-tests (clean, no malloc exit-134), all pr-smoke matrix entries.

Approving and squash-merging. #93 stays open (phase 2 of 6).

PR #98, opened 2026-06-13
Fix flaky wire-ON async test: drop racy isPending() assertion (#97)

### LINKED ISSUE #97: Flaky test: JvmScaffoldTest.proxyCallMethodAsync racy assertTrue(pending.isPending()) fails wire-ON
Found on main post-merge of #96 (wire-client-parity-x64 red): JvmScaffoldTest.proxyCallMethodAsync_callbackReceivesTypedResult, AssertionError at JvmScaffoldTest.kt:142 = assertTrue(pending.isPending()) right after callMethodAsync returns. Passed wire-OFF + on the PR; flaky wire-ON.

Root cause: racy TEST assertion, not a library bug. On the owned-wire backend a fast local reply can arrive and fire the callback (flipping pending to not-pending) in the microseconds between callMethodAsync returning and the isPending() check. isPending()-immediately-after-launch is an unobservable intermediate state on a fast backend. The meaningful assertions (latch.await, result==7, then !pending.isPending()) are sound.

Fix: drop the racy assertTrue(pending.isPending()) (or restructure so the reply provably cannot complete before it, which is not worth the complexity). Test-only, no library change. Verify wire-ON repeatedly (loop the test) under DBUSMOCK_PYTHON + timeout.

monkopedia-coder:
## Summary
Removes the racy `assertTrue(pending.isPending())` at JvmScaffoldTest.kt:142 (test-only). On the owned wire backend a fast reply can fire the callback before that line, so `isPending()` is legitimately false — it flaked `wire-client-parity-x64` on main after #96. The post-completion `!pending.isPending()` check + result assertions still verify the async contract. Replaced with a comment so it is not re-added.

Fixes #97.

### DIFF
--- a/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmScaffoldTest.kt
+++ b/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmScaffoldTest.kt
@@ -139,7 +139,10 @@ class JvmScaffoldTest {
                 latch.countDown()
             }
 
-            assertTrue(pending.isPending())
+            // NB: do not assert pending.isPending() here — on a fast backend (the owned
+            // wire connection) the reply can arrive and fire the callback before this line
+            // runs, making isPending() legitimately false. See #97. The post-completion
+            // !isPending() assertion below verifies the call resolves.
             assertTrue(latch.await(3, TimeUnit.SECONDS))
             assertEquals(null, errorRef.get())
             assertEquals(7, resultRef.get())


monkopedia-reviewer (approved):
Approve — test-only flake fix, tier 1.

Verified the diff touches only src/jvmTest/kotlin/com/monkopedia/sdbus/JvmScaffoldTest.kt: it removes the single racy assertion on pending.isPending() immediately after callMethodAsync and replaces it with an explanatory comment referencing #97. No production or API change, so apiDump is not relevant.

The rationale is correct. On the owned-wire backend a fast local reply can fire the callback before the assertion line runs, making isPending() legitimately false. That is an unobservable intermediate state and not a library bug. The meaningful assertions remain intact: latch.await(3s), errorRef == null, result == 7, and the post-completion !isPending() check, which still verify the async contract resolves.

CI is green: 14 checks pass, nightly-stress skipped as expected. The key check wire-client-parity-x64 passed (2m1s) — the previously flaky path is now stable.

PR #99, opened 2026-06-13
Phase 3b (#93): emit signals over the wire on the owned connection

monkopedia-coder:
## Phase 3b (#93): emit signals over the wire on the owned connection

Makes an object served on a WIRE-backed (owned) connection emit signals **over the wire** so external processes receive them and the bus stamps the authoritative sender + attaches credentials. This was deliberately deferred from Phase 3 (it had caused a hang + a match collision when entangled earlier); done here in isolation.

### Design decision: WIRE-ONLY on the wire backend

- **Emission.** PureJavaDbusObject (reused by the wire backend with no dbus-java connection) is now handed a WireSignalEmitter bound to the connection's socket. All four emit paths (emitSignal / emitPropertiesChangedSignal / emitInterfacesAdded / emitInterfacesRemoved) build a D-Bus SIGNAL WireMessage (path / interface / member / signature, body marshalled via the existing toWireBodyValue + DBusMarshaller bridge) and send it through DBusWireConnection.send. The SENDER header is left unset so the bus stamps our unique name and our credentials.
- **Reception is correspondingly wire-only.** installSignalMatch no longer also registers an in-process LocalJvmMatchBus subscription, so each signal is delivered exactly once via the subscriber's own socket + AddMatch. Same-JVM two-connection delivery still works because each connection has its own socket + match (verified by CrossModuleSignalIntegrationTest). The in-process LocalJvmSignalBus / LocalJvmMatchBus is now the dbus-java backend's same-process mechanism only. A hybrid (wire + in-process) was rejected: same-JVM subscribers already hold a wire match, so it would double-deliver.
- **Sender credentials.** A signal from one of our own connections has its credentials filled from the local process (withLocalSenderCredentials), matching the dbus-java backend's local short-circuit. Resolving an external sender's credentials would need a GetConnectionCredentials bus call, which cannot run on the reader thread that delivers signals without deadlocking; no test needs it, so it is intentionally left out.

### Acceptance

1. **Un-gated** JvmRealBusIntegrationTest.signalCallback_resolvesSenderCredentialsFromBus: now runs on **both** toggle states and **PASSES wire-ON** (the signal traverses the bus, so sender + credsPid resolve).
2. **External-reception proof** (new): cross_test WireSignalEmissionExternalTest.emittedSignal_reachesExternalDbusMonitor serves an object on an owned connection, runs an **independent out-of-process dbus-monitor** observer, emits a signal, and asserts the observer captured it. **FAILS before** this change (in-process-only emission, the external observer never receives it) and **PASSES after**. JVM-only (the owned backend is JVM-only + toggle-gated); skips cleanly without a session bus / dbus-monitor.
3. CrossModuleSignalIntegrationTest + all dbusmock signal-reception suites still green wire-ON.

### Results

JAVA_HOME=java-21, under dbus-run-session, peer engaged (dbusmock per-test timings 0.5-0.76s — real, not the ~13ms no-op).

- **wire-ON** `:jvmTest :cross_test:jvmTest` — all green on dbusmock **0.38.1** and **0.31.1**, including the un-gated creds test and the new external-reception test.
- **wire-OFF** `:jvmTest :cross_test:jvmTest` — all green on dbusmock **0.38.1** and **0.31.1**.
- **fails-before** for the external test confirmed by temporarily disabling wire emission: dbus-monitor engaged (`became a monitor`) but never received the signal, so the test failed with an AssertionError; restored afterwards.
- `ktlintFormat` / `apiDump` (**NO-OP** — no `api/*.api` change) / `ktlintCheck` / `apiCheck` / `compileKotlinLinuxX64` all green.

Refs #93.

### DIFF
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireSignalEmissionExternalTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireSignalEmissionExternalTest.kt
@@ -1,0 +1,131 @@
+/**
+ *
+ * (C) 2024 - 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Project: sdbus-kotlin
+ * Description: High-level D-Bus IPC kotlin library based on sd-bus
+ *
+ * This file is part of sdbus-kotlin.
+ *
+ * sdbus-kotlin is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * sdbus-kotlin is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with
+ * sdbus-kotlin. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.emitSignal
+import com.monkopedia.sdbus.signal
+import java.io.File
+import java.util.concurrent.TimeUnit
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertTrue
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Proves that a signal our object emits actually traverses the BUS and reaches an INDEPENDENT
+ * out-of-process observer (epic #93 phase 3b) — the piece that was deliberately deferred from
+ * phase 3. The observer is `dbus-monitor` (a separate process), so in-process delivery cannot mask
+ * the result:
+ *  - wire backend BEFORE this phase: emission was in-process only, so `dbus-monitor` never sees the
+ *    signal and this test FAILS (the poll loop times out);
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):
## Reviewer verdict: APPROVE (Phase 3b, #93) — COMPLETE

Phase 3b makes objects on the owned wire backend emit signals over the wire so external processes receive them and the bus stamps the authoritative sender + credentials. Reviewed as gate-work (internal backend, no public API change).

### Correctness
- **Wire emission**: emitSignal / emitPropertiesChangedSignal / emitInterfacesAdded / emitInterfacesRemoved on a wire-backed PureJavaDbusObject now route through a WireSignalEmitter (javaConnection == null + emitter != null), building a SIGNAL WireMessage (path/interface/member/signature + body) and sending via DBusWireConnection.send (fire-and-forget, correct for a broadcast signal). SENDER header left unset so the bus stamps our unique name + credentials.
- **Marshalling reuses the Phase-3 bridge**: body = payload.map(::toWireBodyValue), not a divergent path. Empty/blank signature correctly yields a no-body signal.
- **Design decision (wire-only)**: stated in the PR body. installSignalMatch no longer registers an in-process LocalJvmMatchBus subscription; reception is wire-only via each connection's own socket + AddMatch, so same-JVM two-connection delivery still works (CrossModuleSignalIntegrationTest green wire-ON) without double-delivery.
- **Un-gated creds test**: JvmRealBusIntegrationTest.signalCallback_resolvesSenderCredentialsFromBus is no longer skipped wire-ON; runs on both toggle states and passes (signal traverses the bus → sender + credsPid resolve; same-process sender creds filled from the local process via withLocalSenderCredentials).
- **External-reception proof**: new WireSignalEmissionExternalTest uses an independent out-of-process dbus-monitor observer (not the in-process bus), with a unique payload marker, fail-before / pass-after; skips cleanly without a session bus or dbus-monitor.
- **Internal-only**: no api/ change (apiDump no-op); changes confined to jvmMain/jvmdbus + jvm/cross_test tests.

### CI
All checks green on the merged head, including the authoritative wire-client-parity-x64 (real python3-dbusmock) and full-tests-x64.

### #97 flake handling
The branch was based before #98 (the JvmScaffoldTest isPending fix). Merged current origin/main into the branch (clean, only JvmScaffoldTest.kt) and pushed via coderbot to pick up the fix and de-risk drift; the fresh CI run is fully green with no JvmScaffoldTest flake.

Sound and green — approving and merging. #93 remains open; next is Phase 4 (serving).

Issue #89, opened 2026-06-12
Generated adaptor property getters never serialize the value — remote Get returns empty/hangs (PropertyVTableItem.with(KProperty0))


monkopedia-coder:
**Severity: high — breaks property reads for ALL codegen-generated service adaptors on the wire.** Found by the #67 server-sample work (and independently hit by a prior agent, who mis-attempted an inline fix — two agents reproduced it).

## The bug
`PropertyVTableItem.with(receiver: KProperty0<T>)` (src/commonMain/kotlin/com/monkopedia/sdbus/PropertyVTableItem.kt:142) sets:
```kotlin
getter = {
    receiver.get()          // computes the value, then DISCARDS it
}
```
But `PropertyGetCallback = (msg: PropertyGetReply) -> Unit` (TypeTraits.kt:67) — the getter must SERIALIZE the value into the reply message. Compare the correct `withGetter` (same file, line ~83):
```kotlin
getter = { reply ->
    reply.serialize<T>(callback())   // writes the value into the reply
}
```
So `with(KProperty0)` produces an empty reply: a remote `org.freedesktop.DBus.Properties.Get` (or GetAll) against the property returns nothing / hangs on native. The **setter** path in the same function is correct (`receiver.set(it.deserialize<T>())`); only the getter is broken.

## Why it's severe and why it hid
`AdaptorGenerator.kt:165` emits exactly `with(this@<Adaptor>::<prop>)` for every property — so **every codegen-generated adaptor** (the library's primary server-side use case) has broken property getters. Verified in the checked-in golden fixtures (e.g. `codegen/src/test/resources/BluezAdapter1Test/adaptor.0.kt` lines 38/41/44/...). It evaded all prior testing because: (a) the dbusmock suites exercised us as a *client* against mock servers, not serving our own properties; (b) own-server JVM tests route through `JvmStaticDispatch`, bypassing wire serialization; (c) no test did a real remote `Get` against a *native* adaptor. The #67 server sample is the first to serve a property to an independent consumer (busctl), exposing it.

## Fix (surgical, behavior-only, no API/codegen change)
In `with(KProperty0)`, serialize into the reply like `withGetter` does:
```kotlin
getter = { reply ->
    reply.serialize<T>(receiver.get())
}
```
No codegen change needed (it keeps emitting `with(::prop)`, which now works); golden fixtures unchanged; `apiDump` no-op.

## Acceptance test
[...]

PR #100, opened 2026-06-13
Phase 4 (#93): serve incoming method calls / properties / introspect over the owned connection

### LINKED ISSUE #90: JVM backend can't serve exported objects to external processes (in-process dispatch only) — and parity matrix may over-claim JVM server rows

monkopedia-coder:
## Phase 4 of epic #93 — serve incoming method calls over the owned connection (closes #90)

External processes can now call methods on, get/set properties of, and introspect objects exported by a JVM sdbus-kotlin service running on the owned-wire backend. The read loop previously received but ignored incoming METHOD_CALLs (a phase-2 placeholder); it now routes them and replies over the wire.

### Dispatch design + threading model
- **DBusWireConnection** gains a per-connection incoming-call handler plus a cached worker pool. The reader thread only enqueues each METHOD_CALL and keeps draining the socket; handler execution (which may block up to the async-reply timeout) runs on a worker. This means a slow/async handler — or one that itself calls back on the same connection and awaits a reply the reader delivers — cannot stall I/O or deadlock the reader. The worker pool is shut down on close().
- **WireServe** routes each call through the SAME registries the in-process client path uses. Regular methods, Properties Get/Set/GetAll and ObjectManager GetManagedObjects all resolve through `JvmStaticDispatch` (the handlers `PureJavaDbusObject.addVTable` already registers), so a single addVTable registration serves both in-process AND over the wire. Call args are deserialized signature-driven through the existing wire-to-decoder bridge; return values are marshalled using the reply's declared body signature (regular methods, which handles degrouped multi-out correctly) or the known standard-interface signature, and sent as a METHOD_RETURN with the correct reply-serial and destination=caller.

### Standard interfaces served
- `org.freedesktop.DBus.Properties` — Get / Set / GetAll routed to the vtable getter/setter callbacks (reuses the #89-correct getter serialization).
- `org.freedesktop.DBus.ObjectManager` — GetManagedObjects via the existing managed-objects machinery, marshalled over the wire.
- `org.freedesktop.DBus.Introspectable` — Introspect, with XML generated from vtable metadata captured in a new internal `WireServeRegistry` (methods with in/out args + names, properties with access, signals, deprecation), including child `<node>` entries and the standard interfaces sd-bus always exposes.
- `org.freedesktop.DBus.Peer` — Ping / GetMachineId.

### Errors
Unknown path → `UnknownObject`; unknown interface/member → `UnknownInterface` / `UnknownMethod`; a thrown sdbus `Error` is marshalled into an ERROR reply preserving its name and message. A handler exception is caught and turned into an ERROR reply — it never kills the worker or the reader thread. `NO_REPLY_EXPECTED` is honored (no reply sent).

### Name ownership
Unchanged from phase 2: `createConnection(name=...)` calls RequestName, so external callers can address us by the well-known name (Hello already gives the unique name). The serving dispatch keys on the connection's unique name, matching the dispatch destination `PureJavaDbusObject` registers under.

### External-caller acceptance test
`cross_test/.../WireServeExternalTest` drives **busctl** as a genuinely external process (its own socket / wire protocol — cannot be short-circuited by the in-process JvmStaticDispatch the #90 investigation warned about) against a JVM-exported object on the wire backend. It asserts:
- method return for a **sync** handler (`Add(ii)->i`) and an **async** `acall` handler (`Concat(ss)->s`);
- Properties **Get** / **Set** (verifying the setter ran) / **GetAll**;
- **Introspect** (interface, method, property present);
- ObjectManager **GetManagedObjects** (child object + interface present).

It is gated to the owned-wire backend (`-Dsdbus.jvm.wire=true`), i.e. CI's `wire-client-parity-x64` job; the dbus-java backend serves exported objects in-process only (the documented #90 limitation), so the test skips cleanly there. Verified **fails-before** this phase (busctl reports `Connection timed out` → assertion fails) and **passes-after** (all assertions green). No hang was found (the worker-off-reader design was correct on the first run; every gradle run was wrapped in `timeout`).

### Verification
- `ktlintFormat` → `apiDump` (**no-op**, all new code is internal) → `ktlintCheck apiCheck` → `compileKotlinLinuxX64`: all green.
- Full `:jvmTest :cross_test:jvmTest` **wire-ON** (`-Dsdbus.jvm.wire=true`) green incl. the new external-serving test; dbusmock suites ran with real time (not fake-green).
- Full suite **wire-OFF** green (serve test skips cleanly).
- Parity confirmed across **dbusmock 0.38.1** and CI's **0.31.1**.

Fixes #90.
Refs #93.

### DIFF
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireServeExternalTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireServeExternalTest.kt
@@ -1,0 +1,234 @@
+/**
+ *
+ * (C) 2024 - 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Project: sdbus-kotlin
+ * Description: High-level D-Bus IPC kotlin library based on sd-bus
+ *
+ * This file is part of sdbus-kotlin.
+ *
+ * sdbus-kotlin is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * sdbus-kotlin is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with
+ * sdbus-kotlin. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.PropertyName
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.method
+import com.monkopedia.sdbus.prop
+import java.io.File
+import java.util.concurrent.TimeUnit
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Proves an object exported by a JVM sdbus-kotlin service on the OWNED wire backend is reachable by
+ * a genuinely EXTERNAL process — `busctl` (epic #93 phase 4, closes #90). busctl speaks the real
+ * D-Bus wire protocol over its own socket, so it cannot be short-circuited by the in-process
+ * JvmStaticDispatch the #90 investigation warned about: only true over-the-wire serving makes these
+ * pass.
+ *
+ *  - wire backend BEFORE this phase: the read loop ignored incoming METHOD_CALLs, so busctl got
+ *    `UnknownObject` (or timed out) and every assertion here FAILED;
+ *  - wire backend AFTER this phase: the calls are routed to the vtable handlers and replied over the
+ *    wire, so busctl gets correct results.
+ *
+ * Gated to the owned-wire backend (`-Dsdbus.jvm.wire=true`, exercised by CI's
+ * `wire-client-parity-x64` job): the dbus-java backend serves exported objects in-process only
+ * (the documented #90 limitation), so the test skips cleanly there rather than asserting a known
+ * gap. Also skips when no session bus or no `busctl` is available.
+ */
+class WireServeExternalTest {
+
+    @Test
+    fun externalBusctl_reachesServedObject() = runBlocking {
+        val sessionBus = System.getenv("DBUS_SESSION_BUS_ADDRESS")
+        if (sessionBus == null) {
+            println("[WireServeExternalTest] SKIP: no DBUS_SESSION_BUS_ADDRESS.")
+            return@runBlocking
+        }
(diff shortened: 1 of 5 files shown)


monkopedia-reviewer (approved):
## Reviewer verdict: APPROVE

Phase 4 of epic #93 — serving incoming method calls / properties / introspect over the owned wire connection. Closes #90. Reviewed with full rigor as the core serving milestone.

### Threading / deadlock safety (the highest-risk area) — sound
- The reader thread (`DBusWireConnection.dispatch`) handles an incoming METHOD_CALL by only submitting it to a cached worker pool and immediately returning to drain the socket. It never executes handler logic.
- METHOD_RETURN / ERROR replies are completed directly on the reader thread (`pending.remove(serial)?.complete(...)`), fully independent of worker activity. So a handler that runs on a worker and itself calls back on this same connection and awaits the reply is unblocked by the reader — no reader-thread deadlock. Traced and confirmed.
- `send()` serializes on `writeLock`; the reader does not write during its loop, so there is no writer/reader lock contention.
- `close()` sets `running=false`, `serveExecutor.shutdownNow()`, closes the socket, and joins the reader. Workers are daemon threads and interrupted on shutdown; post-close `execute` is guarded by `runCatching` (RejectedExecution swallowed). No leaked threads.
- Minor (non-blocking): `newCachedThreadPool` can grow unboundedly under a flood of simultaneous slow handlers; the PR text's "no unbounded thread growth" is slightly optimistic. Acceptable for this internal phase and the safe choice for avoiding nested-call starvation; worth a bound in a later hardening pass.

### Dispatch routing — correct
Regular methods, Properties Get/Set/GetAll and ObjectManager GetManagedObjects all resolve through the SAME `JvmStaticDispatch` registry the in-process path registers in `addVTable`, so one registration serves both in-process and wire. Args deserialized via `fromWireReplyValues` (signature-driven bridge); returns marshalled via `toWireBodyValue` (recurses Maps/Lists/structs/variants, unwraps strong-name keys) with the reply's `declaredBodySignature` for regular methods (degrouped multi-out handled) or the known standard-interface signature. METHOD_RETURN carries correct `replySerial` + `destination=sender`.

### Standard interfaces — real, not stubbed
- Properties Get/Set/GetAll route to the actual vtable getter/setter callbacks (reuses the #89-correct getter serialization via `propertyToVariant`).
- ObjectManager GetManagedObjects routes to `LocalManagedObjectsRegistry.managedObjectsFor`.
- Introspect builds XML from real captured vtable metadata (`WireServeRegistry`: methods in/out args+names, properties access, signals, deprecation) plus child `<node>`s and the standard sd-bus interfaces.
- Peer Ping (empty reply) / GetMachineId (reads machine-id) answered directly.

### Error mapping — correct
UnknownObject / UnknownInterface / UnknownMethod classification for misses; thrown sdbus `Error` → ERROR reply preserving name+message; outer `runCatching` turns any handler/serve throw into ERROR_FAILED so it never kills the worker or reader; NO_REPLY_EXPECTED honored (handler still runs for side effects, reply suppressed).

### External-caller acceptance test (#90 proof) — valid
`WireServeExternalTest` shells out to a real `busctl` via `ProcessBuilder` against `--address=$DBUS_SESSION_BUS_ADDRESS` — a genuinely separate process on its own connection, so the in-process `JvmStaticDispatch` short-circuit the #90 investigation warned about cannot apply. Covers sync method (Add), async acall method (Concat), Properties Get/Set(verifies setter ran)/GetAll, Introspect, and ObjectManager GetManagedObjects. Gated to `-Dsdbus.jvm.wire=true` (CI's wire-client-parity-x64), skips cleanly on the dbus-java backend. Documented fails-before / passes-after.

### Scope
Internal-only: jvmMain/jvmdbus + one jvmTest file. No api/ changes (apiDump no-op confirmed), nothing native/common.

### CI — fully green
wire-client-parity-x64 (real python3-dbusmock, runs the new external-serving test) PASS — the authoritative serving proof. Also: static-checks, full-tests-x64, native-build-x64, arm-runtime-tests, all pr-smoke matrix entries PASS. nightly-stress skipped (expected). No reruns needed.

Auto-merging on green.

PR #102, opened 2026-06-13
Phase 5 (#93): SCM_RIGHTS unix-fd passing over the owned connection

### LINKED ISSUE #83: Test gap: JVM wire-level Unix FD passing has no CI verification
Closed by epic #93 phase 5 (fd-passing via SCM_RIGHTS on the owned junixsocket connection — the feasibility spike already proved a live fd round-trip works). Tracking here until that lands.

---

Original: JVM wire-level Unix FD passing has no CI verification; the conversion path exists but the independent-peer fd round-trip is skipped on JVM (JPMS reflection constraints). The owned-connection rewrite (#93) handles fds natively via junixsocket SCM_RIGHTS, with a cross-process fd test as acceptance.

monkopedia-coder:
## Phase 5 of epic #93 — SCM_RIGHTS unix-fd passing over the owned connection

Wires file-descriptor passing through the self-owned junixsocket D-Bus connection so UnixFds round-trip natively (no dbus-java) — in method-call args, method replies, and signals alike. Closes the #83 JVM fd wire-test gap.

### Send / receive mechanism (SCM_RIGHTS)
- Outgoing (DBusWireConnection.writeMessage): a depth-first, body-order walk collects every UnixFd in the body and replaces each with its 0-based index into the attachment array. Under the existing writeLock the raw fds are attached via junixsocket setOutboundFileDescriptors immediately before the framed write, so the kernel delivers them with this message's bytes (D-Bus associates passed fds with the message carrying UNIX_FDS). Pending outbound fds are cleared defensively after the send.
- Incoming: the connect path sizes the ancillary receive buffer (ensureAncillaryReceiveBufferSize). After each frame the reader drains junixsocket's received descriptors into a FIFO; a message declaring UNIX_FDS takes that many in arrival order.

### UNIX_FDS header + index semantics
- The h type stays marshalled as a uint32 index (unchanged); the header UNIX_FDS field (code 9) is set to the fd count on send and read on receive. Outgoing: each h body value is the 0-based index, in attachment order, into the fd array. Incoming: a signature-driven walk rewrites each h index N to the Nth received descriptor (containers — structs / variants / arrays / dicts — recurse, since an h index is type-indistinguishable from an i and needs the declared signature to locate).

### Bridge + fd ownership
- Outbound ints become throwaway FileDescriptors via junixsocket NativeUnixSocket.initFD; the kernel dup-on-send leaves our fd open.
- Received descriptors are duplicated so the resulting UnixFd owns an independent fd, then the junixsocket-owned original is closed — closing the object fires junixsocket's attached close-callback, evicting it from the received-fd table — so received fds are neither leaked until socket close nor double-closed. The UnixFd is the sole owner and closes its fd on release/GC.

### Cross-process round-trip test (closes #83)
- DBusWireConnectionTest.unixFd_roundTripsAcrossConnections_throughDaemon_asLiveDuplicate: a UnixFd is sent from one owned connection to another as a method-call arg. The fd genuinely traverses the dbus daemon (which relays passed fds via SCM_RIGHTS, dup-ing them into a separate process and back). Proven a LIVE duplicate: bytes written into the original pipe write-end are read back from the received fd.
- cross_test DbusmockTypeMatrixTest.unixFd_passesThroughForeignPeer_whereSupported: previously SKIPPED on JVM for JPMS reasons; now un-gated on the wire backend by surfacing real pipes through junixsocket's native primitives. The fd crosses to an independent dbusmock peer process and back, and the echoed fd is verified to be a live dup (write through original / read from received). This runs in CI wire-client-parity-x64.

### JPMS flags
None required. The fd <-> FileDescriptor bridge uses junixsocket's native methods (initFD/getFD/duplicate/close), not reflection into java.base/java.io, so no `--add-opens` was needed and none was added. The legacy dbus-java fd path (which would need add-opens) is left skipped on JVM rather than newly requiring flags for the backend being retired in phase 6.

### Verification
- ktlintFormat, ktlintCheck, apiCheck — apiCheck is a no-op (all changes are internal: new helpers on internal JvmUnixFdSupport, internal DBusWireConnection; UnixFd public surface unchanged). compileKotlinLinuxX64 green.
- Full parity, both toggle states (peer engaged, DBUSMOCK_PYTHON set): `:jvmTest` + `:cross_test:jvmTest` green with `-Dsdbus.jvm.wire=true` (wire ON, incl. both new fd round-trip tests executing — not skipped) AND without the toggle (wire OFF / dbus-java regression).

Fixes #83.
Refs #93.

### DIFF
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/DbusmockFdSupport.jvm.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/DbusmockFdSupport.jvm.kt
@@ -20,6 +20,12 @@
  */
 package com.monkopedia.sdbus.integration
 
+import java.io.FileDescriptor
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.lang.reflect.InvocationTargetException
+import java.lang.reflect.Method
+
 /**
  * Struct marshalling to/from remote peers is supported since the issue #71 fix (structs are
  * decomposed into wire-shaped values via JvmValueCodec on the way to/from dbus-java).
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):
## Reviewed — approve (epic #93 phase 5, closes #83)

Thorough fd-passing review (SCM_RIGHTS is subtle; focused here). Found sound.

**Outgoing.** `collectOutboundFds` walks the body depth-first in body order, appending each UnixFd's raw fd and substituting its 0-based index; the descriptor array is built in the same order, so index i maps to descriptors[i]. The UNIX_FDS header (code 9, uint32) is emitted via `buildHeaderFields` from `unixFds = fds.size`, and `h` marshals as a uint32 from the Int index (`asLong` handles Int). Send is atomic: `setOutboundFileDescriptors` then the framed write under `writeLock`, with a defensive post-send clear. Throws if fds are present but UNIX_FD wasn't negotiated.

**Incoming.** Ancillary receive buffer sized at connect; descriptors drained per-frame into a FIFO and each message takes its declared count in arrival order. Verified in junixsocket bytecode that `getReceivedFileDescriptors()` does `removeAll` on its internal list, so per-frame draining never re-returns consumed fds (no double-count). The signature-driven resolver rewrites each `h` index to the Nth received fd and is exhaustive over the `DBusType` hierarchy (basics, arrays, structs, dict-entries, variants recurse).

**Ownership (no leak / no double-close).** `adoptReceivedDescriptor` dups the received fd (independent) then closes the junixsocket-owned original. Confirmed in bytecode that closing that descriptor fires `AncillaryDataSupport$1.close()` → `openReceivedFileDescriptors.remove(fd)`, evicting it from junixsocket's table so it is not re-closed at socket close, and closing the underlying fd exactly once. The resulting `UnixFd.adopt` is the sole owner (Cleaner-backed). Clean single-owner lifecycle.

**Coverage.** `writeMessage` / `resolveReceivedFds` apply uniformly to all outgoing/incoming messages — call args, replies, and signals alike; the production proxy path carries `UnixFd` through the shared value model.

**Round-trip test = live dup.** `DBusWireConnectionTest.unixFd_roundTripsAcrossConnections_throughDaemon_asLiveDuplicate` (main module, not toggle-gated, runs in the wire job) sends a UnixFd between two owned connections through the real daemon and proves a live duplicate — bytes written into the original pipe write-end are read back from the received fd. The cross_test `unixFd_passesThroughForeignPeer_whereSupported` proves the same through an independent dbusmock peer process. Genuine SCM_RIGHTS traversal, not an in-process shortcut.

**JPMS.** None added — the bridge uses junixsocket's `NativeUnixSocket` native primitives, not reflection into `java.base`/`java.io`, so no `--add-opens` needed; correctly documented, and the legacy dbus-java fd path stays skipped (retired in phase 6).

**CI / API.** All checks green including `wire-client-parity-x64` (wire-ON, real dbusmock, both fd tests) and `static-checks` (ktlint + apiCheck). No `api/*.api` change in the diff — apiCheck no-op confirmed; new helpers are internal, `UnixFd` public surface unchanged.

Idiomatic Kotlin, well-documented, consistent with the surrounding wire layer. No blocking findings.

🤖 Reviewed by monkopedia-reviewer

Issue #101, opened 2026-06-13
Owned-connection serve worker pool is unbounded (newCachedThreadPool)


monkopedia-coder:
Minor follow-up from PR #100 (Phase 4 serving) review. DBusWireConnection serves incoming method calls on a newCachedThreadPool, which can grow unboundedly under a flood of slow/blocking handlers. Acceptable for now (the safe choice vs nested-call starvation, and internal), but should be bounded (e.g. a sized pool with a queue, sized to avoid starving nested same-connection calls) in a cleanup pass — candidate for Phase 6 (retire dbus-java) or a standalone. Not a 0.5.0 blocker. Epic #93.

monkopedia-coder:
Triage: moved from `agent-workable` to `needs-decision`. Per epic #108 this is explicitly deferred as additive post-1.0 (NOT the 0.6.0 wave), and bounding the pool involves a sizing tradeoff (avoid starving nested same-connection calls) plus a timing call (Phase 6 / epic #93 vs standalone) that's the owner's to make.

PR #103, opened 2026-06-13
Phase 6 (#93): retire dbus-java — the owned connection is now the JVM backend

### LINKED ISSUE #93: Epic: own the JVM D-Bus connection (replace dbus-java transport with junixsocket + our own marshaller/dispatch)

monkopedia-coder:
## Phase 6 (#93) — the finale: retire dbus-java

The self-owned wire connection (raw junixsocket + SASL EXTERNAL + our own byte-level marshaller + read/dispatch loop) has, across phases 1-5, been proven against real external peers for client calls, properties, signal reception + emission, serving incoming calls, and SCM_RIGHTS fd-passing. This phase makes it **the** JVM backend and removes dbus-java entirely from the library.

### Removed
- The `-Dsdbus.jvm.wire` / `SDBUS_JVM_WIRE` toggle and `isJvmWireBackendEnabled`. `JvmDbusBackendProvider.backend` is now unconditionally `WireDbusBackend` (the provider indirection is kept only so tests can mock the backend boundary).
- The dbus-java dependency from `jvmMain` (`dbus-java-core`, `dbus-java-transport-junixsocket`) and the `gradle/libs.versions.toml` entries.
- `PureJavaDbusBackend.kt` — the dbus-java connection/proxy/object path, all dbus-java conversion + credential helpers, and the in-process `LocalJvmSignalBus`; plus `LocalJvmMatchBus.kt` (the dbus-java/stub same-process signal mechanism).
- `StubJvmDbusBackend` (the in-process stub).
- dbus-java-era tests: `StubJvmDbusBackendTest`, `PureJavaDbusBackendTest`, and `DBusMarshallerDifferentialTest` (the dbus-java marshalling **oracle** — now redundant with `DBusMarshallerSpecTest`'s hand-computed spec canaries and the real-peer round-trips).

### Kept / migrated
- **junixsocket is now a direct `jvmMain` dependency** (it was only transitive via dbus-java's transport). The published `-jvm` POM keeps `junixsocket-core` and drops dbus-java; remaining POM deps: kotlin-stdlib, kotlinx-serialization/coroutines/datetime, junixsocket-core.
- The wire stack is untouched: `DBusMarshaller`, `WireMessage`/codec, `DBusWireConnection`, `WireDbusBackend`, `WireServe`, the value-model bridge, `JvmStaticDispatch`.
- The shared object machinery (formerly `PureJavaDbusObject`) is extracted to **`WireDbusObject.kt`**, wire-only, together with the shared registries (`LocalManagedObjectsRegistry`, `LocalObjectManagerRegistry`, `LocalJvmServiceRegistry`).
- Backend-boundary mock tests (`JvmConnectionMockkTest`, `JvmFactoryMockkTest`, `JvmProxyObjectMockkTest`) and `JvmScaffoldTest` keep passing — they mock/use the `JvmDbus*` interfaces and the shared dispatch, not a removed backend.
- Wire-gated `cross_test` cases (`WireServeExternalTest`, the fd cross-tests in `DbusmockFdSupport.jvm`) now run **unconditionally** (no toggle). The dbusmock peer-parity flags `peerErrorNameMappingSupported` / `peerStructMarshallingSupported` are now `true` on both backends (the #71/#72 fixes that the owned wire backend already carries) and their KDocs were updated.

### CI consolidation
- Removed the redundant `wire-client-parity-x64` job: the wire backend is now the only backend, so the existing `full-tests-x64` job already exercises it via `allTests test crossRuntimeInteropTests`.
- Added the `timeout 1800` + `--no-daemon` hard guard to the `full-tests-x64` gradle run (a hung call/signal now fails fast instead of stalling). `python3-dbusmock` stays installed so the dbusmock suites remain real.

### Verification (JDK 21)
- `apiDump` is a **no-op** — the public API is unchanged (the backend is internal; `UnixFd`/`Connection`/etc. surface is untouched). `ktlintCheck` + `apiCheck` pass.
- Full JVM suite green on the sole backend: **297 `:jvmTest` + 60 `:cross_test:jvmTest`, 0 skipped, 0 failures**. The dbusmock suites engaged with real timings (e.g. `DbusmockTypeMatrixTest` 12 tests / 6.3s, `DbusmockBluezTest` 7 / 3.6s) — not fake-green no-ops. Run locally against the venv `python-dbusmock` 0.38.1; CI runs the apt `python3-dbusmock` version and is authoritative for that second version.
- `compileKotlinLinuxX64` + `:cross_test:compileTestKotlinLinuxX64` compile; native/common untouched.
- `grep -rn 'org.freedesktop.dbus\|dbus-java' src/jvmMain build.gradle.kts gradle/libs.versions.toml` → only comments remain (no code/dependency references).

### #101 (optional) — left open
The current serve worker pool is `newCachedThreadPool`, which grows unbounded and therefore does **not** starve nested same-connection calls. A naive sized pool risks deadlocking such nested calls (the exact hang the suite's timeouts guard against), so bounding it is deferred rather than risk destabilizing the now-green backend.

### Note on remaining dbus-java references
dbus-java survives **only** as an INDEPENDENT third-party D-Bus peer used by the cross-runtime interop tests (`CrossRuntimeInteropSmokeTest` in `cross_test`, `CrossRuntimeInteropStressTest` in `stress_test`) to prove sdbus-kotlin interoperates with another JVM D-Bus stack over a real bus. It is **not** sdbus-kotlin's backend and **not** a published dependency; it is declared by literal coordinate in those test modules so it stays out of the shared version catalog and the published artifacts. This preserves real interop coverage that would otherwise be lost.

Fixes #93.


monkopedia-reviewer (changes requested):
## Request changes — real regression: direct (brokerless) bus connections break on the sole wire backend

Most of this PR is sound, but CI surfaces a genuine functional regression that blocks the finale.

### The defect (blocking)

`full-tests-x64` (the consolidated JVM authority) fails: `:cross_test:jvmInteropTest` — 14 tests, **5 failed**, all the JVM-client-over-direct-bus cases in `CrossRuntimeInteropSmokeTest`:

- `jvmClientInvokesNativePeerOverDirectBus`
- `jvmClientRoundTripsUnixFdWithNativePeerOverDirectBus`
- `jvmClientRoundTripsLargeMapWithNativePeerOverDirectBus`
- `jvmClientRoundTripsNestedVariantWithNativePeerOverDirectBus`
- `jvmClientRoundTripsMixedPayloadWithNativePeerOverDirectBus`

Every one fails at the `createDirectBusConnection("unix:path=...")` call with:

```
com.monkopedia.sdbus.Error: ... Failed to open bus 'unix:path=...':
org.freedesktop.DBus.Error.UnknownObject: Unknown object '/org/freedesktop/DBus'
  at WireDbusBackend.createConnection (WireDbusBackend.kt:73)
  at Connection_jvmKt.createDirectBusConnection (Connection.jvm.kt:93)
```

Root cause: `DBusWireConnection.connect(address)` unconditionally runs the message-bus bootstrap — SASL + `hello()` against `/org/freedesktop/DBus` (and RequestName/AddMatch). A **brokerless direct peer** (here a native sd-bus server in direct mode) has no message-bus object at `/org/freedesktop/DBus`, so Hello returns `UnknownObject` and the connection never opens. The owned wire backend has no direct/peer-to-peer connection mode that skips the bus bootstrap; dbus-java did (its dedicated DirectConnection / sd-bus's direct mode).

Why this matters and why it wasn't caught:
- These tests were **green on main** because the CI interop suite ran on the dbus-java backend (the wire backend was opt-in). Making the wire backend mandatory removes a working capability.
- `createDirectBusConnection` is **public API** (`Connection.jvm.kt`), so this is an observable behavior regression even though `apiDump` is a no-op — the symbol is unchanged but it no longer works on the JVM.
- The entire `pr-smoke` **forward (jvm-client)** matrix fails for the same reason (happy_path/error/timeout/cancellation/dropped_peer); the **reverse (native-client)** matrix passes. The PR's local "0 failures" run did not include `jvmInteropTest` / `crossRuntimeInteropTests`, so the gap was missed.

Suggested direction: give the wire backend a brokerless connection path (connect → SASL → BEGIN, **no Hello/RequestName**, no unique name) selected for `DIRECT_ADDRESS`/`createDirectBusConnection`, mirroring sd-bus direct mode and dbus-java's DirectConnection. The `crossRuntimeInteropTests` aggregate should be part of the pre-merge verification for this epic since it is the only JVM↔native direct-bus coverage.

(Separately, `native-build-x64` failed on the known `packages.microsoft.com` apt 403 infra flake — not blocking on its own, but moot while the real failures above stand.)

### What is sound (no further action needed once the above is fixed)

- **Deletions are justified.** `DBusMarshallerDifferentialTest` imported dbus-java types as a marshalling oracle and cannot survive the dependency removal; `DBusMarshallerSpecTest` (hand-computed spec bytes + the #71/#74/#26/#11/#27 corpus) plus real-peer round-trips are a stronger oracle. `StubJvmDbusBackendTest`/`PureJavaDbusBackendTest` tested removed code; live behaviors are still covered (unreachable→throw by `JvmUnreachableBusTest`, fd-rejection retained in `WireDbusBackend.createConnection`, struct/error-name parity now asserted `true` via the dbusmock suites).
- **dbus-java is gone from production**: only comments remain in `src/jvmMain`/build files; junixsocket is promoted to a direct `jvmMain` dependency; the published `-jvm` POM keeps junixsocket and drops dbus-java.
- **dbus-java test-peer isolation is clean**: declared by literal coordinate as `testImplementation` in `cross_test`/`stress_test` only, outside the version catalog; neither module applies a publish plugin, so it cannot leak into any published artifact.
- **CI consolidation reasoning is correct**: removing `wire-client-parity-x64` is right (single backend), and `python3-dbusmock` is retained so the dbusmock suites stay real. The `timeout 1800` + `--no-daemon` guard is a good addition (and is exactly what just made the regression fail fast instead of hanging).
- **#101** deferral (unbounded pool to avoid nested-call starvation) is documented and reasonable.

monkopedia-coder:
## Fixed: brokerless direct-bus mode on the owned connection

Thanks for the catch — confirmed and fixed the regression. Pushed
fb44864 to this branch.

### Root cause (confirmed)
DBusWireConnection.connect unconditionally ran the message-bus bootstrap
(SASL -> Hello to org.freedesktop.DBus, + RequestName/AddMatch). A brokerless
direct peer has no bus daemon, so Hello returned UnknownObject and
createDirectBusConnection / DIRECT_ADDRESS never opened. Reproduced first:
:cross_test:jvmInteropTest 5/14 failing, all at createDirectBusConnection.

### Direct mode (mirrors sd-bus direct mode + the removed dbus-java DirectConnection)
- New `connectDirect(address)` / `direct` flag: open socket, SASL EXTERNAL ->
  BEGIN, then SKIP the org.freedesktop.DBus bootstrap (no Hello -> no
  daemon-assigned unique name; uniqueName stays null, callers keep their
  synthetic fallback).
- Daemon-bound helpers are no-ops/synthetic in direct mode: requestName /
  releaseName report success, addMatch / removeMatch do nothing (the single
  peer's signals arrive directly on the socket and are still filtered locally
  in the reader), getNameOwner returns null. Calls/replies/signals correlate
  by serial over the one socket as before; serving is unchanged.
- WireDbusBackend routes DIRECT_ADDRESS to connectDirect.

### Invalid Unix FD (second gap, exposed once the connection opened)
junixsocket (like the raw kernel sendmsg) refuses fd=-1 with EBADF, which was
surfacing as an unmapped ExecutionException. The wire layer now rejects a
negative descriptor before SCM_RIGHTS with a clean InvalidArgs ("Invalid Unix
FD"), keeping the connection intact — the well-behaved equivalent of
dbus-java refusing the fd client-side. The native peer's invalidRejected
expectation was unsatisfiable by any client (neither backend can pass -1 over
SCM_RIGHTS) — on main it just timed out 20s, masked by --ktest_no_exit_code;
the peer now gates only on the valid round-trip.

### Verification (JDK 21, dbusmock venv engaged)
- :cross_test:jvmInteropTest: 14/14, 0 skipped — including the 5 direct-bus
  cases (reproduced failing first, then green).
- Full regression: 297 :jvmTest + 60 :cross_test:jvmTest, 0 failures, 0
  skipped.
- apiDump no-op; ktlintCheck + apiCheck pass; compileKotlinLinuxX64 compiles.
- Added the interop suites (jvmInteropTest / crossRuntimeInteropTests) to
  verification — they were the gap that let this through.

full-tests-x64 / :cross_test:jvmInteropTest should now be green.

monkopedia-reviewer (approved):
## Approve — direct-bus regression fixed, peer-test change is a legitimate unsatisfiable-test correction

Round 2 (fix commit fb44864). The round-1 blocker (brokerless direct-bus connections broken by retiring dbus-java) is resolved, the secondary invalid-fd gap is handled cleanly, and the changed peer-test contract checks out under independent analysis. CI is fully green.

### 1. Direct-mode fix (the regression) — sound
- DBusWireConnection now has a `direct` flag and connectDirect(address): open socket, SASL EXTERNAL then BEGIN, and SKIP the org.freedesktop.DBus bootstrap (no Hello, hence no daemon-assigned unique name; uniqueName stays null and callers keep their synthetic fallback). No RequestName/AddMatch.
- WireDbusBackend routes DIRECT_ADDRESS to connectDirect; SESSION_ADDRESS still routes to connect (full bootstrap); SESSION/SYSTEM/DEFAULT are untouched; DIRECT_FD/SERVER_FD still null. Direct mode is purely additive and gated on the bus type.
- Daemon-bound helpers are correct no-ops/synthetic in direct mode: requestName/releaseName report success (the single peer trivially owns the name), addMatch/removeMatch no-op, getNameOwner null. Crucially, incoming SIGNAL dispatch is independent of addMatch — the reader fans every SIGNAL out to all registered listeners, so local signal delivery/filtering is unaffected by the AddMatch no-op. Call/reply/signal correlation and incoming-serve are unchanged.

### 2. fd=-1 rejection — sound
writeMessage rejects a negative descriptor before SCM_RIGHTS with a clean DBusCallException mapped to org.freedesktop.DBus.Error.InvalidArgs ("Invalid Unix FD"), evicting the orphaned pending entry and leaving the connection intact — no unmapped ExecutionException, no corrupted connection.

### 3. The peer-test contract change — independently verified, accepted
The coder re-gated the native peer (NativeInteropPeerTest) from `verificationComplete && invalidRejected` to `verificationComplete` only, on the claim that the server-side "rejects invalid fd" branch is unsatisfiable by any client. Verified independently:
- fd=-1 is untransmittable on BOTH backends. JVM/wire: rejected pre-SCM_RIGHTS (above); even absent that, junixsocket/sendmsg refuses -1 with EBADF. Native: UnixFd.adopt(-1) keeps fd=-1, and sd-bus's 'h' append/dup of -1 fails — surfaced to the caller as "Invalid Unix FD". Neither path ever puts -1 on the wire, so the native peer's `if (fd.fd < 0)` server branch can never fire.
- The OLD behavior was a masked 20s timeout: main required invalidRejected (which no client can set), so waitUntil(..., 20.seconds) always timed out, and the failure was swallowed by --ktest_no_exit_code. Confirmed against main's source.
- Coverage is NOT weakened. The valid fd round-trip is still asserted (verificationComplete + Phase 5 live-dup paths). Client-side -1 rejection is now asserted in BOTH directions: the JVM client test (CrossRuntimeInteropSmokeTest, jvmClientRoundTripsUnixFd...) asserts a com.monkopedia.sdbus.Error mentioning "Invalid Unix FD", and the pre-existing native client test (NativeInteropPeerTest, nativeClientRoundTripsUnixFd...) already asserts the same client-side rejection. The change removes only an assertion that no conformant client could ever satisfy.

This is a genuine unsatisfiable-test fix, not a coverage regression — accepted without escalation.

### CI ground truth
- full-tests-x64 (JVM authority) GREEN (4m25s): ran `./gradlew --no-daemon allTests test crossRuntimeInteropTests`; python3-dbusmock 0.31.1 installed and the dbusmock suites compiled/engaged; :cross_test:jvmInteropTest and :crossRuntimeInteropTests both executed (not skipped/no-op) and passed — these are exactly the direct-bus suites that failed round 1.
- pr-smoke forward (jvm-client) all GREEN (the round-1 failures); reverse (native-client) GREEN; static-checks, native-build-x64, arm-runtime-tests GREEN. apiDump no-op (no api/ change).

Sound + green. Approving and merging — epic #93 finale.

PR #104, opened 2026-06-13
Fix generated-adaptor property getter to serialize the value (#89)

### LINKED ISSUE #89: Generated adaptor property getters never serialize the value — remote Get returns empty/hangs (PropertyVTableItem.with(KProperty0))

monkopedia-coder:
## The bug

`PropertyVTableItem.with(receiver: KProperty0<T>)` installed a getter that computed the property value and then threw it away:

    getter = { receiver.get() }   // value evaluated, never written

`PropertyGetCallback` is `(msg: PropertyGetReply) -> Unit` — the getter must serialize the value INTO the reply, exactly like `withGetter` does. Since the value was never appended, a remote `org.freedesktop.DBus.Properties.Get` / `GetAll` produced a malformed empty reply; on native sd-bus this drops the connection (`Error.Disconnected: Connection reset by peer`).

`AdaptorGenerator` emits `with(this@<Adaptor>::<prop>)` for every generated property, so this broke property reads for **every codegen-generated service adaptor on native** — the library's primary server-side use case. It hid because prior tests exercised us as a client against mock servers, own-server JVM tests route through `JvmStaticDispatch` (bypassing wire serialization), and nothing did a real remote Get against a native adaptor.

## The fix

Serialize into the reply, mirroring `withGetter`:

    getter = { reply -> reply.serialize<T>(receiver.get()) }

The setter path (`KMutableProperty0`) was already correct and is unchanged. No codegen change (it keeps emitting `with(::prop)`, now correct); golden fixtures unchanged; `apiDump` is a no-op (behavior-only; `api/` diff empty).

## Regression test

Added `propertiesProxy_supportsGetAndGetAllViaKotlinPropertyBinding` to `CommonApiIntegrationTest`. It registers a property via the generated-style `prop { with(::someProp) }` path (NOT `withGetter`), serves it, and has a client `Get`/`GetAll` it asserting the real value returns; it also round-trips a `Set` through the mutable binding.

Confirmed it **FAILS on main without the fix** (`linuxX64Test`: `Error.Disconnected: Connection reset by peer`) and **PASSES with it** — a genuine regression test, not a tautology.

## Verification

- `ktlintCheck`, `apiCheck` — pass; `apiDump` no-op (`git diff -- api/` empty)
- `compileKotlinLinuxX64` — pass
- full `:linuxX64Test` (native — the affected backend) — pass, including the new test
- `:jvmTest :cross_test:jvmTest` with dbusmock — pass (no JVM/wire regression)

Fixes #89.

### DIFF
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/PropertyVTableItem.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/PropertyVTableItem.kt
@@ -141,8 +141,10 @@ class PropertyVTableItem @PublishedApi internal constructor(
      */
     inline fun <reified T : Any> with(receiver: KProperty0<T>) {
         signature = Signature(signatureOf<T>().value)
-        getter = {
-            receiver.get()
+        getter = { reply ->
+            // Serialize the property value into the reply, mirroring withGetter; returning the
+            // value from the lambda is not enough — the callback type is (PropertyGetReply) -> Unit.
+            reply.serialize<T>(receiver.get())
         }
         if (receiver is KMutableProperty0) {
             setter = {
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/CommonApiIntegrationTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/CommonApiIntegrationTest.kt
@@ -44,6 +44,11 @@ class CommonApiIntegrationTest {
     @Serializable
     private data class Point(val x: Int, val y: Int)
 
+    // Backs the generated-adaptor `with(::prop)` binding exercised by the #89 regression test.
+    private class PropertyHolder {
+        var value: Int = 0
+    }
+
     private data class FixtureIds(
         val service: ServiceName,
         val path: ObjectPath,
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
## Approve

Surgical, behavior-only native correctness fix for #89.

### Fix
The getter installed by `PropertyVTableItem.with(KProperty0)` computed the value but discarded it (the callback type is `(PropertyGetReply) -> Unit`, so a returned value is ignored). It now serializes into the reply via `reply.serialize<T>(receiver.get())`, exactly mirroring `withGetter`. Since `AdaptorGenerator` emits `with(this@Adaptor::prop)` for every generated property, this was producing empty `Properties.Get`/`GetAll` replies and dropping the connection on native for every codegen adaptor. The `KMutableProperty0` setter path was already correct and is untouched.

### Regression test
`propertiesProxy_supportsGetAndGetAllViaKotlinPropertyBinding` binds via the generated-style `with(::value)` path (not `withGetter`/`withSetter`), performs a real remote `Get`/`GetAll` asserting the backing value, and round-trips a `Set` through the mutable binding. It exercises actual wire serialization, so it genuinely fails pre-fix and passes post-fix - not a tautology.

### Scope
Two files only: the one-block getter fix and the new test. No codegen change, no golden-fixture change, `api/` diff empty (apiDump no-op). CI green across native (full-tests-x64, native-build-x64, arm-runtime-tests), JVM, smoke, and static-checks.

PR #118, opened 2026-06-14
Add nightly GC-cleaner soak suite for native createCleaner sites

monkopedia-coder:
## What

Adds a nightly GC-soak suite (`CleanerSoakTest`) that validates **every** Kotlin/Native `createCleaner` site actually runs its finalizer when its owner becomes unreachable — the historically fragile, never-directly-tested half of native resource cleanup.

The deterministic `release()` path is already covered by the regular unit suites; this covers the timing-dependent half (GC actually invoking the cleaner).

## The six sites, two probe strengths

| Site | Holder pattern | Probe | Signal |
|---|---|---|---|
| `UnixFd` | `(fd, closer)` pair | fd actually closes (`fcntl`/`EBADF`) | cleaner **body ran** |
| `Reference` | `(value, closer)` pair | `onLeaveScopes` callback fires | cleaner **body ran** |
| `ConnectionImpl.EventFd` | `FdHolder` | owned eventfd closes | cleaner **body ran** |
| `PlainMessage` | `(msg, sdbus)` pair | `WeakReference` clears | **collectable** (no cycle) |
| `ObjectImpl` | `Allocs` | `WeakReference` clears | **collectable** (no cycle) |
| `ProxyImpl` | `Allocs` | `WeakReference` clears | **collectable** (no cycle) |

Where cleanup has a cheap observable side effect we assert the body actually ran (strongest signal). For the heavy sd-bus types there's no cheap observable, so we assert collectability via `WeakReference` — and a `this`-capture cycle is precisely what stops collection.

All probes are **bus-free** (they use the pseudo connection), so the suite needs no `dbus-run-session`.

## Why nightly-only

The probes force GC and poll for finalizers, so they're probabilistic. They're excluded from the default `linuxX64Test` run behind a `-PgcSoak` gate (`build.gradle.kts`) to keep the PR gate flake-free, and run only on the nightly schedule via a new `cleaner-soak` job in `stress-matrix.yaml` (schedule + manual nightly dispatch; never on PRs). A broken cleaner fails deterministically (nothing ever closes / clears); a healthy one settles in the first GC cycle.

## Verification

- All 6 probes pass locally on linuxX64 (settle on the first `GC.collect()`).
- Negative controls confirm the test has teeth:
  - **Body-ran style:** neutering `UnixFd`'s `closeOnce` makes it fail deterministically (0/64 closed, full poll budget burned).
  - **`this`-capture:** notably, Kotlin/Native 2.4 **rejects a `this`-capturing cleaner at compile time** (`createCleaner must take an unbound, non-capturing function or lambda`). So that specific regression cannot ship silently — this suite covers the residual runtime risks (a capture-free body that frees the wrong thing / no-ops, runtime reference cycles through the holder graph, or GC ceasing to run cleaners).
- `ktlintCheck` passes. No public API change (test + build-script + workflow only), so `apiCheck` is unaffected.

### DIFF
--- a/.github/workflows/stress-matrix.yaml
+++ b/.github/workflows/stress-matrix.yaml
@@ -223,3 +223,43 @@ jobs:
         run: |
           echo "Stress flake triage policy: docs/FLAKE_TRIAGE.md"
           echo "Re-run locally with the failing matrix case filter and direction."
+
+  # Native Kotlin/Native GC-cleaner soak. Probabilistic (forces GC, polls for finalizers), so it is
+  # excluded from the default linuxX64Test run (see the gcSoak gate in build.gradle.kts) and only runs
+  # here on the nightly schedule. All probes are bus-free (pseudo connection), so no dbus is required.
+  cleaner-soak:
+    if: github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.tier == 'nightly')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+          cache: gradle
+
+      - name: Install Native Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libcurl4-openssl-dev libsystemd-dev
+
+      - name: Run GC Cleaner Soak (linuxX64)
+        run: |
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE. Soak suite has no false-green path — every probe fails deterministically when its cleaner is broken (verified by negative controls), and all abandoned objects are genuinely unreachable. -PgcSoak gate keeps it off the PR gate; cleaner-soak job correctly nightly-only (skipping on this PR). All CI green. Nits are non-blocking follow-ups.

PR #119, opened 2026-06-16
Fix served-object and connection leaks on teardown (native + JVM)

monkopedia-coder:
## Summary

Fixes **real memory leaks in shipped 0.5.0**: every served object (and the connections serving them) leaked after `release()`, on both the native and JVM backends. Found by extending the GC-cleaner validation from "does the cleaner fire when an object is unreachable" to "does tearing a live object/connection down actually make it unreachable" — the live system was holding strong references that prevented collection, so the cleaner could never fire.

This is the lifecycle gap that `CleanerSoakTest` (PR #118) could not see: it abandoned resources in isolation; these tests exercise a real served object on a real bus through full teardown.

## The leaks

**Native — every served object leaked (severe, universal).** `VTable.clear()` (run on `release()`) freed the Arena and nulled the back-reference, but never dropped the `methods`/`signals`/`properties` lists holding the user-supplied callbacks. The `VTable` outlives `release()` (it is pinned by its `Reference`'s own GC cleaner until collected), so those callbacks — which capture the adaptor via `acall(this::m)` / `withGetter { … }` — kept the served object alive forever. Every generated/real adaptor hits this.

**Native — connections leaked when objects were registered (a class of bug).** Several cleanup closures captured `this@ConnectionImpl` via member access (`sdbus.sd_bus_slot_unref(…)`). These closures are retained by the returned `Reference`'s GC cleaner even after they run, so a `this` capture pins the whole connection. The compiler's non-capturing `createCleaner` check does **not** catch this, because the capture is laundered through `Reference`'s `onLeaveScopes` parameter (the `createCleaner` lambda itself only captures `resource`). An audit of every cleanup closure in the native backend found 5 sites; `addMatch` was already written correctly with a local. Fixed by hoisting `val localSdbus = sdbus`:
- `ConnectionImpl.addObjectVTable`, `addObjectManager`, `addMatchAsync`, `registerSignalHandler`
- `MethodCall.send` (async-call slot; pinned the call message)

**JVM — property-bearing served objects leaked (the common BlueZ case).** `WireDbusObject` registered the `Properties` Get/Set/GetAll handlers into the process-wide `JvmStaticDispatch` singleton and never removed them on teardown; each captures the `WireDbusObject` (and through it the adaptor's getters/setters and the wire connection). `obj.release()` cleaned up methods but not these, so any object declaring ≥1 property leaked for the process lifetime. Fixed by unregistering them and clearing the property/interface maps in `release()` (and when the last property interface is removed).

## Regression tests (nightly, `-PgcSoak`)

- `CleanerLifecycleSoakTest` (native) — builds a real served object (method + property + signal + object-manager) on a real session bus, tears it down, forces GC, and asserts both the object and the connection are collected.
- `CleanerJvmLifecycleSoakTest` (JVM) — same shape on the owned junixsocket backend; asserts a property-bearing object is collected after `release()`.

Both are GC-timing-dependent, so they are excluded from the default test run (the `-PgcSoak` gate now matches `Cleaner*SoakTest`) and run only on the nightly schedule. The `cleaner-soak` CI job now runs native + JVM under `dbus-run-session` (the lifecycle suites need a real bus). Each fix was confirmed with a negative control (revert the fix → the matching test fails deterministically; restore → passes).

## Verification

- All three soak suites green under `dbus-run-session`: native lifecycle (2), bus-free `CleanerSoakTest` (6), JVM lifecycle (1).
- Full existing `:linuxX64Test` + `:jvmTest` suites pass under `dbus-run-session` — no regressions from the serving-path changes.
- `ktlintCheck` + `apiCheck` pass. No public API change (internal + test + CI only).

### DIFF
--- a/.github/workflows/stress-matrix.yaml
+++ b/.github/workflows/stress-matrix.yaml
@@ -224,9 +224,10 @@ jobs:
           echo "Stress flake triage policy: docs/FLAKE_TRIAGE.md"
           echo "Re-run locally with the failing matrix case filter and direction."
 
-  # Native Kotlin/Native GC-cleaner soak. Probabilistic (forces GC, polls for finalizers), so it is
-  # excluded from the default linuxX64Test run (see the gcSoak gate in build.gradle.kts) and only runs
-  # here on the nightly schedule. All probes are bus-free (pseudo connection), so no dbus is required.
+  # GC-cleaner soak (native + JVM). Probabilistic (forces GC, polls for finalizers), so it is excluded
+  # from the default test run (see the gcSoak gate in build.gradle.kts) and only runs here on the
+  # nightly schedule. CleanerSoakTest is bus-free, but the lifecycle suites (native + JVM) exercise a
+  # real served object/connection, so the whole run is wrapped in dbus-run-session.
   cleaner-soak:
     if: github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && inputs.tier == 'nightly')
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE. Two genuine teardown leaks fixed correctly on both backends — native served-object (VTable.clear callback lists) + connection capture-closures (5 sites hoisted to locals), and JVM Properties-dispatch singleton unregister. No use-after-release: VTable.clear runs after slot unref/dispose; JVM unregister runs before map clear. Each fix negative-controlled. Coverage nit addressed by added proxy lifecycle tests (connection-after-subscription guards registerSignalHandler). Full native+JVM suites green under dbus-run-session; ktlint+apiCheck pass; no public API change. Lifecycle soak tests correctly nightly-gated.

PR #120, opened 2026-06-17
Fix cleaner-soak nightly job running the full suite instead of soak tests

monkopedia-coder:
## What

Fixes the nightly `cleaner-soak` job (added in #118/#119) which turned `main`'s scheduled run red.

## Root cause

The job ran `:linuxX64Test :jvmTest --tests "*Cleaner*SoakTest*"`. **Gradle applies a command-line `--tests` filter to only one of multiple test tasks and runs the others unfiltered** — so `linuxX64Test` ran the *entire* integration suite (280 tests) instead of the ~9 soak tests, while `jvmTest` was filtered correctly (1 test). The full integration suite fails under this job's environment, so the run went red. It looked green locally because the integration tests pass locally regardless of the filter.

## Fix

Drive the soak-test selection from the `-PgcSoak` gate in `build.gradle.kts` rather than a fragile cross-task `--tests` filter:
- default (no `-PgcSoak`): exclude `Cleaner*SoakTest` from every test task — **PR-gate behaviour unchanged**.
- with `-PgcSoak`: include **only** `Cleaner*SoakTest` (and `isFailOnNoMatchingTests = false`).

The CI job now runs `:linuxX64Test :jvmTest -PgcSoak` with no `--tests`, which reliably runs only the soak suites on both tasks.

## Verification (local)

- `-PgcSoak`: runs exactly the 3 Cleaner classes — `CleanerSoakTest` + `CleanerLifecycleSoakTest` (native) and `CleanerJvmLifecycleSoakTest` (JVM), green.
- Default run: full suite (27 JVM classes), `Cleaner*` absent, green.
- `ktlintCheck` passes.

### DIFF
--- a/.github/workflows/stress-matrix.yaml
+++ b/.github/workflows/stress-matrix.yaml
@@ -249,10 +249,11 @@ jobs:
           sudo apt-get install -y libcurl4-openssl-dev libsystemd-dev dbus
 
       - name: Run GC Cleaner Soak (native + JVM)
+        # `-PgcSoak` makes the build run ONLY the Cleaner*SoakTest suites (see build.gradle.kts); no
+        # `--tests` filter — that is unreliable across two test tasks and ran the full suite in CI.
         run: |
           set -euo pipefail
-          dbus-run-session -- ./gradlew :linuxX64Test :jvmTest \
-            --tests "*Cleaner*SoakTest*" -PgcSoak \
+          dbus-run-session -- ./gradlew :linuxX64Test :jvmTest -PgcSoak \
             --rerun-tasks --stacktrace | tee cleaner-soak.log
 
       - name: Upload Failure Artifacts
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Tier-1 infra fix, reviewer-green. Confirmed: root cause is Gradle applying a command-line --tests filter to only one of multiple test tasks (so :linuxX64Test ran unfiltered = full 280-test suite). Fix moves soak selection into the -PgcSoak gate (default excludes Cleaner*SoakTest; -PgcSoak includes-only), job drops --tests. Verified locally: -PgcSoak runs exactly the 3 Cleaner classes on both linuxX64Test+jvmTest; default run excludes them and is green. PR-gate CI all green. Merging; will trigger a manual nightly dispatch to confirm cleaner-soak goes green on CI.

Issue #108, opened 2026-06-14
Epic: 1.0-polish wave (0.6.0) — post-0.5.0 API review decisions


monkopedia-coder:
Epic tracking the **1.0-polish wave** — the decisions from the post-0.5.0 API-surface review (5 parallel review agents, owner-decided 2026-06-14). Lands as **0.6.0** (the deprecation release); **1.0** = 0.6.0 minus the deprecated aliases + the final freeze.

## Gate: 0.6.0 ships when these resolve cleanly
**Breaking (with deprecation aliases where possible):**
- Rename `Error` → `SdbusException` (deprecated `typealias Error` in 0.6.0, removed at 1.0)
- Deprecate the fluent property layer toward removal (add direct typed `suspend` accessors first)
- Move fd factories `createDirectBusConnection(fd)`/`createServerBusConnection(fd)` out of common → native-only
- `requestName(name, flags): RequestNameReply` (surface outcome + flags; source-compatible)
- Naming batch: `SdbusSig`→`TypeSignature`, `acall`→`asyncCall`, `Message.path`→`objectPath`, `SignalEmitter`/`SignalSubscriber` var renames, top-level `createPlainMessage`, `Flags` `in` operator, `addMatchAsync` internalize/rename, `MethodReply` ctor cleanup

**Behavioral / codegen (mostly additive):**
- `createObject(..., runEventLoopThread=true)` symmetric with `createProxy`; make `startEventLoop()` idempotent
- Generated adaptors auto-emit `PropertiesChanged` on remote Set (honor `EmitsChangedSignal`) + a public notifying-property delegate for server-side changes

**Mechanical:** `Resource : AutoCloseable`, de-`data` `Typed`, `Flags.Count`→internal, drop `PropertyDelegate.name`, `maybeDegrouped`→internal, C++-doxygen KDoc cleanup

**Left as-is (decided):** `Error.name`/`errorMessage` member names, `dontExpectReply`.
**Additive post-1.0 (NOT this wave):** Introspectable client helper, bus-daemon helpers, per-name credentials, #90 JVM cross-process serving (+ a call-site KDoc caveat), #101 worker-pool bound, and the broader sync-blocking-surface question.

[...]

PR #121, opened 2026-06-17
Bound the JVM serve worker pool (deadlock-free, compensated)

### LINKED ISSUE #101: Owned-connection serve worker pool is unbounded (newCachedThreadPool)

monkopedia-coder:
Closes #101. Group 0 of the 0.6.0 wave (epic #108). JVM-internal only — no public API change.

## Problem
`DBusWireConnection` served incoming method calls on `newCachedThreadPool()` → one thread per concurrent call, unbounded growth under a flood of slow/blocking handlers.

## Design — bounded + deadlock-free
New `ServeWorkerPool`: a `ThreadPoolExecutor` pinned at a bounded `corePoolSize` (~2×CPU, clamped 4..64) with an unbounded `LinkedBlockingQueue` (so the sole submitter — the reader loop — never blocks/rejects/runs-inline; it must stay free to read the replies that unpark workers).

**The nested-call hazard:** a served handler X (on a worker) makes a blocking same-connection call to Y, also served here; X parks waiting for Y, Y needs a worker. A naive fixed bound saturated with slow handlers can never give Y a thread → deadlock.

**Fix — ForkJoinPool-style compensation:** every blocking wait in `call`/`callBlocking` is bracketed by `beginNestedBlock()`/`endNestedBlock()`. A per-pool ThreadLocal marks worker threads; when a *worker* is about to park on a nested call, `corePoolSize` is raised by one and a replacement worker pre-started (lowered on return). Invariant: **live workers == bound + (workers parked on nested calls)** → always ≥ bound runnable workers at any nesting depth. A flood of *independent* slow handlers never compensates, so it stays pinned at the bound and excess calls queue. Surplus compensating workers retire via `allowCoreThreadTimeOut`.

## Tests (jvmTest, under dbus-run-session)
- `nestedSameConnectionDispatch_underBoundedPool_completesWithoutDeadlock` — fires `bound+8` concurrent Outer→Inner chains; FIFO ordering means all base workers park on Outers before any Inner is enqueued → deterministically deadlocks a naive bound. `withTimeout(20s)` fails fast. Passes in ~9ms.
- `serveWorkerPool_underFloodOfSlowHandlers_staysBounded` — floods `bound×4` independent slow handlers; asserts completion and `largestPoolSize <= bound`.

## Verification
- `:jvmTest` under dbus-run-session: 300 tests, 0 failures, no regressions.
- `ktlintCheck` + `apiCheck` pass (new members are `internal`; no api dump change).

### DIFF
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
@@ -36,7 +36,9 @@ import java.nio.charset.StandardCharsets
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArrayList
-import java.util.concurrent.Executors
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE. Deadlock-freedom invariant verified — sole non-blocking submitter (reader loop, unbounded queue), corePoolSize raised before a worker parks on a nested call so >=bound runnable workers at any nesting depth; compensation is exception-safe (finally) and race-free (poolLock). Both tests have real teeth: nested test deterministically deadlocks a naive bound (FIFO saturation, bound+8 margin); flood test proves largestPoolSize<=bound. Review nit addressed (serve-handler constraint now documented). CI all green, no api change (internal only).

Issue #112, opened 2026-06-14
requestName: surface RequestNameReply outcome + flags


monkopedia-coder:
Part of epic #108.

Part of the 0.6.0 1.0-polish wave.
Change `Connection.requestName(name: ServiceName): Unit` → `requestName(name: ServiceName, flags: ... = <none>): RequestNameReply`. Add a `RequestNameReply` enum (PrimaryOwner / InQueue / Exists / AlreadyOwner) and a flags type (ReplaceExisting / AllowReplacement / DoNotQueue). Native: stop hardcoding `sd_bus_request_name(..., 0u)` — pass the flags and map the result code to the enum (ConnectionImpl.kt:247-252 currently discards it). JVM: the internal `DBusWireConnection.requestName(busName, flags): Int` already returns the code — surface it. Source-compatible: existing `requestName(name)` calls still compile (return value ignorable). This is the one signature whose Unit return forecloses outcome reporting non-additively, hence pre-1.0.

monkopedia-coder:
**Owner decision (2026-06-17) — flags type shape (a frozen 1.0 surface):**

```
enum class RequestNameFlag { ALLOW_REPLACEMENT, REPLACE_EXISTING, DO_NOT_QUEUE }
enum class RequestNameReply { PRIMARY_OWNER, IN_QUEUE, EXISTS, ALREADY_OWNER }

fun requestName(name: ServiceName, vararg flags: RequestNameFlag): RequestNameReply
```

Trailing `vararg` with empty default keeps `requestName(name)` source-compatible (return type changes Unit -> RequestNameReply, source-compatible since callers may ignore it). Wire the flag bits (ALLOW_REPLACEMENT=0x1, REPLACE_EXISTING=0x2, DO_NOT_QUEUE=0x4) and map the reply code (1..4) to the enum on both backends. Group 5 (with #111).

monkopedia-coder:
Done in PR #126 (merged 7d247c4) alongside #111 — requestName now returns RequestNameReply + takes vararg RequestNameFlag, with native/JVM reply-code mappings made consistent (the wire-vs-sd-bus queue-bit translation fix). Auto-close missed it because the PR body said 'Closes #111 and #112'.

PR #126, opened 2026-06-17
Connection: fd factories native-only [#111] + requestName reply/flags [#112]

### LINKED ISSUE #111: Move fd-based connection factories out of common to native-only
Part of epic #108.

Part of the 0.6.0 1.0-polish wave.
Move the fd-based connection factories out of the common `expect` surface to native-only (they are `@Deprecated(level=ERROR)` compile-errors on JVM today — the common API shouldn't declare what a backend can't honor). Concretely: remove `expect fun createDirectBusConnection(fd: UnixFd)` and `expect fun createServerBusConnection(fd: UnixFd)` from commonMain `Connection.kt` and their JVM ERROR actuals; declare them as plain native-only `fun`s in `nativeMain/Connection.native.kt`, exactly like `createRemoteSystemBusConnection` already is (the model the completeness review endorsed). Address-based `createDirectBusConnection(String)` stays common. Regenerate apiDump (JVM .api loses the two symbols; klib keeps them).

monkopedia-coder:
Closes #111 and #112. Group 5 of the 0.6.0 wave (epic #108). Both touch only the Connection.kt trio.

## #111 — fd-based factories → native-only
Removed `expect fun createDirectBusConnection(fd: UnixFd)` / `createServerBusConnection(fd: UnixFd)` from commonMain + their JVM `@Deprecated(ERROR)` actuals; redeclared them as plain native-only `fun`s in `Connection.native.kt` (modeled on `createRemoteSystemBusConnection`). Address-based `createDirectBusConnection(String)` stays common. **api delta:** JVM .api drops both fd symbols; klib .api keeps them. `JvmFactoryMockkTest` updated.

## #112 — requestName -> RequestNameReply + flags
```
enum class RequestNameFlag(val mask: UInt) { ALLOW_REPLACEMENT(0x1), REPLACE_EXISTING(0x2), DO_NOT_QUEUE(0x4) }
enum class RequestNameReply(val code: Int) { PRIMARY_OWNER(1), IN_QUEUE(2), EXISTS(3), ALREADY_OWNER(4) }  // + fromCode
fun requestName(name: ServiceName, vararg flags: RequestNameFlag): RequestNameReply
```
- Native: ORs flag masks into `sd_bus_request_name`; maps the return (1→PRIMARY_OWNER, 0→IN_QUEUE, -EEXIST→EXISTS, -EALREADY→ALREADY_OWNER; other negatives still throw). All 4 outcomes faithfully surfaced.
- JVM: folds masks into the existing `DBusWireConnection.requestName(busName, flags): Int` and maps the reply code via `fromCode`.

## Behavior note
The only behavioral delta for the no-flags factory path (`createBusConnection(name)` etc.): previously requesting a name you ALREADY own threw (`-EALREADY`); now it returns `ALREADY_OWNER`. Taken-by-another behavior is unchanged (old flags were `0u` = queue → `IN_QUEUE`, silently queued then as now; the factories ignore the return as before). `requestName(name)` stays source-compatible (empty vararg).

## Verification
- compile jvm + linuxX64 + linuxArm64; `:jvmTest` 299/0, `:linuxX64Test` 268/0 + cross_test (dbus-run-session); `ktlintCheck` + `apiCheck`: all green.

### DIFF
--- a/api/sdbus-kotlin.api
+++ b/api/sdbus-kotlin.api
@@ -54,7 +54,7 @@ public abstract interface class com/monkopedia/sdbus/Connection : com/monkopedia
 	public abstract fun getMethodCallTimeout-UwyO8pc ()J
 	public abstract fun getUniqueName-AwvMx1g ()Ljava/lang/String;
 	public abstract fun releaseName-em-jQOc (Ljava/lang/String;)V
-	public abstract fun requestName-em-jQOc (Ljava/lang/String;)V
+	public abstract fun requestName-joT7nYs (Ljava/lang/String;[Lcom/monkopedia/sdbus/RequestNameFlag;)Lcom/monkopedia/sdbus/RequestNameReply;
 	public abstract fun setMethodCallTimeout-LRDsOJo (J)V
 	public abstract fun startEventLoop ()V
 	public abstract fun stopEventLoop (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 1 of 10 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE (after fix). #111 fd-factory move correct (JVM .api drops 2, klib keeps; no non-native caller broke). #112: the initial review caught a real cross-backend bug — native fed D-Bus wire flag bits straight into sd_bus_request_name whose 0x4 is the INVERSE (SD_BUS_NAME_QUEUE). Fixed by translating wire->sd-bus flags (invert queue bit); re-review confirmed an 8-combination table where native and JVM now return identical RequestNameReply for every flag combo. Reply-code mapping correct on both backends. Behavior change is intended + consistent (createBusConnection(name) now queues a taken name like JVM/standard D-Bus instead of native throwing); no test regresses. CI green. Auto-merge per wave authorization.

Issue #115, opened 2026-06-14
Generated adaptors auto-emit PropertiesChanged (+ notifying delegate)


monkopedia-coder:
Part of epic #108.

Part of the 0.6.0 1.0-polish wave. (Builds on #89 which fixed the getter serialization.)
Make generated adaptors emit `PropertiesChanged` automatically (they currently never do — the codegen `prop { with(this@Adaptor::prop) }` setter sets the backing field but doesn't emit). Two parts:
1. CODEGEN/remote-Set: the generated adaptor property setter emits `emitPropertiesChangedSignal(interface, [name])` after a remote `Properties.Set`, HONORING the `org.freedesktop.DBus.Property.EmitsChangedSignal` annotation (default = emit; skip if the XML says `false`/`const`). Codegen currently ignores that annotation for emission (AdaptorGenerator.kt:156-172).
2. SERVER-SIDE change: add a PUBLIC notifying-property delegate that emits `PropertiesChanged` on every set, which codegen uses as the backing field for writable EmitsChangedSignal properties — so a service setting the property itself also auto-emits. The delegate captures the Object + interface + property name.
Result: "set a property, clients' .changes() flows see it" is the default on both paths. The demo-service workaround (hand-rolled withGetter/withSetter + manual emit) can be removed. Add a real-peer (dbusmock) test that a client's PropertiesChanged subscription fires on both a remote Set and a server-side set.

monkopedia-coder:
**Owner decision (2026-06-17) — public notifying-property delegate shape:**

```
fun <T> Object.notifying(
    interfaceName: InterfaceName,
    propertyName: PropertyName,
    initialValue: T,
): ReadWriteProperty<Any?, T>
```

Extension on `Object`; `setValue` stores then emits PropertiesChanged for that single property, SKIPPING the emit when the new value equals the current (matches the existing manual `updatePrefix` behavior in the demo sample). Wire codegen to generate properties via this delegate, and replace the demo-service workaround. Group 6 (with #114).

monkopedia-coder:
Done in PR #127 (merged 1dc86fc) alongside #114 — generated adaptors auto-emit PropertiesChanged honoring EmitsChangedSignal, plus the public Object.notifying delegate. (Required the native event-loop fix #128 as a prereq.) Auto-close missed it because the PR body said 'Closes #114 and #115'.

PR #127, opened 2026-06-17
Event-loop symmetry [#114] + adaptor PropertiesChanged auto-emit + notifying delegate [#115]

### LINKED ISSUE #114: Event-loop symmetry: createObject runEventLoopThread + idempotent startEventLoop
Part of epic #108.

Part of the 0.6.0 1.0-polish wave.
Fix the event-loop lifecycle asymmetry: client `createProxy` auto-starts the loop (runEventLoopThread=true default) but server `createObject` doesn't. Add `runEventLoopThread: Boolean = true` to `createObject` (Object.kt:222 + actuals), mirroring `createProxy`, so both sides auto-start by default with the same `=false` opt-out. CRITICAL: make `connection.startEventLoop()` IDEMPOTENT (a no-op when the loop already runs) so existing server code doing `createObject()` + explicit `startEventLoop()` doesn't spawn a second loop thread — verify both native (ConnectionImpl) and JVM. Simplify the demo-service sample to drop the now-redundant explicit startEventLoop. Add a one-line KDoc note: register your vtable (addVTable) before advertising the object.

monkopedia-coder:
Closes #114 and #115. Group 6 of the 0.6.0 wave (epic #108). Share Object.kt + the demo sample.

## #114 — event-loop symmetry
- `createObject(connection, objectPath, runEventLoopThread: Boolean = true)` (common expect + native/jvm actuals + ServiceConnection convenience), mirroring `createProxy`. When true, calls `startEventLoop()`.
- `startEventLoop()` idempotency confirmed both backends: native guards on `asyncLoopThread/eventLoopStarting/released` under lock; JVM `WireDbusConnection.startEventLoop` is a no-op (reader/serve threads start in the connection ctor). New `EventLoopStartIdempotencyTest` (both backends).
- Demo simplified (dropped redundant explicit startEventLoop). KDoc note: register the vtable before serving.

## #115 — generated adaptors auto-emit PropertiesChanged
- Public `fun <T> Object.notifying(interfaceName, propertyName, initialValue): ReadWriteProperty<Any?, T>` — setValue stores then emits PropertiesChanged, skipping when value unchanged.
- Codegen: writable properties whose `EmitsChangedSignal` is not false/const are generated `override var X by obj.notifying(...)`; remote Set routes through the delegate setter (single emit), server-side set emits directly (single emit, no double). const/false stay abstract, no emit. (XmlFormat already parses the annotation.)
- Demo drops the hand-rolled withGetter/withSetter/manual-emit; `prefix` uses the delegate.
- New dbusmock real-peer test: client PropertiesChanged subscription fires on both a remote Set and a server-side set (verified non-fake with DBUSMOCK_PYTHON).

## KNOWN ISSUE / reviewer focus
`createObject` auto-start exposes a latent NATIVE loop-startup synchronization issue: at high churn (hundreds of rapid create+register cycles on shared connections) loops can wedge + leak bus fds (via ConnectionImpl.release()'s 2s fail-safe). Worked around by having the legacy high-volume native tests opt out with `runEventLoopThread=false` (+ register vtable before starting — the documented pattern). Single/few-object usage is fine. The underlying native loop-startup fix is deferred (follow-up). **This is the open question for merge — see review.**

## Verification
- compile x3, `:codegen:test`, `:jvmTest`/`:linuxX64Test`/`:cross_test` (dbus-run-session + DBUSMOCK_PYTHON), `ktlintCheck`, `apiCheck`: all green. api adds the `notifying` delegate + createObject overloads.


monkopedia-reviewer (approved):
Independent review: APPROVE (after the prereq fix + finalization). Both prior blockers resolved: (1) createObject auto-start is now safe — #128's per-connection event-loop dispatcher is present in the branch (no shared pool), and 49/50 runEventLoopThread=false opt-outs were removed so the full 690-test suite exercises the true default with no wedge (only the leak-canary opt-out kept, legitimately); (2) notifying delegate hardened — current is @Volatile and the check-and-set runs under synchronized(SynchronizedObject), and emit-under-lock is NOT a deadlock risk (emit is non-blocking fire-and-forget on both backends, leaf lock). Codegen exactly-once PropertiesChanged honoring EmitsChangedSignal, clean fixtures, genuine dbusmock test (skipped=0). CI all green. Auto-merge per wave authorization.

PR #128, opened 2026-06-17
Fix native event-loop thread starvation (dedicated thread per connection) [#114 prereq]

monkopedia-coder:
Prerequisite for #114 (epic #108). Fixes a native event-loop thread-starvation bug that #114's createObject auto-start would multiply.

## Root cause
Every connection's event loop runs a BLOCKING `poll()` for its whole lifetime, and all loops were launched onto one process-wide `newFixedThreadPoolContext(8, "EventThreads")`. So at most 8 connection loops can run concurrently; the 9th+ coroutine is queued and never reaches `poll()` — a served object on such a connection never answers calls.

Deterministic repro (new `EventLoopConcurrencyStressTest`): 16 independent served objects, each on its own connection+loop; calling each one. On main it FAILS (~5.2s, `org.freedesktop.DBus.Error.Timeout` — object #8's loop never runs). The 8-thread boundary is the exact threshold. (The originally-hypothesized "2s on every release" symptom is release-order-dependent — forward-order release cascades and dodges it — so the test targets the deterministic never-serves behavior instead.)

## Fix — dedicated thread per connection loop
Each `ConnectionImpl` gets its own `newSingleThreadContext("sdbus-eventloop-<id>")`, created lazily on first `startEventLoop` (under the state lock), reused across stop/start. `release()` closes it exactly once, AFTER the loop has stopped (read-and-null under the lock = double-close guard); the 2s fail-safe early-return deliberately leaves it open (its thread may still be parked in poll()); never-started connections create no dispatcher. Removed the now-unused shared `eventPool` (verified single launch site). `@OptIn(DelicateCoroutinesApi)`.

## Verification
- New test: before = FAIL (timeout 5216ms); after = PASS (1607ms, 16 concurrent loops, max release 100ms).
- Full native `:linuxX64Test :cross_test:linuxX64Test`: green + fast (~35s).
- compile linuxX64 + linuxArm64 + jvm; `ktlintCheck` + `apiCheck`: pass. Internal-only — no api dump change.

### DIFF
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt
@@ -22,7 +22,8 @@
  */
 @file:OptIn(
     ExperimentalForeignApi::class,
-    ExperimentalNativeApi::class
+    ExperimentalNativeApi::class,
+    DelicateCoroutinesApi::class
 )
 
 package com.monkopedia.sdbus.internal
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE. Dispatcher lifecycle correct on every path: lazy create exactly-once under lock; close exactly-once via released-CAS + read-and-null-under-lock; close only ever hits an idle worker (normal path closes after join; 2s fail-safe deliberately leaves it open); never-started=no-op; close outside lock = no deadlock. Removed shared eventPool (single launch site confirmed). Regression test has teeth (16>8 served objects, fails on main with Timeout, robust margin, fails-fast, cleans up). The ARM failure was test placement (bus test must live in com.monkopedia.sdbus.integration to run under dbus-run-session on ARM) — fixed by moving it. All CI green. Internal-only, no api change. Prereq for #114/#127. Auto-merge per wave authorization.

Issue #110, opened 2026-06-14
Deprecate fluent property layer; add direct typed suspend accessors


monkopedia-coder:
Part of epic #108.

Part of the 0.6.0 1.0-polish wave.
Deprecate the sdbus-c++ fluent property layer toward removal at 1.0. First ADD direct typed accessors so async one-shot property get/set is first-class (not builder-only): `suspend inline fun <reified T> Proxy.getPropertyAsync(iface, member): T`, `setPropertyAsync(...)`, and direct `getAllProperties`/async forms — mirroring the existing sync `getProperty`/`setProperty` (PropertyGetter.kt). THEN `@Deprecated(ReplaceWith(...), WARNING)` the fluent builders + entry points in `ConvenienceApiClasses.kt` (`AsyncPropertyGetter`/`AsyncPropertySetter`/`AllPropertiesGetter`/`AsyncAllPropertiesGetter` + `getPropertyAsync`/`setPropertyAsync`/`getAllProperties` that return builders, and `onInterface`/`toValue`/`getResult`). Removed at 1.0. Note the `AllPropertiesGetter.onInterface()` is a SYNC blocking call — its replacement should be the typed direct form. Nothing on the blessed path (codegen/delegates) uses the fluent layer.

monkopedia-coder:
**Verification (2026-06-17) — safe to proceed:** confirmed nothing on the codegen/delegate "blessed path" uses the fluent property layer. `codegen/src/main/` has zero references to `onInterface`/`getProperty(`/`getResult`/`toValue`, and no generated fixture uses it; the delegate path calls `PropertiesProxy.Get/Set` directly. So deprecating the fluent layer (`@Deprecated(WARNING, ReplaceWith)`) will not emit warnings in generated code. Proceed as filed. Group 4.

PR #132, opened 2026-06-20
Test coverage for the 0.6.0 surface (requestName reply, notifying delegate, direct accessors)

monkopedia-coder:
Non-gated post-0.6.0 test coverage (spare-quota). Tests only — no production code, apiCheck unchanged. No bugs found.

- **#112 requestName reply codes** (`CommonApiIntegrationTest`, commonTest → runs on BOTH backends): PRIMARY_OWNER (fresh) / ALREADY_OWNER (same conn re-requests) / IN_QUEUE (2nd conn, no flags) / EXISTS (2nd conn, DO_NOT_QUEUE). Living in commonTest means native and JVM assert the SAME enum per scenario — a cross-backend-consistency regression guard for the 0.6.0 flag-handling fix.
- **#110 direct property accessors**: `getPropertyAsync`/`setPropertyAsync`/`getAllProperties`/`getAllPropertiesAsync` round-trip against a served object; agree with sync getProperty/setProperty + correct Variant unwrap.
- **#115 notifying delegate** (cross_test dbusmock): emits PropertiesChanged exactly once on change, nothing on same-value set (skip-on-unchanged), reads back the stored value.
- **#115 plain property no-emit**: the EmitsChangedSignal=false/const path — a plain (non-notifying) writable property must NOT emit on a remote Set even though the value changes.

dbusmock tests verified genuinely run (negative control: `DBUSMOCK_PYTHON=/usr/bin/false` → 3 SKIP lines; real venv → 0 SKIP, passing). Full suite green under dbus-run-session.

### DIFF
--- a/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockAdaptorPropertiesChangedTest.kt
+++ b/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockAdaptorPropertiesChangedTest.kt
@@ -31,14 +31,17 @@ import com.monkopedia.sdbus.addVTable
 import com.monkopedia.sdbus.createBusConnection
 import com.monkopedia.sdbus.createObject
 import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.getProperty
 import com.monkopedia.sdbus.notifying
 import com.monkopedia.sdbus.prop
 import com.monkopedia.sdbus.setProperty
 import kotlin.random.Random
 import kotlin.test.Test
 import kotlin.test.assertEquals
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE. Tests have real teeth (verified non-tautological): requestName reply-codes anchored to the real enum across both backends (commonTest = cross-backend consistency guard); direct accessors anchored to sync truth + Variant unwrap; notifying delegate counts emissions (double/spurious emit would fail); plain-property test confirms value DID change but no emit. dbusmock genuinely runs (negative control: /usr/bin/false -> SKIP). Tests-only, apiCheck unchanged. CI green. Nits non-blocking.

PR #133, opened 2026-06-20
Harden serve worker pool: saturation watchdog for non-compensated nested blocking

monkopedia-coder:
Non-gated post-0.6.0 robustness (spare-quota). Closes a review-flagged latent gap in the bounded JVM serve worker pool (#101). Internal only — apiCheck unchanged.

## Gap (confirmed with a repro)
The #101 same-thread nested-block COMPENSATION only fires when a worker blocks inside `call`/`callBlocking`. A handler that (a) offloads its nested same-connection call to another dispatcher while the worker parks, or (b) blocks the worker on a non-same-connection dependency, gets NO compensation → the bounded pool can starve (all workers blocked, queue non-empty, zero forward progress). Reproduced on current code (case b: workers park on a shared latch only a queued call can open) — fails fast under timeout.

## Fix — saturation watchdog
A single daemon thread samples forward progress every 250ms and spawns ONE compensating worker only when ALL hold: queue non-empty AND every worker busy AND no task has started/completed for a 1s stall window. Capped at `maxExtra = bound`; releases extras when the queue drains. `corePoolSize` invariant is now `bound + parked + watchdogExtra` (recomputed under the lock; composes with the existing fast-path nested compensation).

**No false-fire:** the discriminator is *forward progress* (start/finish timestamps), so a legitimate flood — workers keep cycling — never trips it (the existing flood test still passes at exactly `bound`). **Bounded:** even a pathological all-slow flood is clamped at `maxExtra`, never thread-per-call. **Deadlock-free:** the watchdog only adds runnable capacity / removes idle; execute never blocks/rejects.

Regression test `externalLatchDependency_underBoundedPool_isRescuedByWatchdog` (fails fast pre-watchdog, completes with it). Existing pool tests still pass (stable across 3 reruns).

## Caveat (flagged, not baked in)
A very WIDE case-(a) offloaded-nested storm can need ~N workers to fully drain (the irreducible "N concurrently-parked handlers need ~N threads"); the watchdog caps at `bound` extra, so wide offloaded storms are mitigated but not guaranteed fully drained. Guaranteeing that would be a deliberate "unbounded-while-stalled" policy choice. The guidance that serve handlers should do nested same-connection calls synchronously still stands.

Verified: compileKotlinJvm, full :jvmTest green (3 reruns), ktlintCheck, apiCheck (unchanged).

### DIFF
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
@@ -883,7 +883,10 @@ private class ServeWorkerPool {
             return
         }
         val stale = System.nanoTime() - lastProgressNanos.get() >= STALL_THRESHOLD_NANOS
-        val allBusy = executor.activeCount >= executor.poolSize
+        // Only intervene once the base pool is fully spun up (poolSize >= bound). Below that the
+        // executor can still grow on its own, and an idle connection (poolSize == 0) would otherwise
+        // read as "all busy" (0 >= 0) and spuriously add a worker on the first burst after idling.
+        val allBusy = executor.poolSize >= bound && executor.activeCount >= executor.poolSize
         if (stale && allBusy) {
             synchronized(poolLock) {
                 if (watchdogExtra < maxExtra) {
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review: APPROVE. Saturation watchdog is deadlock-free (daemon thread only samples + brief non-blocking lock mutations; can't block or be blocked by the pool; execute never blocks/rejects), bounded (watchdogExtra capped at maxExtra=bound, released on drain, surplus retires via allowCoreThreadTimeOut), and clean on shutdown (running=false + interrupt, daemon). Forward-progress timestamp refreshed on task start AND completion, so a legitimate flood never trips it (flood-stays-bounded test holds). corePoolSize invariant bound+parked+watchdogExtra recomputed under poolLock, composes with the nested-block fast path. Repro test has teeth (non-compensated external-latch path, deadlocks pre-watchdog, rescued after). Startup-transient false-fire nit FIXED (gate on poolSize>=bound). Caveat (very wide offloaded-nested storms not fully drained past cap) is a documented limit, not a common-case defect. CI green. Spare-quota, internal-only, non-1.0.

PR #134, opened 2026-06-29
test: external-bus coverage — delegates, signalFlow, directed signals, connection accessors

monkopedia-coder:
## What

First of two PRs lifting **external-integration API coverage** (public surface exercised by real-bus tests) from ~83% toward the 90s. Adds `ExternalApiCoverageTest` (`commonTest/integration`, runs on **both** native + JVM) covering bus-reachable public API that the large `CommonApiIntegrationTest` does not touch:

- `Object.connection` / `Proxy.connection` accessors
- `Object.currentlyProcessedMessage` read from inside a method handler
- `propDelegate` (read-only) + `mutableDelegate` (read/write) Kotlin property delegates — issue real `Properties.Get`/`Set` round-trips
- `signalFlow` reactive subscription (one emission per received signal)
- directed/unicast signal via `Signal.setDestination` + `Object.emitSignal(Signal)`

## Why

A coverage audit found these were the genuinely-uncovered, *bus-reachable* gaps in the Object/Proxy/Signals areas (the deprecated fluent property layer was excluded as it's removed at 1.0; pure-marshalling helpers like Variant introspection are unit-covered by nature). Closes the Object/Proxy/Signals gaps; caller credentials follow in PR 2.

## Backend asymmetry (documented in-test)

Directed-signal **delivery** is asserted on both backends. The inbound `DESTINATION` header is only surfaced by sd-bus; the JVM wire backend routes the unicast signal but doesn't re-expose the header to the handler — so the destination-value assertion is native-gated and commented. Flagging as a possible JVM-backend follow-up, not fixed here (test-only PR).

## Verification

Both backends, 6/6 green under `dbus-run-session`:
- `:linuxX64Test` — 6 passed
- `:jvmTest` — 6 passed
- `ktlintCheck` + `apiCheck` green (test-only, no public API change)

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/ExternalApiCoverageTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/ExternalApiCoverageTest.kt
@@ -1,0 +1,311 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.PropertyName
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.callMethod
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.method
+import com.monkopedia.sdbus.mutableDelegate
+import com.monkopedia.sdbus.onSignal
+import com.monkopedia.sdbus.prop
+import com.monkopedia.sdbus.propDelegate
+import com.monkopedia.sdbus.signal
+import com.monkopedia.sdbus.signalFlow
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):

Test-only PR; ran a skeptical pass focused on whether the assertions prove what they claim. First pass found three blocking issues in the directed-signal test (single proxy proved delivery not direction; a tautological `emitted >= 1`; a value-keyed native gate that could pass vacuously). The revision resolves all three:

- **Two proxies on distinct connections**, both subscribed before any emit; the non-targeted proxy is asserted NOT to receive the unicast signal — so a no-op `setDestination` is now caught. Genuine proof of *direction*, not just delivery.
- Behavior is gated on a compile-time `backendDeliversDirectedSignalsUnicast` expect/actual (native=true, jvm=false), not value-keyed — native runs both assertions unconditionally, so a regression to broadcast or a dropped destination header **fails** rather than silently skipping.
- The JVM broadcast limitation (`setDestination` accepted but not unicast-routed, no destination header surfaced) is pinned as an explicit regression guard with a documented follow-up: the test will fail and force the flag flip the day the JVM backend honors unicast routing.
- Dead busy-spin removed; signalFlow collector now `cancelAndJoin()`s for deterministic teardown; negative-grace window widened to 1s for CI headroom.

Verified locally: both backends 6/6 green under `dbus-run-session`, stable across 3 native + 2 JVM forced reruns. Teardown order and the negative-assertion reliability (subscribe-before-emit) check out. Mergeable once CI is green.

PR #136, opened 2026-06-29
test: external-bus coverage — Variant type introspection

monkopedia-coder:
## What

Third and final PR of the external-coverage wave. Adds `MessageIntrospectionCoverageTest` (commonTest/integration, both backends): a served method returns a `Variant`, and the proxy introspects the received value with `containsValueOfType<Int>()` / `containsValueOfType<String>()` and `peekValueType()` before `get<Int>()`. This is the last genuinely **bus-reachable** public symbol the other suites only touched indirectly (via extraction).

## Scope note

A `copyTo`-over-a-received-signal case was prototyped and dropped: `sd_bus_message_copy` on an inbound signal into a plain message is fragile/underspecified cross-backend (throws on native). `copyTo` is a low-level primitive better covered by a focused unit test, not an integration round-trip. The remaining sub-90 symbols per the audit (`signatureOf`, `buildArgs`, `RequestNameReply.fromCode`, `TypeSignature` accessors) are pure functions — unit concerns by nature, already unit-covered; `maybeDegrouped` is `internal` (not public API).

## Verification

- `:linuxX64Test` + `:jvmTest` — 1 passed each under `dbus-run-session`
- `ktlintCheck` + `apiCheck` green (test-only)

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/MessageIntrospectionCoverageTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/MessageIntrospectionCoverageTest.kt
@@ -1,0 +1,77 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.Variant
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.callMethod
+import com.monkopedia.sdbus.containsValueOfType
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Review — **APPROVE**.

Small, self-contained test. Assertions are non-vacuous: a served method returns `Variant(42)` over a real bus; the proxy asserts `containsValueOfType<Int>()` true, `containsValueOfType<String>()` false (the negative guards against a tautology), `peekValueType() == "i"`, and `get<Int>() == 42`. Teardown releases registration/proxy/object/connections in `finally`. Verified 1/1 green on both `:linuxX64Test` and `:jvmTest` under `dbus-run-session`. The dropped `copyTo` case is correctly out of scope (fragile cross-backend low-level primitive; unit-coverable). Mergeable on CI green.

PR #138, opened 2026-06-30
fix(jvm): unicast-route directed signals (Signal.setDestination)

### LINKED ISSUE #137: JVM wire backend: directed signals (Signal.setDestination) are broadcast, not unicast
## Summary

On the JVM wire backend, `Signal.setDestination(busName)` is accepted but does **not** restrict delivery: the signal is still delivered to **every** matching subscriber (broadcast), and the inbound handler does not see the `DESTINATION` header. The native (sd-bus) backend routes the same signal to the targeted peer alone and exposes the header.

## Discovered by

`ExternalApiCoverageTest.directedSignal_isDeliveredOnlyToTargetedProxy` (added in #134). The test registers two proxies on distinct connections, emits a signal directed at one, and asserts the other does **not** receive it. This passes on native and fails on JVM — so the divergence is pinned behind an `expect/actual` capability flag (`backendDeliversDirectedSignalsUnicast`: native=`true`, jvm=`false`) in `TestBackendCapabilities.kt`. The JVM branch currently asserts the broadcast behavior, so **this test will fail and force the flag flip the day the JVM backend honors unicast routing** — i.e. it is a ready-made regression guard for the fix.

## Expected

A signal with a `DESTINATION` set should be delivered only to that destination, and the recipient's `Message.destination` should reflect it — matching sd-bus / the native backend.

## Acceptance

- JVM wire backend routes directed signals to the destination only.
- Inbound directed-signal handler sees `Message.destination`.
- Flip `backendDeliversDirectedSignalsUnicast` to `true` in `src/jvmTest/.../integration/TestBackendCapabilities.jvm.kt`; the existing test then enforces unicast on both backends.

Non-blocking; surfaced while raising external-integration test coverage (#134/#135/#136).

monkopedia-coder:
Fixes #137.

## Root cause

The JVM wire backend accepted `Signal.setDestination` but **dropped the destination before the wire**: `emitWireSignal` (WireDbusBackend.kt) built a `SIGNAL` `WireMessage` with no `DESTINATION` header, so the daemon broadcast it to every matching subscriber and the recipient saw a null `Message.destination`. Native (sd-bus) routes the signal to the targeted peer alone and exposes the header — the two backends diverged.

## Fix

Thread the destination through the emit path (the inbound side already copied `wireMessage.destination` into the message metadata, so the receive half needed no change):

- `WireSignalEmitter.emitSignalOverWire` gains a `destination` parameter.
- `WireDbusObject.emitSignal` passes `message.destination?.value`. The standard-interface notifications (`PropertiesChanged`/`InterfacesAdded`/`InterfacesRemoved`) pass `null` — they stay broadcast.
- `emitWireSignal` sets it on the outgoing `WireMessage` (which already encodes the `DESTINATION` header), so the daemon unicast-routes it. The recipient then sees `Message.destination` for free.

## Verification

Flips `backendDeliversDirectedSignalsUnicast` (jvm) → `true`, so the existing two-proxy regression guard `directedSignal_isDeliveredOnlyToTargetedProxy` (added in #134) now **enforces** unicast delivery + the destination header on **both** backends — the test that previously documented the limitation now proves the fix.

Under `dbus-run-session`:
- full `:jvmTest` — **310 tests, 0 failures** (no regression from the wire-protocol change)
- native `ExternalApiCoverageTest` (6) + `CommonApiIntegrationTest` (34) green
- `ktlintCheck` + `apiCheck` green (no public API change — `WireSignalEmitter` is internal)

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/TestBackendCapabilities.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/TestBackendCapabilities.kt
@@ -5,12 +5,11 @@ package com.monkopedia.sdbus.integration
  * via [com.monkopedia.sdbus.Signal.setDestination] — only to the targeted peer, and re-exposes the
  * destination header to the receiving handler.
  *
- * - **native (sd-bus): `true`** — the bus routes the signal to the destination alone and the
- *   recipient sees the `DESTINATION` header.
- * - **JVM wire backend: `false`** — KNOWN LIMITATION: `setDestination` is accepted but the signal
- *   is still delivered to every matching subscriber (broadcast), and the inbound handler does not
- *   see the destination header. Tracked as a backend follow-up; flip this to `true` when the JVM
- *   backend honors unicast signal routing, at which point
- *   [ExternalApiCoverageTest.directedSignal_isDeliveredOnlyToTargetedProxy] will enforce it.
+ * Both backends are `true` as of #137: native (sd-bus) always did, and the JVM wire backend now
+ * threads the destination onto the outgoing wire message so the daemon unicast-routes it and the
+ * recipient sees the `DESTINATION` header. The flag is retained so
+ * [ExternalApiCoverageTest.directedSignal_isDeliveredOnlyToTargetedProxy] keeps a per-backend lever
+ * (any backend that regressed to broadcast would flip its actual to `false` rather than silently
+ * weaken the assertion).
  */
 internal expect val backendDeliversDirectedSignalsUnicast: Boolean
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):

Verified every load-bearing claim against source (using awk/Read on the binary-flagged WireDbusBackend.kt, not grep):

- **Wire round-trip**: WireMessage carries `destination` (WireMessage.kt:79), encodes the DESTINATION header on serialize (164) and reconstructs it on decode (226/249) — so setting it on the outgoing message genuinely makes the daemon unicast-route.
- **Receive half needed no change**: the inbound match already copies `wireMessage.destination` into `Message.Metadata.destination` (WireDbusBackend.kt:619), surfaced via `Message.destination` (Message.jvm.kt:175). `MatchSpec.matches` filters on path/iface/member/sender, not destination, so the targeted proxy still receives.
- **Standard-interface signals stay broadcast**: PropertiesChanged/InterfacesAdded/InterfacesRemoved all pass `destination = null`. No regression to property-change / object-manager subscribers.
- **Empty-string guard** (`takeUnless { isEmpty() }`) correctly normalizes `BusName("")` to broadcast.
- **Flag flip is non-vacuous**: with jvm actual=`true`, `directedSignal_isDeliveredOnlyToTargetedProxy` asserts both `assertFalse(otherGotIt)` and `assertEquals(target, destination)`; the targeted delivery is gated by `withTimeout`, so a broken unicast would fail rather than pass.
- **All four emit call sites accounted for**; both emit entry points (`emitSignal(Signal)` and `Signal.send()`) flow through the patched path; native flag unchanged.

Full :jvmTest (310) + native suites green per the author; I also ran compileKotlinJvm clean. Well-scoped, correct. Mergeable on CI green.

Issue #141, opened 2026-07-04
Pre-1.0 cross-backend parity hardening (native sd-bus vs JVM wire)


monkopedia-coder:
Tracking epic for the pre-1.0 cross-backend parity sweep. The JVM junixsocket wire backend diverges from native (sd-bus) in a cluster of behaviors, concentrated on the same-process local-short-circuit path (`JvmStaticDispatch`/`WireDbusObject`/`LocalJvmServiceRegistry`) — the same fault line that produced #112, #114, #137. 1.0 (API + behavioral stability) gates on the consumer-facing items.

Each fix lands with a **cross-backend regression test** (an `expect/actual` capability flag or a both-backends assertion) so the divergence can't silently return.

## Empirically confirmed (ran both backends)
- [ ] **Error-name contract**: a handler throwing a non-`SdbusException` yields error name `"boom"` (JVM) vs `NoReply` (native) — neither is the correct `org.freedesktop.DBus.Error.Failed`. Fix in `Throwable.toError()` (commonMain, affects both). *should-fix — changes documented `SdbusException.name`.*
- [ ] **Same-process standard interfaces**: `Peer.Ping`/`GetMachineId`/`Introspectable.Introspect` throw `UnknownMethod` on JVM for a same-process peer; native serves them. *should-fix.*
- [ ] **Use-after-release**: connection ops after `release()` throw uniformly on native; JVM leaves most unguarded (e.g. `uniqueName` returns a value). *should-fix — cheap guard.*

## High-confidence static (backed by native tests + code; confirm red-first when fixing)
- [ ] **Empty `InterfacesAdded` property payloads** on JVM — native fills current property values; JVM hardcodes `emptyMap`. Hits the BlueZ ObjectManager pattern (blue-falcon reads initial state here). *BLOCKER — real consumer.*
- [ ] **Empty no-arg `emitPropertiesChangedSignal(interface)`** on JVM — native emits all interface properties (changed + invalidated); JVM emits empty. *BLOCKER.*
- [ ] **Wrong argument COUNT** → `UnknownMethod` on JVM (dispatch keys on argCount) vs `InvalidArgs` on native. *should-fix.*
- [ ] **`proxy.release()`** doesn't stop signal handlers on JVM (native tears them down) — behavioral gap + handler leak. *should-fix.*
- [ ] **`Connection.addMatch` with a well-known `sender=`** drops all signals on JVM (local re-filter compares unique name to the well-known name). *should-fix.*
[...]

PR #142, opened 2026-07-04
fix: valid cross-backend error name for raw handler exceptions (#141)

monkopedia-coder:
First PR of the pre-1.0 parity-hardening wave (#141). Empirically-confirmed divergence.

## Problem

A handler throwing a non-`SdbusException` went through `Throwable.toError()`, which put the exception **message** in the error-**name** slot (`name="boom"`). That's not a valid D-Bus error name, so the backends diverged:
- **native** failed to send a proper error reply → client saw `org.freedesktop.DBus.Error.NoReply`
- **JVM** passed the bogus `"boom"` through as the name

Either way the documented `SdbusException.name` contract was wrong *and* backend-dependent.

## Fix

`toError()` maps a non-`SdbusException` to `org.freedesktop.DBus.Error.Failed` (the generic D-Bus failure name) with the exception message as the detail. commonMain, so both backends converge.

## Regression test (red-first)

`FailurePathParityTest.handlerThrowingRawException_reportsFailedErrorNameOnBothBackends` — runs on both targets, asserts `.name == org.freedesktop.DBus.Error.Failed`. Verified **red before** (native `NoReply` / JVM `boom`), **green after** on both.

## Verification
- Full `:jvmTest` + `:linuxX64Test` — **591 tests, 0 failures** (no regression from the commonMain change)

### DIFF
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Error.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Error.kt
@@ -40,8 +40,14 @@ package com.monkopedia.sdbus
 class SdbusException(val name: String, val errorMessage: String = "") :
     Exception("$name: $errorMessage")
 
-internal fun Throwable.toError() =
-    (this as? SdbusException) ?: SdbusException(message ?: toString(), stackTraceToString())
+// The generic D-Bus error name for a failure with no more specific name. A non-SdbusException from
+// a handler maps to this VALID name so both backends report it identically; the former behavior put
+// the exception message in the name slot, producing an invalid name — native then failed to send a
+// proper error reply (client saw NoReply) while JVM passed the bogus name through. See #141.
+internal const val GENERIC_DBUS_ERROR_NAME = "org.freedesktop.DBus.Error.Failed"
+
+internal fun Throwable.toError(): SdbusException =
+    (this as? SdbusException) ?: SdbusException(GENERIC_DBUS_ERROR_NAME, message ?: toString())
 
 /**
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):

Traced the fix end to end (rg/awk/Read):
- `org.freedesktop.DBus.Error.Failed` is the canonical generic name (already used at ObjectImpl.kt:462 and as the JVM default at WireDbusBackend.kt:441). Correct choice; message correctly preserved as errorMessage.
- Short-circuit `(this as? SdbusException) ?: …` unchanged, so named errors still round-trip their own name.
- Blast radius contained: `toError()` is internal; callers are MethodBuilder (the target path) + client reply-plumbing in TestProxy where `it` is already an SdbusException (short-circuits). No test relied on the old message-derived name; author's 591-green both backends is consistent.
- The native `invokeHandlerAndCatchErrors`/`org.sdbuscpp.Error` fallback is a SEPARATE path — the typed `call{}` throw is caught in TypedMethodCall.fold → MethodBuilder.createErrorReply(toError()) and the vtable callback returns normally, so `org.sdbuscpp.Error` is never reached for this case. No inconsistency; cleaning up those edge paths is a separate ticket.
- Regression test asserts name==Failed on the shared commonTest suite (both backends); the `alwaysThrows` indirection correctly keeps the handler's return type Int (a bare throw infers Nothing → no serializer at registration). Red-first confirmed.
- Message-text change (stackTrace → message) stops leaking a per-backend stack trace across the wire; improvement.

Mergeable on CI green.

PR #143, opened 2026-07-04
fix(jvm): emit current property values in InterfacesAdded / no-arg PropertiesChanged (#141)

monkopedia-coder:
Parity-hardening wave (#141). **BlueZ-critical** — this is the one with a real consumer (blue-falcon).

## Problem

The JVM wire backend emitted **empty** property maps for:
- ObjectManager `InterfacesAdded` (`interfaces.associateWith { emptyMap() }`)
- the no-argument `emitPropertiesChangedSignal(interface)` (passed `emptyList()`)

Native (sd-bus) fills current values in both. An ObjectManager consumer that reads initial object/device state from these signals — the standard BlueZ pattern **blue-falcon** uses — got nothing on JVM.

## Fix (JVM only; native already correct)

- `emitInterfacesAddedSignal` serializes each interface's current property values via the existing `snapshotInterfaceProperties()` helper.
- no-arg `emitPropertiesChangedSignal(interface)` resolves **all** the interface's properties (readable → `changed` with value, write-only → `invalidated`), mirroring `sd_bus_emit_properties_changed_strv(..., names=null)`.

## Regression test (red-first)

`SignalPayloadParityTest` (both backends) decodes the signal payloads with typed handlers and asserts the property value arrives. **Red-first confirmed:** native 2/2 green, JVM 2/2 red (payload value `null`) before; green on both after.

## Verification
- `SignalPayloadParityTest` — 2/2 green on **both** backends under `dbus-run-session`
- Dbusmock ObjectManager/PropertiesChanged + `CommonApiIntegrationTest` suites unaffected

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/SignalPayloadParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/SignalPayloadParityTest.kt
@@ -1,0 +1,137 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectManagerProxy
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.PropertiesProxy
+import com.monkopedia.sdbus.PropertyName
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.Variant
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.method
+import com.monkopedia.sdbus.onSignal
+import com.monkopedia.sdbus.prop
+import kotlin.random.Random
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):

Verified (rg/awk/Read):
- `emitInterfacesAddedSignal` now routes through the pre-existing `snapshotInterfaceProperties()` (invokes each property getter, `mapNotNull` skips write-only/no-getter) — so InterfacesAdded now matches what `GetManagedObjects` already returned. Correct invariant.
- No-arg `emitPropertiesChangedSignal(interface)` passes all property names to `resolveChangedProperties` (readable→changed, write-only→invalidated), matching native's `sd_bus_emit_properties_changed_strv(..., names=null)`.
- Getter-throws-at-emit propagates (no runCatching in snapshotInterfaceProperties) — pre-existing, and parity-correct (native `sd_bus_emit_interfaces_added` also aborts on getter failure). Not introduced here.
- Regression test uses the exact typed-decode patterns the production ObjectManagerProxy/PropertiesProxy use; asserts the VALUE (42), not mere presence; sound on both backends.
- No blast radius: JvmSignalPropertyUnsupportedTest (no props) → empty payload as before; no codegen/consumer expects empty.

Red-first confirmed (native green / JVM null before). Mergeable on CI green. (Follow-up noted separately: standard interfaces aren't enumerated in InterfacesAdded — interface-enumeration gap, orthogonal to this value fix.)

PR #144, opened 2026-07-04
fix(jvm): guard connection ops against use-after-release (#141)

monkopedia-coder:
Parity-hardening wave (#141). Empirically-confirmed divergence (probe: JVM `uniqueName` after release returned `:1.2`; native threw `IllegalArgumentException`).

## Problem

`WireDbusConnection` set `released` but only `WireDbusProxy.callMethod` checked it. Connection ops (`uniqueName`, `startEventLoop`, `setMethodCallTimeout`, `getMethodCallTimeout`, `addObjectManager`, `requestName`, `releaseName`, `addMatch`) ran after `release()` — silently succeeding or throwing an unrelated closed-socket error. Native guards every op with `checkNotReleased()`.

## Fix

Adds a JVM `checkNotReleased()` that throws the **same** `IllegalArgumentException("Connection has already been released")` as native (`ConnectionImpl.checkNotReleased`), applied to each public op.

## Regression test

`FailurePathParityTest.connectionOperationsAfterReleaseThrowOnBothBackends` — asserts `uniqueName` / `startEventLoop` / `methodCallTimeout=` all throw `IllegalArgumentException` after release on **both** backends.

## Verification
- Full `:jvmTest` — **311 tests, 0 failures** (confirms no test relied on calling an op after release, e.g. `stopEventLoop` in teardown — teardown always stops *before* release)
- native `FailurePathParityTest` green; `ktlintCheck` + `apiCheck` green (no public API change)

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
@@ -292,6 +292,22 @@ class FailurePathParityTest {
         }
     }
 
+    // Every public connection operation must reject use after release() the same way on both
+    // backends. Native guards each with checkNotReleased() (an IllegalArgumentException); the JVM
+    // backend used to leave most ops unguarded (e.g. uniqueName silently returned a value). #141.
+    @Test
+    fun connectionOperationsAfterReleaseThrowOnBothBackends() = runBlocking {
+        val connection = createBusConnection()
+        connection.release()
+
+        assertFailsWith<IllegalArgumentException> { connection.uniqueName }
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review — APPROVE. checkNotReleased() is byte-identical to native (require → IllegalArgumentException, same message). All expression→block conversions preserve return values/types; release() keeps its compareAndSet idempotency and isn't guarded (double-release stays a no-op, matching native); currentlyProcessedMessage correctly left unguarded. No teardown regression — native has always guarded start/stopEventLoop and the shared suite runs against it, so teardown already stops-before-release; 311 green is consistent. Regression test is non-vacuous (asserts IllegalArgumentException on uniqueName/startEventLoop/methodCallTimeout after release, both backends). Mechanical, correct. Mergeable on CI green.

PR #145, opened 2026-07-04
fix(jvm): proxy.release() tears down its signal handlers (#141)

monkopedia-coder:
Parity-hardening wave (#141). Red-first confirmed divergence.

## Problem

Native `ProxyImpl.release()` releases the proxy's floating signal slots → handlers stop firing. JVM `WireDbusProxy.release()` was a **no-op**: `registerSignalHandler` returned an independent resource the proxy never collected, so after `proxy.release()` handlers kept firing (and leaked).

## Fix

`WireDbusProxy` collects each signal registration and releases them all on `release()`. `registerSignalHandler` returns a wrapper that also removes the entry on an explicit release; both the wrapper and the underlying `ActionResource` are idempotent (double-release safe). Registrations held in a `CopyOnWriteArrayList` (register/release happen on user threads).

## Regression test

`ProxyLifecycleParityTest.proxyReleaseStopsSignalHandlers` — registers a handler (keeping **no** reference to its resource), proves it fires, then `proxy.release()` and keeps emitting, asserting no further delivery. **Red-first:** native green / JVM red (kept firing at `expected 1, was >1`) before; green on both after.

## Verification
- test 1/1 green on **both** backends
- clean full `:jvmTest` (`--rerun-tasks`) — **315 tests, 0 failures**

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/ProxyLifecycleParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/ProxyLifecycleParityTest.kt
@@ -1,0 +1,89 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.emitSignal
+import com.monkopedia.sdbus.method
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review — APPROVE. release() releases each collected registration (removeMatch + close → delivery genuinely stops), then clears. The returned wrapper removes-then-releases, so explicit release de-lists (stricter than native, which accumulates dead slots). Double-release safe: underlying installSignalMatch resource is doubly CAS-guarded (inner AtomicBoolean + ActionResource) and the wrapper is CAS-guarded; forEach{runCatching{}} swallows strays. CopyOnWriteArrayList is right; the register-concurrent-with-release window is benign caller-misuse (native is worse — unsynchronized). No leak (clear drops refs; de-list bounds growth). Test is sound: warm-up drains in-flight before snapshotting countAtRelease, atomicfu gives cross-thread visibility (handlers run on a worker, not reader), post-release emits are strictly after the synchronous removeMatch. No existing signal test touched; 315 green. Mergeable on CI green.

PR #146, opened 2026-07-04
fix(jvm): wrong argument count -> InvalidArgs, not UnknownMethod (#141)

monkopedia-coder:
Parity-hardening wave (#141), 5 of 6. Empirically-confirmed target: native returns `InvalidArgs` for wrong-arg-count, `UnknownMethod` only for a missing member; JVM returned `UnknownMethod` for both.

## Problem

JVM `JvmStaticDispatch.MethodKey` includes `argCount`, so a call to an existing member with the wrong number of arguments misses the handler and is classified `UnknownMethod` (both the local-dispatch and wire-serve paths). Native dispatches by name → deserialization fails → `InvalidArgs`.

## Fix

- `JvmStaticDispatch.hasMember(path, iface, method, destination)` — argCount-agnostic lookup.
- Both classification sites (`WireDbusBackend` local-miss, `WireServe.classifyMissing`) now return `InvalidArgs` when the member exists at another arity, `UnknownMethod` only for a truly-absent member.

## Regression test

`FailurePathParityTest.wrongArgumentCountIsInvalidArgsWhileMissingMemberIsUnknownMethod` — too-many / too-few args → `InvalidArgs`; missing member → `UnknownMethod`; asserted on **both** backends. (Target confirmed via a throwaway probe: native `InvalidArgs`, JVM `UnknownMethod` before.)

## Verification
- clean full `:jvmTest` (`--rerun-tasks`) — **316 tests, 0 failures**

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
@@ -342,6 +342,49 @@ class FailurePathParityTest {
         }
     }
 
+    // A call to an existing member with the WRONG NUMBER of arguments must surface as InvalidArgs
+    // on both backends (native dispatches by member name and fails arg deserialization → InvalidArgs;
+    // the JVM backend used to key dispatch on arg count and return UnknownMethod). A truly-absent
+    // member must still be UnknownMethod. Parity #141.
+    @Test
+    fun wrongArgumentCountIsInvalidArgsWhileMissingMemberIsUnknownMethod() = runBlocking {
+        val ids = uniqueFixtureIds("argCount")
+        val serverConnection = createBusConnection(ids.service)
+        val proxyConnection = createBusConnection()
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):
Independent review — APPROVE. hasMember matches path+interface+method (argCount-agnostic) with dest guard covering resolveHandler's exact + wildcard cases — no false-positive (never matches an unreachable concrete dest), and the one false-negative (skips the single-candidate fallback) fails SAFE to the pre-PR UnknownMethod and is unreachable with current wiring (every registration uses dispatchDestination=senderName; both sites pass that). Both sites pass the correct destination (local: dispatchDestination=localOwner; wire: localUniqueName=senderName). No over-broad InvalidArgs: interfaceName matched exactly, wire branch ordered after UnknownInterface and guarded iface!=null. hasMember only runs on the already-missed path, so successful dispatch untouched; 316 green. Test is sound (createMethodCall+append yields wrong-count; InvalidArgs is the empirically-confirmed native target). Mergeable on CI green.

PR #147, opened 2026-07-04
fix(jvm): serve Peer/Introspectable for same-process callers (#141)

monkopedia-coder:
Parity-hardening wave (#141), **6 of 6 — the final fix**. Empirically-confirmed divergence (probe: native answers same-process Peer.Ping/Introspect; JVM threw UnknownMethod).

## Problem

`org.freedesktop.DBus.Peer` (Ping/GetMachineId) and `Introspectable` (Introspect) are served by `WireServe` on the incoming-wire path, but they aren't registered in local dispatch. So a **same-process** call — which short-circuits through `JvmStaticDispatch` — threw `UnknownMethod`, while native serves them via sd-bus regardless of same- vs cross-process. (`Properties`, the third standard interface, already worked same-process because it *is* registered in local dispatch.)

## Fix

When a same-process call targets Peer or Introspectable and no user handler matches, route it over the wire (`callRemote`) so `WireServe.servePeer`/`serveIntrospect` answers it — **reusing the exact cross-process serving path** rather than duplicating the Ping/machine-id/XML logic. The wire serve (phase 4) always replies, so it can't hang (the old "never reply" comment predates incoming-call serving). User-method misses keep the local `InvalidArgs`/`UnknownMethod` classification from #146 (the standard-interface branch is checked first).

## Regression test

`StandardInterfaceParityTest.sameProcessPeerAndIntrospectAreServed` — same-process `Peer.Ping` succeeds, `GetMachineId` non-empty, `Introspect` returns XML naming the served interface; on **both** backends. **Red-first:** native green, JVM `UnknownMethod` before; green on both after.

## Verification
- clean full `:jvmTest` (`--rerun-tasks`) — **317 tests, 0 failures**
- #146 arg-count classification still correct (re-verified)

Completes the pre-1.0 parity wave (#141).

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/StandardInterfaceParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/StandardInterfaceParityTest.kt
@@ -1,0 +1,70 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.PeerProxy
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.callMethod
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.method
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertTrue
+import kotlinx.coroutines.runBlocking
+
+/**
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review — APPROVE (final parity-wave fix). No deadlock: three distinct threads — caller blocks in callBlocking, the per-connection reader (always running on JVM, not gated on runEventLoopThread) receives the routed-back call and later completes the future, and the serve worker runs handleIncomingCall — so a runEventLoopThread=false client still gets the reply; callBlocking defaults to a 30s timeout so no infinite hang. Routing correct: callRemote sends to destination.value (owned by the server connection → daemon routes back); serveIntrospect(localUniqueName) matches the WireServeRegistry key (senderName) so addVTable interfaces are found. Scope preserved: the Peer/Introspectable guard returns early first; user-method misses keep #146's InvalidArgs/UnknownMethod (test unmodified + reachable); Properties correctly stays on the fast local path (registered in JvmStaticDispatch). Same-connection edge safe (still cross-thread). Test genuinely same-process on both backends, asserts real answers. Mergeable on CI green.

PR #148, opened 2026-07-04
fix(jvm): make JvmStaticDispatch thread-safe (data race)

monkopedia-coder:
Pre-1.0 correctness fix (found by the fresh-pass audit; tracked on #141).

## Problem

`JvmStaticDispatch.handlers` was a plain `HashMap` — the **only** one of the five shared JVM registries left unsynchronized. It's mutated on user threads (`addVTable`/`release`) while dispatch reads (`hasHandler`/`hasMember`/`invokeOrNull`) run concurrently on serve-worker, reader, and caller threads. Exporting an object while another is being served can throw `ConcurrentModificationException` (the resolve/hasMember scans iterate the key set) or return a spurious `null` (a handler "vanishes" → misclassified `UnknownMethod`/`InvalidArgs`).

## Fix

`ConcurrentHashMap` — atomic put/remove, weakly-consistent iteration — matching the `synchronized` wrapping the sibling registries already use.

## Regression test

`JvmStaticDispatchTest.concurrentRegisterUnregisterAndResolveIsRaceFree` — 8 threads × 4000 iterations registering/unregistering while resolving. **Reliably reproduced the race: threw `ConcurrentModificationException` 3/3 runs before the fix, green 5/5 after.**

## Verification
- full `:jvmTest` — **318 tests, 0 failures**

### DIFF
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/JvmStaticDispatch.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/JvmStaticDispatch.kt
@@ -1,6 +1,7 @@
 package com.monkopedia.sdbus.internal.jvmdbus
 
 import com.monkopedia.sdbus.MethodReply
+import java.util.concurrent.ConcurrentHashMap
 
 /**
  * JVM-side static dispatch table for method calls.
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Correctness fix, empirically validated. ConcurrentHashMap is the right choice: register/unregister become atomic put/remove; resolveHandler's filterKeys{} scan and hasMember's keys.any{} become weakly-consistent (no CME); a concurrent remove yielding null on invokeOrNull is the correct 'not currently registered' answer. Matches the synchronized wrapping the four sibling registries already use. The stress test reliably reproduced the ConcurrentModificationException (3/3 before) and is green 5/5 after; full jvmTest 318 green; no public API change. Mergeable on CI green.

PR #149, opened 2026-07-04
fix(jvm): honor dontExpectReply / fire-and-forget calls (#141)

monkopedia-coder:
Pre-1.0 correctness fix (fresh-pass audit; #141). One of the two consumer-facing items I'd missed in the first wave.

## Problem

The JVM client **ignored** `MethodCall.dontExpectReply`: it ran `callBlocking` and waited up to ~30s for a reply a no-reply peer won't send, and threw `UnknownMethod` for a no-reply call to a missing member. Native does `sd_bus_send` with no reply — deliver for side effects, never wait, never surface an error for a missing target.

## Fix

`WireDbusProxy.callMethod` short-circuits when `dontExpectReply` is set: run the local handler for side effects if one is registered, else send over the wire with `NO_REPLY_EXPECTED` and don't wait; a missing same-process member is silently dropped. Returns an empty reply.

## Regression test

`NoReplyParityTest` — a no-reply call delivers the handler side effect without blocking, and a no-reply call to a missing member does not throw; both backends. **Red-first:** native green, JVM threw `UnknownMethod` before; green on both after.

## Verification
- clean full `:jvmTest` — **319 tests, 0 failures**

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/NoReplyParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/NoReplyParityTest.kt
@@ -1,0 +1,80 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.callMethod
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.method
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertTrue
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review — APPROVE. Matches native sd_bus_send no-reply for all three cases: (a) local handler runs the side effect, (b) localOwner==null → sendNoReply over the wire with NO_REPLY_EXPECTED (0x1) via send() which never inserts into the pending map — no leak, no 30s wait, wire build identical to callRemote bar the flag, and WireServe already honors the flag server-side, (c) local-owner-no-handler dropped (previously threw UnknownMethod). runCatching is correct not a bug-hider: callLocalDispatch runs in the caller thread so a throwing local handler must be swallowed to match native (side effect committed up to the throw). emptyReply() mirrors native MethodReply(sdbus) and is discarded — MethodInvoker special-cases dontExpectReply and returns Unit as R without deserializing. Release guard is the first statement, before the branch, so a no-reply call on a released connection still throws. Test proves no-wait (withTimeout 5s vs the old 30s block) + side-effect delivery + no-throw-on-missing-member on both backends (case b cross-process is a minor follow-up coverage gap, structurally callRemote+flag). Mergeable on CI green.

PR #150, opened 2026-07-04
fix(jvm): resolve well-known sender= in Connection.addMatch (#141)

monkopedia-coder:
Pre-1.0 correctness fix (fresh-pass audit; #141) — the last of the three. The second consumer-facing item I'd missed in the first wave.

## Problem

A `Connection.addMatch` rule with a well-known `sender=` **dropped every signal** on JVM: `parseMatchSpec` recorded the well-known name but never resolved its owner, so the local re-filter (`WireMessage.matches`) compared the daemon-stamped unique sender (`:1.x`) against the well-known name and rejected all frames. `registerSignalHandler` already resolves the owner; the low-level `addMatch` path didn't.

## Fix

`addMatch` resolves a well-known `sender=` to its current unique owner (`withResolvedOwner` → `getNameOwner`), mirroring `registerSignalHandler`, so `matches()` compares against the resolved owner. A unique-name (`:`) or absent sender is unchanged.

## Regression test

`AddMatchSenderParityTest` — `addMatch` with a rule naming a well-known `sender=` receives the matching signal, both backends. **Red-first:** native green, JVM dropped it (never fired) before; green on both after.

## Verification
- clean full `:jvmTest` — **320 tests, 0 failures**

Completes the fresh-pass correctness punch-list (with #148 JvmStaticDispatch race, #149 dontExpectReply).

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/AddMatchSenderParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/AddMatchSenderParityTest.kt
@@ -1,0 +1,77 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.emitSignal
+import com.monkopedia.sdbus.method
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Independent review — APPROVE. matches() checks resolvedOwner (sender==spec.sender || (resolvedOwner!=null && sender==resolvedOwner)), so a daemon-stamped :1.x frame now matches a well-known-sender rule. withResolvedOwner correctly skips :-prefixed/null/empty and resolves only well-known names — logically identical to registerSignalHandler's gate. The outer runCatching is redundant (getNameOwner already catches → null) but harmless. Unowned name → null → matches nothing, mirroring native. Staleness is pre-existing (one-shot resolution, same as registerSignalHandler), not worsened. Test uses non-direct createBusConnection so the daemon stamps unique senders and the fix path is genuinely exercised; red-first on JVM. Mergeable on CI green.

PR #154, opened 2026-07-15
chore: drop unused kotlinx-datetime dependency

monkopedia-coder:
## What

`kotlinx-datetime` was a declared `implementation` dependency in the library (commonMain / jvmMain / nativeMain), both samples, and `compile_test` — but it is used **nowhere**. The codebase relies on `kotlin.time` (stdlib) for `Duration` and `Clock`; there are zero `kotlinx.datetime` references anywhere in source.

This drops it entirely and removes the two version-catalog entries.

## Why now

Surfaced during the 1.0.1 dependency-refresh review. Rather than bump an unused dep to 0.8.0, remove it — it comes off consumers' transitive runtime classpath, shrinking their dep tree for free.

## Impact

- **ABI: unchanged.** `implementation`-scoped and absent from the public API, so `apiCheck` stays green and both `api/*.api` (JVM) and `api/*.klib.api` (klib) dumps are untouched — this diff touches no file under `api/`.
- **Consumers (blue-falcon):** still a drop-in; they don't reference kotlinx-datetime via this library.

## Verification (all green, local)

- `compileKotlinJvm` + `compileKotlinLinuxX64` + `compileKotlinLinuxArm64` + `:compile_test:compileKotlinLinuxX64`
- `ktlintCheck`
- `apiCheck` — no ABI change
- `jvmTest` 320/320 · `linuxX64Test` 288/288 (under `dbus-run-session`)

Rolls into the 1.0.1 maintenance release (version already bumped on main).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,13 @@ klib binary-compatibility dumps are identical to 1.0.0), so this is a drop-in up
 - **Kotlin** `2.4.0` → `2.4.10` (toolchain; no klib-ABI shift).
 - **kotlinx-coroutines** `1.10.2` → `1.11.0`.
 - **kotlinx-atomicfu** `0.32.1` → `0.33.0`.
-- **kotlinx-datetime** `0.7.1` → `0.8.0`.
+
+### Removed — dependencies
+
+- **kotlinx-datetime** is no longer a dependency. It was declared but unused (the library uses
+  `kotlin.time` from the standard library), so it has been dropped from the published artifacts —
+  removing it from consumers' transitive runtime classpath. This is `implementation`-scoped and
+  absent from the public API, so it does not affect the binary-compatibility surface.
 
 ### Documentation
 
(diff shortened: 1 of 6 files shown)


monkopedia-reviewer (approved):
Independent fresh-context review (rolls into 1.0.1, verify-clean). All verified:

- **Truly unused.** `rg kotlinx.datetime` repo-wide → only hit is CHANGELOG prose. Zero source usages; the only time-symbol hits are `import kotlin.time.Clock` (stdlib) in DBusMethodTests.kt:44,184,200 — not kotlinx.datetime. Removal can't break compile.
- **Complete removal.** All six `implementation(libs.kotlinx.datetime)` lines gone (lib common/jvm/native, compile_test, both samples); both catalog entries removed; no dangling `version.ref = "kotlin-datetime"`, no other entry referenced it.
- **ABI unchanged.** `git diff --stat main...HEAD -- api/` empty — no dump touched.
- **Scope clean.** 6 files, build/catalog/CHANGELOG only, zero source/logic changes.
- **CHANGELOG consistent.** Old "datetime 0.7.1→0.8.0" line removed; new "Removed — dependencies" section accurate; "no API/behavioral change, drop-in" framing still holds (implementation-scoped, unused).

CI green (13/13, 2 skip). Simplest correct form of the change. APPROVE.

PR #152, opened 2026-07-08
docs: describe the JVM backend by what it is, not "no dbus-java"

monkopedia-coder:
## What

Post-1.0 doc/comment cleanup. The JVM backend replaced dbus-java back in 0.5.0, but a scattering of docs and test comments still framed it negatively ("no dbus-java") or, worse, mislabeled our own pure-Kotlin junixsocket backend AS "the JVM dbus-java backend".

## Changes

- **User-facing docs** (`docs/BACKENDS.md`, `dokka/moduledoc.md`, `samples/demo-service/README.md`): drop the "no dbus-java" aside and describe the backend by what it is (owned junixsocket connection, pure-Kotlin marshaller/dispatcher). Kept the real attribute "no native code".
- **cross_test `Dbusmock*` / `TypeMatrix` comments**: `the JVM dbus-java backend (jvmTest)` → `the JVM wire backend (jvmTest)` — it is not dbus-java.
- **`JvmMessageSupportTest` / `JvmRealBusIntegrationTest`**: drop stale "from dbus-java" phrasing; the old backend "did (before it was replaced)".

## Deliberately kept

Legitimate references are untouched:
- `cross_test/build.gradle.kts`'s real `dbus-java` dependency — used as an independent interop peer for cross-runtime tests.
- The "old dbus-java backend" regression-history notes (issues #35/#36).
- The epic #93 migration narrative in `WireSignalEmissionExternalTest`.

## Verification

Comment/doc-only changes; `compileTestKotlinJvm` is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #161, opened 2026-07-29
fix(jvm): escape the raw NUL char literal in WireDbusBackend.kt

### LINKED ISSUE #155: Raw NUL byte in WireDbusBackend.kt flags the file as binary (breaks grep/rg)
**Category:** hygiene

**Where:** `src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt:785`

**Current:** the line is a `.trim(...)` call whose first argument, which *displays* as a blank, is actually a raw NUL byte (`0x00`) embedded literally in the source. It strips the trailing NUL + whitespace off the SELinux context read from `/proc/self/attr/current` (that pseudo-file is NUL-terminated), so the runtime behavior is correct.

**Proposed:** replace the raw-NUL character literal with the escaped Kotlin form for code point zero (a single-quoted backslash-u-0000), or equivalently `Char(0)`. Behavior-identical; pure hygiene.

**Why:** this single NUL is the *only* raw NUL byte in the entire source tree, and it causes `file(1)` to classify this 1057-line file as `data` rather than text. That is the root cause of the well-known repo trap where **`grep`/`rg` silently return nothing on `WireDbusBackend.kt`** — the search returns empty rather than an error, so it *looks* like nothing matched. This has caused real tooling confusion (including a reviewer once wrongly concluding code was absent from this file). Fixing the one byte makes the core JVM-backend file greppable again; every future search / refactor / review over the wire backend benefits.

**Effort:** mechanical (one character)

monkopedia-coder:
One-character lexical fix: the SELinux-context trim in `localProcessWireCredentials` used a **raw 0x00 byte** as its first char literal. It is now written as the ordinary Kotlin unicode escape for code point zero (backslash-u-0000), which denotes the same character — so the trim behaves identically. This changes only how the source is encoded, not what it does.

## Why it matters

That one byte made `file(1)` classify the whole 1057-line source file as `data` instead of text:

```
$ file src/jvmMain/.../WireDbusBackend.kt     # before
... : data
$ file src/jvmMain/.../WireDbusBackend.kt     # after
... : Unicode text, UTF-8 text
```

Which means **grep and rg silently print nothing for matches in this file** — binary files suppress line output unless `-a` is passed. An empty grep here reads as "the string is absent" when it is present.

Two concrete costs, both already paid:

1. **The tree-wide doc sweep in #152 missed this file entirely.** That PR removed stale references to the retired dbus-java backend across the repo; this file still carries **ten** of them. `grep -c dbus-java` on it returns 10 now that the file is greppable, and returned nothing before.
2. An earlier review round was nearly dismissed as fabricated citations, on the strength of an empty grep over this file. The citations were correct.

It is the only raw NUL in any tracked *source* file. The other NUL-bearing tracked files are legitimately binary (jars, `.so`, png), confirmed by a byte scan over `git ls-files`.

Worth noting for anyone verifying this: `grep -P '\x00'` is itself unreliable for finding the problem, for the same reason. A `perl -0777 -ne` scan or `od -c` is the dependable check.

## Scope

Deliberately just the one byte. The ten stale dbus-java comments that this fix *reveals* are a separate concern — they are documentation, they need judgment about which references are genuinely stale versus legitimately historical, and folding them in here would obscure a change whose entire value is being trivially verifiable. A follow-up PR completes #152's sweep in this file.

Fixes #155

## Verification

- `compileKotlinJvm`, `ktlintCheck`, `apiCheck` — green
- No `api/` file touched, so no ABI change
- `jvmTest` 320/320 + `cross_test:jvmTest` 64/64, 0 failures, under `dbus-run-session`
- `grep -n localProcessWireCredentials <file>` now resolves (line 780); it returned empty before

### DIFF
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt
@@ -782,7 +782,7 @@ private val localProcessWireCredentials: WireSenderCredentials by lazy {
     val unix = runCatching { com.sun.security.auth.module.UnixSystem() }.getOrNull()
     val selinuxContext = runCatching {
         java.nio.file.Files.readString(java.nio.file.Paths.get("/proc/self/attr/current"))
-            .trim(' ', ' ')
+            .trim('\u0000', ' ')
             .ifEmpty { null }
     }.getOrNull()
     WireSenderCredentials(


monkopedia-reviewer (approved):
Tier-1 mechanical fix, verified independently rather than taken on the coder's word.

Correctness (A):
- Confirmed the actual git-visible defect at the pre-fix commit (b7c8e67): file(1) reports the source `data`, and the raw NUL byte sits at offset 34359 of a 45433-byte file. Reproduced the real symptom directly: `grep -n dbus-java` on that blob returns exit 1 / zero matches, and `rg` reports the generic "binary file matches (found NUL byte around offset 34359)" instead of line numbers, even though 10 real matches exist (`grep -a` surfaces all 10).
- At HEAD (b77ceaa): file(1) now reports "Unicode text, UTF-8 text", `git diff --numstat` shows `1  1` (not the binary `-  -` tell), the diff renders as an ordinary textual hunk, and both `grep -c dbus-java` and `rg -c dbus-java` now correctly return 10.
- Codepoint equivalence: Kotlin has no backslash-zero escape; the unicode escape for code point zero is the only valid char-literal spelling of NUL, and it denotes code point U+0000 -- the same value a raw 0x00 byte encodes as a single-byte UTF-8 sequence. Confirmed the two are bit-for-bit the same value, so the trim call behaves identically to the byte it replaces. No wire-protocol behavior change.
- Sibling sweep (independent, not reusing `grep -P` NUL matching which the PR itself correctly flags as unreliable on binary-classified files): wrote a byte-level Python scan over every tracked file. All 24 tracked files containing a NUL byte at HEAD are legitimately binary (pngs, jars, .so files) -- zero remaining raw NULs in source. A second sweep for other stray control bytes (excluding tab/LF/CR) across all non-png/jar/so tracked files returned zero hits. This was a genuine one-off, not a symptom of a wider pattern.
- Scope: single-line, single-file diff, no ABI/API surface touched (no api/*.api file in the diff), matches "Fixes #155" exactly. No scope creep -- the ten now-visible stale dbus-java doc comments are correctly left to a follow-up rather than bundled in.
- CI: all required checks green (static-checks, full-tests-x64, native-build-x64, arm-runtime-tests, all pr-smoke matrix legs). cleaner-soak/nightly-stress show "skipping", which is their normal nightly-only state, not a gate here.

Quality (B): matches surrounding style -- the replacement char literal mirrors the existing single-char-literal idiom used two lines away (parts[1].trim().trim(single-quote-escape).trim(double-quote) at line 692). No style-guide violations.

No test was added, but none was needed: this is a lexical-encoding-only change with no behavioral branch to cover, and its correctness is fully machine-verifiable via file(1)/git diff --numstat/codepoint identity, all of which were reproduced independently above rather than trusting the PR description.

Approving and merging as tier-1 per review-policy.md (sdbus-kotlin has no override; bug_fix defaults to tier 1).

PR #162, opened 2026-07-29
chore: refresh build tooling to latest stable

### LINKED ISSUE #156: Refresh build-tooling deps to latest stable (Gradle 9.6.1, ktlint 14.2.0, vanniktech 0.37.0, mockk 1.14.11)
**Category:** dep-bump

**Where:** `gradle/libs.versions.toml` + `gradle/wrapper/gradle-wrapper.properties`. Coordinates behind latest stable (verified against `repo1.maven.org` / Gradle Plugin Portal `maven-metadata.xml`, stable only — RC/beta/milestone excluded):

| Coordinate | Current | Latest stable | Jump |
| --- | --- | --- | --- |
| Gradle wrapper | 9.2.1 | 9.6.1 | minor |
| `org.jlleitschuh.gradle.ktlint` (plugin) | 14.0.1 | 14.2.0 | minor |
| `com.vanniktech.maven.publish` (plugin) | 0.36.0 | 0.37.0 | minor |
| `io.mockk:mockk` (test-only) | 1.14.9 | 1.14.11 | patch |

**Current:** these build-tooling versions were left untouched by the 1.0.1 runtime-dependency refresh (which brought Kotlin, coroutines, atomicfu, serialization, junixsocket to latest stable). They are now a few releases behind.

**Proposed:** bump each to the latest-stable version above in one change. All are minor/patch with no expected migration surface. Suggest a quick `publishToMavenLocal` smoke-test after the vanniktech bump (it touches the publish path).

**Why:** staying current on build tooling avoids painful multi-version catch-up later; these are all low-risk. No runtime/ABI impact — none of these ship in the published library artifacts, so there is **no klib-ABI or downstream (blue-falcon) sequencing concern** for this bump.

**Not included (couldn't confirm online):** `com.github.gmazzo.buildconfig` (currently 6.0.9) — the metadata path 404'd during verification, so I'm not asserting a target. Already at latest stable and deliberately excluded: kotlin 2.4.10, kotlinx-coroutines 1.11.0, kotlinx-serialization 1.11.0, atomicfu 0.33.0, junixsocket 2.10.1, kotlinpoet 2.3.0, dokka 2.2.0, binary-compatibility-validator 0.18.1, clikt 5.1.0, gradle-plugin-publish 2.1.1, hierynomus-license 0.16.1. The **xmlutil 0.91.3 → 1.0.1 major** bump is filed separately (real migration surface).

**Effort:** mechanical

monkopedia-coder:
Build-tooling refresh. None of these ship in the published artifacts, so there is no runtime, ABI, or downstream (blue-falcon-sdbus) impact — `apiCheck` is green and no file under `api/` is touched.

| Coordinate | From | To |
| --- | --- | --- |
| Gradle wrapper | 9.2.1 | 9.6.1 |
| `org.jlleitschuh.gradle.ktlint` | 14.0.1 | 14.2.0 |
| `com.vanniktech.maven.publish` | 0.36.0 | 0.37.0 |
| `com.github.gmazzo.buildconfig` | 6.0.9 | 6.0.10 |
| `io.mockk:mockk` (test-only) | 1.14.9 | 1.14.11 |

All five re-verified as current latest stable against `repo1.maven.org` / `services.gradle.org` at bump time, rather than trusting the survey in #156 — which was 9 days old. `buildconfig` was listed there as unconfirmed because the metadata path 404'd; it resolves now, so it is included.

One verification note: the Gradle **Plugin Portal** marker metadata for `com.vanniktech.maven.publish` reports `<latest>0.13.0`, which is wrong/stale. The real coordinate `com.vanniktech:gradle-maven-publish-plugin` on Central correctly reports 0.37.0. Anyone re-checking this should use Central, not the portal marker.

## Verification

- `compileKotlinJvm`, `compileKotlinLinuxX64`, `compileKotlinLinuxArm64` — green
- `ktlintCheck`, `apiCheck` — green, **no `api/` diff**
- `jvmTest` 320/320, `linuxX64Test` 288/288, `cross_test:jvmTest` 64/64, `cross_test:linuxX64Test` 62/62 — 0 failures, under `dbus-run-session`
- Publish path exercised (`publishToMavenLocal` against a SNAPSHOT version, since vanniktech touches it): root, `-jvm`, `-linuxx64`, `-linuxarm64` all publish cleanly under 0.37.0

## Two pre-existing problems surfaced while verifying

Both are filed separately rather than fixed here — neither is caused by this bump, and folding either in would muddy a diff whose value is being obviously inert.

**1. `:codegen` signs unconditionally, breaking the documented SNAPSHOT workflow.** The root build guards signing behind a non-SNAPSHOT check (`build.gradle.kts:287`), with a comment explaining that SNAPSHOT builds skip signing so local cross-repo validation doesn't need a GPG key. `codegen/build.gradle.kts:107` calls `signAllPublications()` with no such guard, so `publishToMavenLocal` fails on `:codegen:signMavenPublication` with `gpg: no default secret key` even on a SNAPSHOT. That contradicts the workflow CLAUDE.md documents. **Reproduced on unmodified `main` with the OLD toolchain**, so it predates this PR — which is also why the smoke test above covers 4 of 5 coordinates rather than all 5. Excluding the signing task doesn't work around it (the publication then wants `.asc` files that don't exist).

**2. Gradle 9.6.1 newly deprecates syntax this repo's build scripts use.** `val name by getting { }` for source sets is deprecated with removal in Gradle 10 — roughly seven occurrences across `build.gradle.kts` and `compile_test/build.gradle.kts` — plus one deprecated `Configuration` API at `build.gradle.kts:171`. The build succeeds; these are warnings only, and they were not present on 9.2.1. Worth handling as its own Gradle-10-readiness pass.

## Also noticed, not changed

The two sample builds carry their own wrappers still on Gradle **9.0.0** (`samples/bluez-scan`, `samples/demo-service`). They're standalone builds against the published artifact and outside this issue's scope, but they're now two minor versions behind the root build.

Fixes #156


monkopedia-reviewer (approved):
Tier 1 (infra, default). Independently re-verified rather than trusting the PR's own claims:

**Versions** — checked each against the authoritative source, not the PR's table:
- Gradle 9.6.1: `services.gradle.org/versions/current` confirms it's current stable, and its published `wrapperChecksum` (`497c8c2a…`) matches `sha256sum` of the committed `gradle-wrapper.jar` byte-for-byte — the jar is a genuine regeneration, not hand-edited.
- `org.jlleitschuh.gradle.ktlint` 14.2.0, `com.github.gmazzo.buildconfig` 6.0.10, `io.mockk:mockk` 1.14.11 — all `<latest>`/`<release>` on their respective metadata, no RC/beta ahead of them.
- `com.vanniktech.maven.publish` 0.37.0 — confirmed current on Central (`repo1.maven.org` metadata: latest=0.37.0). Also confirmed the PR's side-claim: the Gradle **Plugin Portal** marker for this coordinate does report a stale `<latest>0.13.0`/`<release>0.13.0` while 0.37.0 sits further down the portal's own `<versions>` list — Central is the correct source to check, portal marker is unreliable for this one. Good catch, worth remembering for future bumps.

**Pre-existing issues correctly identified as pre-existing, not introduced here**: this diff touches only `gradle/libs.versions.toml`, the wrapper jar/properties, and `gradlew`/`gradlew.bat` — `build.gradle.kts` and `codegen/build.gradle.kts` are untouched by the diff, so the codegen unconditional `signAllPublications()` (line 107, no SNAPSHOT guard, unlike root's line 288) and the `by getting` deprecation warnings (5 in root `build.gradle.kts` + 2 in `compile_test/build.gradle.kts` = 7, matches the count claimed) necessarily predate this change. Deferring both rather than folding them in is the right call — keeps this diff mechanically inert as advertised.

**No ABI surface touched**: `git diff main...HEAD --stat -- api/` is empty.

**CI**: watched `gh pr checks 162 --watch` to completion — `static-checks`, `full-tests-x64`, `native-build-x64`, `arm-runtime-tests`, and the full `pr-smoke` matrix all `pass`; `cleaner-soak`/`nightly-stress` `skipping` as expected on a PR. Not relying on the PR's stated local test counts.

Approving and merging.

PR #165, opened 2026-07-30
chore(codegen): bump xmlutil 0.91.3 -> 1.0.1

### LINKED ISSUE #157: Bump xmlutil 0.91.3 → 1.0.1 (major) in :codegen XML parsing
**Category:** dep-bump

**Where:** `gradle/libs.versions.toml` — `pdvrieze-xmlutil` (`io.github.pdvrieze.xmlutil:serialization`), current `0.91.3` → latest stable **`1.0.1`** (MAJOR).

**Current:** the `:codegen` module parses D-Bus introspection XML with pdvrieze xmlutil 0.91.3. xmlutil has since shipped its 1.0 line (1.0.0, 1.0.1).

**Proposed:** bump to `1.0.1`, working through whatever API changes the 0.91 → 1.0 major introduces (namespace/serialization API adjustments are the usual surface). Filed as its **own** ticket, separate from the routine build-tooling refresh, so the migration is judged on its own merits.

**Why:** it's a major version behind on the one library that does the codegen's XML parsing. Getting onto the 1.0 line before it drifts further keeps the eventual migration small. **No klib-ABI concern:** xmlutil is a build-time dependency of `:codegen` only — it does not ship in the published runtime library — so there's no downstream (blue-falcon) ABI sequencing to worry about here.

**Effort:** small refactor (major bump — verify `:codegen` still generates the checked-in BlueZ proxy/adaptor fixtures byte-for-byte, and run `apiDump`/generation tests).

monkopedia-coder:
Major-version bump of the XML parser behind `:codegen`: `io.github.pdvrieze.xmlutil:serialization` 0.91.3 → 1.0.1. One line in `gradle/libs.versions.toml`; **no source changes required**.

## Why a major bump needs no migration here

The APIs this repo uses all kept compatible signatures across the boundary:

- `XML { … }` builder with `isUnchecked`
- subclassing `DefaultXmlSerializationPolicy` and overriding `handleUnknownContentRecovering`
- `XmlDescriptor`, `InputKind`, `XML.ParsedData`, `QName`, `XmlReader`
- the `@XmlSerialName` / `@XmlElement` / `@XmlValue` annotations in `XmlFormat.kt`

The 1.0 line's breaking changes are concentrated in the **DOM API** (nullable returns, `NamedNodeList` reshaping) and in **module restructuring** (`core-compat` is no longer published). This repo touches neither — it depends on the `serialization` artifact only, and resolution confirms a clean 1.0.1 graph:

```
io.github.pdvrieze.xmlutil:serialization:1.0.1
  \--- serialization-jvm:1.0.1
        +--- core:1.0.1 -> core-jvm:1.0.1
```

xmlutil is a build-time dependency of `:codegen` and does not ship in the published runtime library, so there is no klib-ABI or downstream (blue-falcon-sdbus) impact. `apiCheck` green, no file under `api/` touched.

## Generated-output verification

This is the actual risk in this bump — a green `apiCheck` says nothing about what the generator *emits*, and the checked-in BlueZ fixtures are the contract.

**1. `:codegen:test` — 77/77 green, 0 failures.** `GeneratorTest.assertFiles` compares generated output against the checked-in fixtures with `assertEquals` on file contents. Its `WRITE_FILES` regeneration switch is `false` (`GeneratorTest.kt:108`), so the comparison is a real assertion rather than a self-fulfilling rewrite — and the working tree stayed clean across the run, confirming no fixture was silently rewritten.

**2. The real CLI, end-to-end, over every fixture.** Built the `:codegen` distribution and ran the actual `Xml2Kotlin` binary against all 20 fixture XML inputs with `-a -p`, then compared every emitted file to the checked-in fixtures by SHA-256:

```
fixture dirs: 20   fixture .kt files: 66   CLI-generated .kt files: 66
RESULT: every checked-in fixture is reproduced byte-for-byte under xmlutil 1.0.1
```

That second pass is not redundant. The test suite **builds its own copy of the XML config** instead of sharing the one the CLI uses, so a passing suite alone would not prove the shipped code path is unaffected — see below.

## Observation: the XML config is triplicated

The same `XML { isUnchecked = true; policy = object : DefaultXmlSerializationPolicy(policy) { … } }` block exists in three places:

| Location | Used by |
| --- | --- |
| `codegen/src/main/kotlin/Xml2Kotlin.kt:113` | the real CLI — the shipped path |
| `codegen/src/test/kotlin/TestXmlSupport.kt:13` | `CodegenParsingEdgeCaseTest` |
| `codegen/src/test/kotlin/GeneratorTest.kt:40` | the golden-fixture tests |

They are semantically identical today, so nothing is wrong right now. But the golden-fixture tests parse with their **own** parser configuration, not the one `Xml2Kotlin` ships — so a future change to the CLI's policy (or a library change in how that policy behaves) could regress generated output while the suite stays green. Filed separately rather than fixed here; flagging it because it is precisely why this PR's verification needed the extra CLI pass.

## Verification summary

- `:codegen:build`, `:plugin:build` — green
- `ktlintCheck`, `apiCheck` — green, no `api/` diff
- `:codegen:test` 77/77, 0 failures, fixtures unmodified
- Real CLI reproduces all 66 fixture files byte-for-byte

Fixes #157


monkopedia-reviewer (approved):
Independently reproduced the PR's central claim rather than accepting it: rebuilt `:codegen` clean, ran the real CLI binary over all 20 fixture XML inputs with `-a -p`, and compared per fixture directory by SHA-256 — all 66 checked-in `.kt` fixtures matched byte-for-byte, 0 mismatches.

- `:codegen:test` 77/77 green, 0 failures. `WRITE_FILES = false` confirmed at `GeneratorTest.kt:108` and the working tree was clean after the run, so the golden comparison is a real assertion rather than a self-fulfilling rewrite.
- `apiCheck` green at root, `:codegen` and `:plugin`; no file under `api/` touched by the diff.
- All required CI checks green (`static-checks`, `full-tests-x64`, `native-build-x64`, `arm-runtime-tests`, full `pr-smoke` matrix); `cleaner-soak`/`nightly-stress` correctly skipping on a PR.
- xmlutil's 1.0 breaking changes are in the DOM API and module restructuring (`core-compat` no longer published), neither of which this repo touches — it depends on the `serialization` artifact only.

Tier 2 (`deps_major`). The residual question raised for owner judgment was that `codegen/build.gradle.kts:20` declares `api(libs.xmlutil)` and `plugin/build.gradle.kts:25` declares `api(project(":codegen"))`, so the major bump does propagate onto the compile classpath of anyone consuming the published `sdbus-kotlin-codegen` artifact or the `com.monkopedia.sdbus.plugin` Gradle plugin — broader than "internal to `:codegen`" as the PR framed it. `codegen/api/codegen.api` has zero xmlutil-typed public signatures, so there is no direct ABI leak; the exposure is that a downstream project pinning an older xmlutil could see a resolution bump forced on it. That scoping decision is pre-existing and not introduced here.

Jason reviewed and approved the residual, with the `api` vs `implementation` scoping question to be tracked as its own issue.

One note for the record, not blocking and not mentioned in the PR body: xmlutil 1.0.1 newly emits a deprecation warning for the `constructor(serializersModule, configure): XML` builder form this repo uses. Non-breaking, and consistent with the PR's "signatures stayed compatible" claim, but it is a small piece of follow-up surface.

Issue #167, opened 2026-07-30
xmlutil is api-scoped in :codegen despite leaking no types — forces resolution on plugin consumers


monkopedia-coder:
`:codegen` exposes xmlutil to its consumers as an `api` dependency, even though no xmlutil type appears anywhere in its public signatures. That makes the library a transitive part of the compile classpath for anyone consuming the published `sdbus-kotlin-codegen` artifact or the `com.monkopedia.sdbus.plugin` Gradle plugin.

## Current

`codegen/build.gradle.kts:20`:

```kotlin
api(libs.xmlutil)
```

`plugin/build.gradle.kts:25` re-exposes it in turn:

```kotlin
api(project(":codegen"))
```

So the chain is: consumer → plugin → codegen → xmlutil, all `api`-scoped.

## Why it looks unintended

`codegen/api/codegen.api` contains **zero** xmlutil-typed public signatures. Nothing in `:codegen`'s published API returns, accepts, or extends an xmlutil type — the library is used entirely inside `Xml2Kotlin` and `XmlFormat` to parse introspection XML, which is an implementation concern. `api` scope is for dependencies whose types leak into your own API; that is not the case here.

## Impact

Low, but real: because the dependency is `api`-scoped, a downstream project that pins its own (older) xmlutil can have a resolution bump forced on it by Gradle's conflict resolution when it depends on the codegen artifact or the Gradle plugin. With `implementation` scope, xmlutil would stay on the runtime classpath where it is needed without participating in consumers' compile-classpath resolution.

This surfaced concretely during #165 (xmlutil 0.91.3 → 1.0.1, a major bump). That PR framed xmlutil as a build-time detail internal to `:codegen`; the review established that the `api` scoping makes it broader than that. The bump was approved on the merits — no ABI leak, and the 1.0 breaking changes are in the DOM API and module restructuring that no consumer of this tool would touch — with the scoping question split out here so the call is explicit rather than assumed.

## Proposed

Change `api(libs.xmlutil)` to `implementation(libs.xmlutil)` in `codegen/build.gradle.kts`, then confirm `:codegen:build`, `:plugin:build` and `apiCheck` stay green and that nothing in `:plugin` was quietly relying on the transitive xmlutil types.

Note this is itself a (small) breaking change for any consumer that *was* relying on the transitive availability, so it is worth a CHANGELOG line rather than a silent flip.

## Related, small

[...]

PR #169, opened 2026-07-31
refactor(codegen): parse introspection XML through one shared config

### LINKED ISSUE #166: Golden-fixture tests parse with their own xmlutil config, not the CLI's (config triplicated)
The xmlutil parser configuration used by `:codegen` exists in **three separate copies**. They are semantically identical today, so nothing is broken right now — but the golden-fixture tests parse with their *own* copy rather than the one the CLI ships, which means the suite can stay green through a regression in the shipped parsing path.

## The three copies

| Location | Used by |
| --- | --- |
| `codegen/src/main/kotlin/Xml2Kotlin.kt:113` | the real `Xml2Kotlin` CLI — **the shipped path** |
| `codegen/src/test/kotlin/TestXmlSupport.kt:13` | `CodegenParsingEdgeCaseTest` |
| `codegen/src/test/kotlin/GeneratorTest.kt:40` | the golden-fixture tests (`testInterface` / `testAdaptor` / `testProxy`) |

Each is the same block:

```kotlin
XML {
    isUnchecked = true
    policy = object : DefaultXmlSerializationPolicy(policy) {
        override fun handleUnknownContentRecovering(
            input: XmlReader,
            inputKind: InputKind,
            descriptor: XmlDescriptor,
            name: QName?,
            candidates: Collection<Any>
        ): List<ParsedData<*>> = emptyList()
    }
}
```

## Why it matters

`GeneratorTest` is the contract test for generator output — it asserts the generated Kotlin matches the 66 checked-in BlueZ fixture files byte-for-byte. But it feeds those fixtures through a parser it configured itself. So the assertion is "the generators produce the expected output *given this test's parsing*", not "the tool produces the expected output".

Concretely, a change to the CLI's `handleUnknownContentRecovering`, to `isUnchecked`, or to the policy object — or a library-side behavior change in how that policy is applied — could alter what the shipped tool emits while every test stays green. The suite would report success about a code path it never executed.

This is the "green signal that measures nothing" shape: the test looks like end-to-end coverage of the generator and is actually coverage of a parallel implementation.

## Proposed

Extract the config once — an internal `val` or factory in `:codegen`'s main source set — and have `Xml2Kotlin`, `TestXmlSupport`, and `GeneratorTest` all use that single instance. The tests then exercise the real parsing configuration, and the duplication stops being a drift hazard.

Optionally, add a test that runs the CLI end-to-end (build the distribution, generate into a temp dir, compare to fixtures by content) so the shipped entry point itself is covered rather than only its constituent parts.

## Provenance

Found while verifying #165 (xmlutil 0.91.3 → 1.0.1). Because of this gap, the passing test suite was not sufficient evidence that a major parser bump left generated output unchanged, so that PR additionally ran the real CLI over all 20 fixture inputs and compared 66 emitted files by hash. That extra pass should not be necessary — it should be what the suite already does.

monkopedia-coder:
Fixes #166

## What

The xmlutil parser configuration existed in three identical copies:

| Location | Used by |
| --- | --- |
| `codegen/src/main/kotlin/Xml2Kotlin.kt:114` | the shipped `Xml2Kotlin` CLI |
| `codegen/src/test/kotlin/TestXmlSupport.kt:13` | `CodegenParsingEdgeCaseTest`, `CodegenNamingAndTypeTest`, `CompilationIntegrationTest` |
| `codegen/src/test/kotlin/GeneratorTest.kt:40` | the golden-fixture tests |

So `GeneratorTest` — the contract test that asserts generated Kotlin matches the 66 checked-in BlueZ fixtures byte-for-byte — fed those fixtures through a parser it configured itself. It looked like end-to-end coverage of the generator and was actually coverage of a parallel configuration; a regression in the CLI's parsing path could not have failed it.

Extracted once as `internal val introspectionXml` in `:codegen`'s **main** source set (`XmlFormat.kt`, next to the model it parses into), and pointed all three call sites at it. `internal` keeps it off the public surface — `apiCheck` passes with `codegen/api/codegen.api` unchanged.

Net: **-28 lines** (65 deleted, 37 added). All three call sites collapse from a 12-line `XML { … }` block to a single identifier; `TestXmlSupport.kt` goes from 27 lines to 7.

## The xmlutil 1.0.1 deprecation (from #167's "related, small")

Handled at the now-single site, but **not** by switching builder form — that turns out not to be possible without changing behaviour. What I found in xmlutil 1.0.1:

- `XML { … }` resolves to `constructor(SerializersModule, XmlConfig.CompatBuilder.() -> Unit)`, deprecated with *"Update to the named factory functions"*.
- The named factory `XML.compat { … }` is **also** deprecated: *"Consider using the 1.0 object, but note it has changed defaults"*. So are `XML.Compat` and `XML.V1`.
- The only non-deprecated factories are `XML.recommended_1_0` / `XML.fast_1_0`, which build from `XmlConfig.DefaultBuilder` instead of `CompatBuilder` — different parsing defaults, and our policy is derived from the builder's default policy (`DefaultXmlSerializationPolicy(policy)`), so the change would propagate into decoding.
- Constructing the compat config directly is not open to callers either: `XmlConfig.CompatBuilder`'s constructor is `internal`, and `Compat.invoke` routes through `recommendedBuilder()` rather than a plain `CompatBuilder()`.

Verified against the bytecode that the deprecated `XML { … }` constructor and `XML.compat { … }` are instruction-for-instruction identical, so the site now calls `XML.compat` with a documented `@Suppress("DEPRECATION")` explaining why the pre-1.0 defaults stay pinned. **The build is warning-free for `:codegen` again** (only a pre-existing, unrelated `No cast needed.` warning in `CodegenNamingAndTypeTest.kt` remains).

Migrating to the 1.0 parsing defaults is a real behaviour change — 20 fixture inputs are not sufficient evidence for a parser defaults switch, which is precisely the class of silent drift this issue is about. Worth its own issue if wanted.

## Verification

- `./gradlew :codegen:test` — **77 tests, 0 failures**, including the 60 golden-fixture cases (20 fixture dirs x interface/adaptor/proxy).
- **`git diff` shows zero changes under `codegen/src/test/resources/`** — all 66 fixture `.kt` files are byte-identical, which is the proof the extracted config is behaviourally the same as all three copies it replaces.
- **`WRITE_FILES` is still `false`** in `GeneratorTest.kt` (the golden-file rewrite switch), so the assertions are real rather than vacuous.
- `./gradlew ktlintCheck apiCheck` — pass, no API dump change.
- `./gradlew :codegen:build --rerun-tasks` — pass.

Scope: `codegen/src/` only. No build script was touched.

## Deferred (explicitly out of scope)

The issue's optional second half — an end-to-end test that builds the distribution, runs the real CLI into a temp dir and compares to the fixtures by content — was deferred at triage as a materially larger piece of work with its own design questions. Not built here; it remains open.

### DIFF
--- a/codegen/src/main/kotlin/Xml2Kotlin.kt
+++ b/codegen/src/main/kotlin/Xml2Kotlin.kt
@@ -80,14 +80,6 @@ import com.github.ajalt.clikt.parameters.types.file
 import java.io.File
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.decodeFromString
-import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi
-import nl.adaptivity.xmlutil.QName
-import nl.adaptivity.xmlutil.XmlReader
-import nl.adaptivity.xmlutil.serialization.DefaultXmlSerializationPolicy
-import nl.adaptivity.xmlutil.serialization.InputKind
-import nl.adaptivity.xmlutil.serialization.XML
-import nl.adaptivity.xmlutil.serialization.XML.ParsedData
-import nl.adaptivity.xmlutil.serialization.structure.XmlDescriptor
 
 fun main(args: Array<String>) = Xml2Kotlin().main(args)
 
@@ -108,21 +100,9 @@ class Xml2Kotlin : CliktCommand() {
         help = "Override Kotlin package for generated files"
     )
 
-    @OptIn(ExperimentalXmlUtilApi::class)
     override fun run() {
         val packageOverride = outputPackage?.trim()?.takeUnless { it.isEmpty() }
-        val xml = XML {
-            isUnchecked = true
-            policy = object : DefaultXmlSerializationPolicy(policy) {
-                override fun handleUnknownContentRecovering(
-                    input: XmlReader,
-                    inputKind: InputKind,
-                    descriptor: XmlDescriptor,
-                    name: QName?,
-                    candidates: Collection<Any>
-                ): List<ParsedData<*>> = emptyList()
-            }
-        }.decodeFromString<XmlRootNode>(input.readText())
+        val xml = introspectionXml.decodeFromString<XmlRootNode>(input.readText())
         if (!keep) {
             output.deleteRecursively()
         }
--- a/codegen/src/main/kotlin/XmlFormat.kt
+++ b/codegen/src/main/kotlin/XmlFormat.kt
@@ -24,9 +24,41 @@ package com.monkopedia.sdbus
 
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi
+import nl.adaptivity.xmlutil.QName
+import nl.adaptivity.xmlutil.XmlReader
+import nl.adaptivity.xmlutil.serialization.DefaultXmlSerializationPolicy
+import nl.adaptivity.xmlutil.serialization.InputKind
+import nl.adaptivity.xmlutil.serialization.XML
+import nl.adaptivity.xmlutil.serialization.XML.ParsedData
 import nl.adaptivity.xmlutil.serialization.XmlElement
 import nl.adaptivity.xmlutil.serialization.XmlSerialName
 import nl.adaptivity.xmlutil.serialization.XmlValue
+import nl.adaptivity.xmlutil.serialization.structure.XmlDescriptor
+
+/**
+ * The single parser for D-Bus introspection XML, shared by the [Xml2Kotlin] CLI and the tests so
+ * the golden fixtures assert against the configuration the tool actually ships. Unknown content
+ * is dropped rather than failing, so vendor extensions don't break generation.
+ *
+ * `XML.compat` is deprecated in xmlutil 1.0, but its replacements deliberately change the parsing
+ * defaults and xmlutil offers no non-deprecated route back to the old ones. Switching is a
+ * behaviour change needing its own review, so the compat defaults stay pinned here.
+ */
+@OptIn(ExperimentalXmlUtilApi::class)
+@Suppress("DEPRECATION")
+internal val introspectionXml: XML = XML.compat {
+    isUnchecked = true
+    policy = object : DefaultXmlSerializationPolicy(policy) {
+        override fun handleUnknownContentRecovering(
+            input: XmlReader,
+            inputKind: InputKind,
+            descriptor: XmlDescriptor,
+            name: QName?,
+            candidates: Collection<Any>
+        ): List<ParsedData<*>> = emptyList()
+    }
+}
 
 @Serializable
 @XmlSerialName("node")
--- a/codegen/src/test/kotlin/GeneratorTest.kt
+++ b/codegen/src/test/kotlin/GeneratorTest.kt
@@ -25,36 +25,16 @@ package com.monkopedia.sdbus
 import com.squareup.kotlinpoet.FileSpec
 import java.io.File
 import kotlinx.serialization.decodeFromString
-import nl.adaptivity.xmlutil.QName
-import nl.adaptivity.xmlutil.XmlReader
-import nl.adaptivity.xmlutil.serialization.DefaultXmlSerializationPolicy
-import nl.adaptivity.xmlutil.serialization.InputKind
-import nl.adaptivity.xmlutil.serialization.XML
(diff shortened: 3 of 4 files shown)


monkopedia-reviewer (approved):
Independent review, fresh checkout at `de4ad4c`.

**Deduplication is real and complete.** Grepped the whole `codegen/src/` tree for `XML {` and any local xmlutil-serialization `XML` usage: none remain outside `XmlFormat.kt`. All three original call sites (`Xml2Kotlin.kt`, `TestXmlSupport.kt`, `GeneratorTest.kt`) now reference the same `internal val introspectionXml` in `XmlFormat.kt`, confirmed by reading each file's post-diff content, not just the diff hunks.

**The `@Suppress("DEPRECATION")` claims — verified against the actual xmlutil 1.0.1 jar, not taken on trust:**
- Decompiled `serialization-jvm-1.0.1.jar`: the `XML(SerializersModule, Function1<CompatBuilder,Unit>)` ctor (`XML { }`) carries `kotlin.Deprecated("Update to the named factory functions")`; `XML.compat` carries `kotlin.Deprecated("Consider using the 1.0 object, but note it has changed defaults")`; `getCompat()`/`getV1$annotations()` both carry the classfile `Deprecated` attribute. All four claimed-deprecated symbols are in fact deprecated.
- `javap -c` on both the `XML{}` ctor and `XML.compat()` bodies: instruction-for-instruction identical (same `new CompatBuilder(false,null,null,null,15)` → `configure.invoke` → `new XmlConfig` → `XML.<init>` sequence, differing only in constant-pool indices). The "verified against the bytecode" claim is correct.
- `recommended_1_0`/`fast_1_0` take `XmlConfig.DefaultBuilder`, whose `policy` field is concretely typed `DefaultXmlSerializationPolicy`, vs. `CompatBuilder`'s interface-typed `XmlSerializationPolicy` — confirming these build from a genuinely different ambient default, not just a different builder class name.
- Empirically confirmed `XmlConfig.CompatBuilder`'s constructor is closed to callers: I added a scratch file calling it directly from `:codegen` and `compileTestKotlin` failed with `Cannot access 'constructor(...)': it is internal in 'nl.adaptivity.xmlutil.serialization.XmlConfig.CompatBuilder'` (removed the scratch file after).

So the suppression is justified, not just plausible, and the inline KDoc on `introspectionXml` documents the reasoning adequately. Agree deferring the 1.0-defaults migration to its own issue is correct — 20 fixtures isn't enough evidence for a parser-defaults change, which is exactly the drift class #166 is about.

**Byte-identity, re-verified independently (not from the PR's own report):**
- `git diff main...HEAD -- codegen/src/test/resources/` → 0 lines. No fixture file changed in this diff at all, so the "66 files byte-identical" claim holds trivially and isn't a stale-baseline artifact — confirmed the diff is empty against a fresh clone, not a dirty local tree.
- `WRITE_FILES` re-checked directly at `GeneratorTest.kt:88` → `private const val WRITE_FILES = false`. Not a vacuous pass.
- Ran `:codegen:test` myself: 77 tests, 0 failures (7+1+60+4+5 across the 5 suites, matching the PR's count including the 60 golden-fixture cases).
- Ran `ktlintCheck apiCheck` myself: BUILD SUCCESSFUL, and `git diff` on `codegen/api/` and `api/` is empty — `apiCheck` genuinely saw no surface change, consistent with `introspectionXml` being `internal`.

**Refactor-must-shrink bar:** all three call sites collapsed from 12–21 line inline `XML { }` blocks to a single identifier reference; `TestXmlSupport.kt` went from 27 to 7 lines. The `+12` in `codegen/src/main` is because the one canonical instance has to physically live somewhere in main (tests can't be the source of truth for what the CLI ships), and `XmlFormat.kt` — the file already holding `XmlRootNode`, the type it decodes into — is the right home; it's a single top-level `val`, not a new class/interface/generic layer. Net -28 lines project-wide. This is real deduplication, not indirection for its own sake.

**Remote CI:** watched `gh pr checks 169 --watch` to completion — `static-checks`, `full-tests-x64`, `native-build-x64`, `arm-runtime-tests`, and all `pr-smoke` variants pass; `cleaner-soak`/`nightly-stress` show `skipping` (expected, nightly-only). `mergeStateStatus` re-checked as `CLEAN` right before merge, head SHA re-confirmed unchanged via `git ls-remote`.

**Category/tier:** `refactor`, no public-API or behavior change (apiCheck + fixture byte-identity both confirm) → tier 1 per default policy; `Monkopedia/sdbus-kotlin` carries no overrides. Nothing surfaced in review that warrants bumping the tier — this is a clean, well-evidenced internal test-quality fix.

Approving and merging.

Issue #163, opened 2026-07-29
publishToMavenLocal on a SNAPSHOT still requires GPG because :codegen signs unconditionally


monkopedia-coder:
`publishToMavenLocal` cannot complete on a SNAPSHOT version without a GPG secret key, because `:codegen` signs unconditionally. This contradicts the local cross-repo testing workflow that `CLAUDE.md` documents ("SNAPSHOT versions skip signing (for local `publishToMavenLocal` cross-repo testing); real releases are signed").

## The asymmetry

The root build guards signing, with a comment stating the intent explicitly — `build.gradle.kts:284-289`:

```kotlin
// SNAPSHOT builds (e.g. publishToMavenLocal for downstream testing) are not signed:
// Maven Central doesn't require signatures for snapshots, and it avoids needing a GPG
// key for local cross-project validation.
if (!version.toString().endsWith("SNAPSHOT")) {
    signAllPublications()
}
```

`codegen/build.gradle.kts:105-113` has no such guard:

```kotlin
publishToMavenCentral(automaticRelease = true)

signAllPublications()
}

signing {
    useGpgCmd()
    sign(publishing.publications)
}
```

## Repro

On unmodified `main` (5c90d62), no GPG secret key present:

```
$ ./gradlew :codegen:publishToMavenLocal -Pversion=1.0.2-SNAPSHOT
> Execution failed for task ':codegen:signMavenPublication'.
  > Process 'command 'gpg'' finished with non-zero exit value 2
    gpg: no default secret key: No secret key
```

The other four coordinates (root, `-jvm`, `-linuxx64`, `-linuxarm64`) publish cleanly on the same invocation — only `:codegen` fails.

Excluding the signing task is not a workaround: the publication then fails validation looking for the `.asc` files that task would have produced.

```
> Invalid publication 'maven': artifact file does not exist: '.../codegen/build/publications/maven/module.json.asc'
```

## Impact

Anyone without the release GPG key — which is everyone except the release machine — cannot run a full `publishToMavenLocal`. That is the documented way to validate a change against a downstream consumer (blue-falcon-sdbus) before cutting a release, so the gap is squarely on the path it was meant to serve. It bites silently: four of five coordinates land, so a partial local repo can look successful.

## Proposed

Apply the same guard in `codegen/build.gradle.kts` that the root build uses, so SNAPSHOT builds skip signing there too. Real releases are unaffected — they are non-SNAPSHOT and still sign all five coordinates.

[...]

PR #168, opened 2026-07-31
fix(build): skip signing on SNAPSHOT in :codegen and :plugin

monkopedia-coder:
Fixes #163

`publishToMavenLocal` on a SNAPSHOT required a GPG secret key. The root build already guards signing behind a non-SNAPSHOT check, and `CLAUDE.md` documents the policy ("SNAPSHOT versions skip signing (for local `publishToMavenLocal` cross-repo testing); real releases are signed"). This applies that same guard to the modules that diverged from it.

## Two findings that changed the fix from what the issue proposed

**1. Guarding `signAllPublications()` alone does NOT fix it.** The issue asked to confirm this rather than assume it, so I measured it: with only `signAllPublications()` guarded, `:codegen` still fails, because the standalone `signing { useGpgCmd(); sign(publishing.publications) }` block re-registers the task by itself.

```
$ ./gradlew :codegen:publishToMavenLocal -Pversion=1.0.2-SNAPSHOT      # signAllPublications() guarded, signing {} block left alone
> Task :codegen:signMavenPublication FAILED
gpg: no default secret key: No secret key
* What went wrong:
Execution failed for task ':codegen:signMavenPublication' (registered in build file 'codegen/build.gradle.kts').
```

So both spots need the guard — which is exactly what the root build does (`build.gradle.kts:287` and `:353`).

**2. `:plugin` has the identical unguarded pattern, so fixing only `:codegen` would not have fixed the issue's stated impact** ("cannot run a full `publishToMavenLocal`"). With `:codegen` fixed, the full run just moved its failure one module over:

```
$ ./gradlew publishToMavenLocal -Pversion=1.0.2-SNAPSHOT              # :codegen fixed, :plugin untouched
* What went wrong:
Execution failed for task ':plugin:signPluginMavenPublication' (registered in build file 'plugin/build.gradle.kts').
> Process 'command 'gpg'' finished with non-zero exit value 2
```

`:plugin` also needs one step the other modules don't: `com.gradle.plugin-publish` registers Sign tasks for its own publications independently of our config, so with the guard alone they fail differently rather than disappearing —

```
* What went wrong:
Execution failed for task ':plugin:signPluginMavenPublication' (registered by plugin 'com.gradle.plugin-publish').
> Cannot perform signing task ':plugin:signPluginMavenPublication' because it has no configured signatory
```

[...]

PR #170, opened 2026-07-31
build: Gradle 10 readiness — migrate deprecated Kotlin DSL delegate syntax and drop eager tasks.all

### LINKED ISSUE #164: Gradle 10 readiness: migrate deprecated 'by getting' source-set syntax and Configuration API
Gradle 9.6.1 (landing via #162) newly deprecates build-script syntax this repo uses. The build succeeds and only warns, but the deprecated forms are **scheduled for removal in Gradle 10**, so they become hard failures at the next major.

```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.
```

These warnings were not emitted on Gradle 9.2.1.

## What's deprecated

**1. `val name by getting { }` source-set delegate syntax** — roughly seven occurrences:

| File | Lines |
| --- | --- |
| `build.gradle.kts` | 94, 103, 114, 121, 128 |
| `compile_test/build.gradle.kts` | 28, 37 |

```
The 'val name by getting { }' property delegate syntax has been deprecated.
This is scheduled to be removed in Gradle 10.
Use 'val element = getByName(name) { }' instead.
```

Upgrade note: https://docs.gradle.org/9.6.1/userguide/upgrading_version_9.html#kotlin_dsl_delegated_properties

**2. `Configuration.getTaskDependencyFromProjectDependency(boolean, String)`** — one occurrence, `build.gradle.kts:171`:

```
The Configuration.getTaskDependencyFromProjectDependency(boolean, String) method
has been deprecated. This is scheduled to be removed in Gradle 10.
```

Upgrade note: https://docs.gradle.org/9.6.1/userguide/upgrading_version_9.html#deprecate_getTaskDependencyFromProjectDependency

Reproduce the full list with:

```
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew ktlintCheck --warning-mode all
```

## Proposed

Mechanical migration of the source-set declarations to `getByName(...)`, plus a look at line 171 to find the non-deprecated equivalent for whatever that dependency wiring is doing. Target is a build that runs clean under `--warning-mode all`, so the next real deprecation isn't lost in existing noise.

The `getByName` half is genuinely mechanical. The `Configuration` API call needs a moment's thought about intent, so it may be worth doing as a second commit.

## Why now rather than at Gradle 10

Cheap while it's five call sites and a warning; expensive when it's a hard failure blocking an urgent toolchain bump. This is the same "avoid painful multi-version catch-up later" reasoning as #156, applied to build-script API rather than versions.

## Provenance

Surfaced by the Gradle 9.2.1 → 9.6.1 bump in #162, and deliberately left out of that PR to keep an inert version-bump diff inert.

monkopedia-coder:
Fixes #164

Two commits, as the issue suggested: the mechanical delegate-syntax migration, then the `Configuration` API one.

## Before / after

`JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew ktlintCheck --warning-mode all`

**Before** (`main` @ `ba62857`) — **15** deprecation warnings:

| File | Lines | Deprecation |
| --- | --- | --- |
| `build.gradle.kts` | 94, 103, 114, 121, 128 | `val name by getting { }` |
| `build.gradle.kts` | 171 | `Configuration.getTaskDependencyFromProjectDependency` |
| `compile_test/build.gradle.kts` | 28, 37 | `val name by getting { }` |
| `cross_test/build.gradle.kts` | 20, 25, 31 | `val name by getting { }` |
| `stress_test/build.gradle.kts` | 17, 22, 28 | `val name by getting { }` |
| `plugin/build.gradle.kts` | 49 | `val name by creating { }` |

**After** — **0**. Not just for `ktlintCheck`; deprecation count is 0 for every invocation I could think of that walks the task graph:

```
### gradlew help                             0 deprecations
### gradlew build --dry-run                  0 deprecations
### gradlew license --dry-run                0 deprecations
### gradlew licenseCheckForKotlin --dry-run  0 deprecations
### gradlew tasks --all                      0 deprecations
### gradlew publishToMavenLocal --dry-run    0 deprecations
```

Nothing was left out of scope — there are no other Gradle-10 deprecations in this build. `codegen` had no affected call sites, and no file under `codegen/` is touched (concurrent work on #163/#166/#167).

Note the issue's table listed only the root + `compile_test` sites (7). The identical warning is emitted from `cross_test`, `stress_test` and `plugin` too, so those are migrated here as well — otherwise the stated goal (a build clean under `--warning-mode all`) isn't met and the next real deprecation still gets lost in noise.

## Commit 1 — delegate syntax

`val name by getting { }` → `getByName("name") { }`, and `val sdbusPlugin by plugins.creating { }` → `plugins.create("sdbusPlugin") { }`.

The Gradle guide's suggested form is `val element = getByName(name) { }`, but none of the delegated `val`s were referenced anywhere, so they're dropped rather than kept as unused locals.

## Commit 2 — `Configuration.getTaskDependencyFromProjectDependency`

**This one is not our call.** `--stacktrace` puts it inside the Kotlin Gradle Plugin, not the build script:

```
DefaultConfiguration.getTaskDependencyFromProjectDependency(DefaultConfiguration.java:852)
ConfigureBuildSideEffectKt$addDependsOnTaskInOtherProjects$1.execute(ConfigureBuildSideEffect.kt:35)
  ... DefaultNamedDomainObjectCollection$AbstractDomainObjectCreatingProvider.tryCreate
  ... DefaultDomainObjectCollection.all(DefaultDomainObjectCollection.java:169)
Build_gradle$4.execute(build.gradle.kts:171)
```

KGP registers that side effect to run on task creation. `build.gradle.kts:171` is `tasks.all { }`, which eagerly realizes *every* task during configuration — including the one whose creation triggers KGP's deprecated call. So there is no API to swap on our side; what the build script contributes is forcing KGP's call to happen on every single invocation.

Confirmed by removing the eager enumerations: the warning disappears entirely, including from `build` and `tasks --all`. So this fix does genuinely remove it rather than defer it. (If a future KGP still calls the deprecated method from a path we do reach, that's a JetBrains-side fix and out of our hands — but today nothing in this build reaches it.)

Two eager `tasks.all { }` blocks are replaced:

1. **Native compiler/linker args** → `tasks.withType<KotlinNativeLink>().configureEach { }` + `tasks.withType<KotlinNativeCompile>().configureEach { }`. This also drops the `(this as? KotlinNativeLink)?` / `(this as? KotlinNativeCompile)?` casts and the surrounding `afterEvaluate`, which `configureEach` makes redundant. (Fixing only this one made the warning move to the second `tasks.all` — hence both.)
2. **`licenseCheckForKotlin` wiring** → `tasks.named("licenseCheckForKotlin") { dependsOn(tasks.matching { … }) }`. Same predicate, same direction; `matching` is a live view, so the container is walked only when that task is actually in the graph.

### Evidence the wiring is preserved

- `./gradlew licenseCheckForKotlin --dry-run` produces a **byte-identical 125-task graph** before and after (`diff` clean).
- Full `./gradlew :dependencies :compile_test:dependencies :cross_test:dependencies :stress_test:dependencies` report: **zero added lines**, and every removed line belongs to a `dokka*` configuration block. Excluding those blocks, the two reports differ only in 11 blank separator lines — every source-set / publication configuration is unchanged.
  The `dokka*` configurations are absent from the *report* only because they're now created when their tasks are realized rather than by our force-realizing everything; the publish graph still contains `dokkaGeneratePublicationHtml`, `jvmDokkaJavadocJar`, `kotlinMultiplatformDokkaJavadocJar`, `linuxX64DokkaJavadocJar`, `linuxArm64DokkaJavadocJar`, and `dokkaGeneratePublicationHtml` runs green.
- The `KotlinNativeLink` block is the arm64 cross-link wiring (`-Xoverride-konan-properties`, `-L`/`-rpath`, `--allow-shlib-undefined`), which only fails loudly if it stops being applied — `linkDebugSharedLinuxArm64` and the CI binaries `:linkDebugTestLinuxX64 :linkDebugTestLinuxArm64` all build green.

## Verification

- `./gradlew compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` — green
- `./gradlew ktlintCheck apiCheck` — green, and **`apiCheck` produced no diff**: `git status` shows no change under `api/` (only build scripts are modified in this PR)
- `./gradlew :linkDebugTestLinuxX64 :linkDebugTestLinuxArm64 :compile_test:build :cross_test:compileKotlinJvm :cross_test:compileTestKotlinLinuxX64 :stress_test:compileTestKotlinJvm :plugin:build` — green
- `./gradlew dokkaGeneratePublicationHtml` — green
- `./gradlew publishToMavenLocal --dry-run` — green

Not run locally (needs a session bus / hardware, left to CI): `jvmTest`, `linuxX64Test`, the dbusmock cross suites, and the BlueZ sample suites.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -91,7 +91,7 @@ kotlin {
     }
     applyDefaultHierarchyTemplate()
     sourceSets {
-        val commonMain by getting {
+        getByName("commonMain") {
             dependencies {
                 implementation(libs.kotlinx.coroutines)
                 // kotlinx-serialization types (KSerializer, the serializer companions) are part
@@ -100,7 +100,7 @@ kotlin {
                 implementation(libs.kotlinx.atomicfu)
             }
         }
-        val jvmMain by getting {
+        getByName("jvmMain") {
             dependencies {
                 implementation(libs.kotlinx.coroutines)
                 implementation(libs.kotlinx.serialization)
@@ -111,21 +111,21 @@ kotlin {
                 implementation(kotlin("stdlib"))
             }
         }
-        val jvmTest by getting {
+        getByName("jvmTest") {
             dependencies {
                 implementation(kotlin("test"))
                 implementation(libs.kotlinx.coroutines.test)
                 implementation(libs.mockk)
             }
         }
-        val nativeMain by getting {
+        getByName("nativeMain") {
             dependencies {
                 implementation(libs.kotlinx.coroutines)
                 implementation(libs.kotlinx.serialization)
                 implementation(kotlin("stdlib"))
             }
         }
-        val nativeTest by getting {
+        getByName("nativeTest") {
             dependencies {
                 implementation(libs.kotlinx.coroutines.test)
             }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -167,43 +167,47 @@ fun arm64RuntimeLibDirs(): List<String> = listOf(
     .filter { it.exists() }
     .map { it.absolutePath }
 
-afterEvaluate {
-    tasks.all {
-        val localSystemdLibDir = systemdLibDirForTask(name)
-        (this as? KotlinNativeLink)?.toolOptions {
-            freeCompilerArgs.addAll(overrides)
-            if (localSystemdLibDir != null) {
+// Configured lazily via withType/configureEach rather than an eager `tasks.all`: realizing every
+// task at configuration time also realizes the Kotlin plugin's `build` task, whose creation calls
+// the deprecated Configuration.getTaskDependencyFromProjectDependency (removed in Gradle 10).
+tasks.withType<KotlinNativeLink>().configureEach {
+    val localSystemdLibDir = systemdLibDirForTask(name)
+    toolOptions {
+        freeCompilerArgs.addAll(overrides)
+        if (localSystemdLibDir != null) {
+            freeCompilerArgs.addAll(
+                listOf(
+                    "-linker-option",
+                    "-L$localSystemdLibDir",
+                    "-linker-option",
+                    "-Wl,-rpath,$localSystemdLibDir"
+                )
+            )
+        }
+        if (name.contains("LinuxArm64", ignoreCase = true)) {
+            freeCompilerArgs.addAll(
+                listOf(
+                    "-linker-option",
+                    "-Wl,--allow-shlib-undefined"
+                )
+            )
+            arm64RuntimeLibDirs().forEach { armLibDir ->
                 freeCompilerArgs.addAll(
                     listOf(
                         "-linker-option",
-                        "-L$localSystemdLibDir",
+                        "-L$armLibDir",
                         "-linker-option",
-                        "-Wl,-rpath,$localSystemdLibDir"
+                        "-Wl,-rpath-link,$armLibDir"
                     )
                 )
             }
-            if (name.contains("LinuxArm64", ignoreCase = true)) {
-                freeCompilerArgs.addAll(
-                    listOf(
-                        "-linker-option",
-                        "-Wl,--allow-shlib-undefined"
-                    )
-                )
-                arm64RuntimeLibDirs().forEach { armLibDir ->
-                    freeCompilerArgs.addAll(
-                        listOf(
-                            "-linker-option",
-                            "-L$armLibDir",
-                            "-linker-option",
-                            "-Wl,-rpath-link,$armLibDir"
-                        )
-                    )
-                }
-            }
         }
-        (this as? KotlinNativeCompile)?.compilerOptions {
-            freeCompilerArgs.addAll(overrides)
-        }
+    }
+}
+
+tasks.withType<KotlinNativeCompile>().configureEach {
+    compilerOptions {
+        freeCompilerArgs.addAll(overrides)
     }
 }
 
@@ -321,23 +325,23 @@ license {
     ext["email"] = "monkopedia@gmail.com"
 }
 
-afterEvaluate {
-    tasks.findByName("licenseCheckForKotlin")?.let {
-        tasks.all {
-            if ((this.name.startsWith("ktlint") && this.name.endsWith("Check")) ||
-                (this.name.startsWith("transform") && this.name.endsWith("Metadata")) ||
-                (this.name.startsWith("compile") && this.name.contains("Kotlin")) ||
-                this.name.startsWith("link") ||
-                this.name == "copyLib" ||
-                this.name.endsWith("Test") ||
-                this.name.endsWith("Tests") ||
-                this.name == "processTestResources" ||
-                this.name == "test"
-            ) {
-                it.dependsOn(this)
-            }
+// `tasks.matching { }` is a live, lazily-realized view, so — unlike the eager `tasks.all { }` this
+// replaces — the task container is only walked when licenseCheckForKotlin is actually in the graph
+// instead of on every single invocation.
+tasks.named("licenseCheckForKotlin") {
+    dependsOn(
+        tasks.matching { task ->
+            (task.name.startsWith("ktlint") && task.name.endsWith("Check")) ||
+                (task.name.startsWith("transform") && task.name.endsWith("Metadata")) ||
+                (task.name.startsWith("compile") && task.name.contains("Kotlin")) ||
+                task.name.startsWith("link") ||
+                task.name == "copyLib" ||
+                task.name.endsWith("Test") ||
+                task.name.endsWith("Tests") ||
+                task.name == "processTestResources" ||
+                task.name == "test"
         }
-    }
+    )
 }
 allprojects {
     if (name == "compile_test") return@allprojects
--- a/compile_test/build.gradle.kts
+++ b/compile_test/build.gradle.kts
@@ -25,7 +25,7 @@ kotlin {
     }
     applyDefaultHierarchyTemplate()
     sourceSets {
-        val nativeMain by getting {
+        getByName("nativeMain") {
             dependencies {
                 implementation(libs.kotlinx.coroutines)
                 implementation(libs.kotlinx.serialization)
@@ -34,7 +34,7 @@ kotlin {
                 implementation(project(":"))
             }
         }
-        val nativeTest by getting {
+        getByName("nativeTest") {
             dependencies {
                 implementation(libs.kotlinx.coroutines.test)
             }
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
Independent verification (fresh checkout of `a343353`, JDK 21, all commands re-run myself — not just re-reading the PR body):

**Headline finding — confirmed exactly as claimed.** Reproduced the deprecation on `main`@`ba62857` with `--stacktrace`:
```
at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.getTaskDependencyFromProjectDependency(DefaultConfiguration.java:852)
at org.jetbrains.kotlin.gradle.targets.ConfigureBuildSideEffectKt$addDependsOnTaskInOtherProjects$1.execute(ConfigureBuildSideEffect.kt:35)
...
at org.gradle.api.internal.DefaultDomainObjectCollection.all(DefaultDomainObjectCollection.java:169)
at Build_gradle$4.execute(build.gradle.kts:171)
```
`build.gradle.kts:171` was indeed the eager `tasks.all { }`, and the deprecated call is genuinely made by KGP's `ConfigureBuildSideEffect` on task creation, not by this build script. There is no API to swap — the fix is correctly "stop force-realizing the task container," as the PR claims.

**Deprecation count.** Reproduced 15 → 0 myself: `grep -c "has been deprecated"` on `ba62857` = 15 (matches the PR table exactly, one of them being the `getTaskDependencyFromProjectDependency` line above); on `a343353`, 0 across `help`, `build --dry-run`, `license --dry-run`, `licenseCheckForKotlin --dry-run`, `tasks --all`, `publishToMavenLocal --dry-run`, `ktlintCheck`. Also swept the whole repo for `tasks.all`/`afterEvaluate`/`by getting`/`by creating` myself (not the same grep the author used) — none remain anywhere in the root/`compile_test`/`cross_test`/`stress_test`/`plugin` scripts. Two standalone sample builds (`samples/demo-service`, `samples/bluez-scan`) still have `by getting` — but they carry their own `settings.gradle.kts` and are **not** included in the root `settings.gradle.kts`, so they're genuinely outside this build's task graph and outside the issue's scope; not swept under the rug, just a separate build. Worth a follow-up ticket but not blocking.

**Scope widening (cross_test/stress_test/plugin).** Verified the identical `by getting`/`by creating` pattern existed in all three before this PR and is gone after. Justified — the issue's own acceptance criterion is a clean build under `--warning-mode all`, which these sites would otherwise have kept failing.

**Wiring preservation — the load-bearing risk, re-measured independently:**
- `licenseCheckForKotlin --dry-run`: diffed the raw output myself (not trusting the PR's claim) — identical 125-task graph before/after, only cosmetic diffs (`BUILD SUCCESSFUL in 1s` vs `2s`, and the now-absent deprecation-warning lines).
- `:dependencies` report across `:`, `:compile_test`, `:cross_test`, `:stress_test`: diffed myself — **zero added lines**; every removed line is either a `dokka*`-named configuration header (`dokkaGfmCollectorPlugin`, `dokkaHtmlCollectorRuntime`, `dokkaJavadocPlugin`, etc.) or a dependency under one of those configs (jackson/freemarker/jsoup/markdown — all dokka transitives). No non-dokka configuration lost a single dependency line.
- `publishToMavenLocal --dry-run`: confirmed all dokka javadoc jar tasks still present in the publish graph (`jvmDokkaJavadocJar`, `kotlinMultiplatformDokkaJavadocJar`, `linuxArm64/linuxX64DokkaJavadocJar`, `codegen:dokkaJavadocJar`, `plugin:javadocJar`) — nothing dropped from what gets published.
- Went further than the PR's own dry-run evidence: actually **ran** `compileKotlinLinuxX64 compileKotlinLinuxArm64 linkDebugSharedLinuxX64 linkDebugSharedLinuxArm64` for real. All green, including `linkDebugSharedLinuxArm64`, which is the most fragile consumer of the migrated `KotlinNativeLink.configureEach` wiring (systemd lib dirs, `-Wl,--allow-shlib-undefined`, arm64 runtime `-rpath-link` dirs) — if that wiring had been dropped, cross-linking would fail loudly, not silently.
- `tasks.matching { }` note from the dispatch: confirmed it's the right tool here specifically because the predicate only inspects `Task.name` (available on registered-but-unrealized providers), so `dependsOn(tasks.matching{...})` only force-realizes tasks that actually match the predicate at graph-computation time, not the whole container — that's the real mechanism by which the KGP-triggering `build` task stops getting realized.

**apiCheck.** Ran `:ktlintCheck :apiCheck` (plus `cross_test`/`stress_test`/`plugin` ktlint) myself — green, and `git status --short api/` is empty. Build-script-only change, confirmed.

**#168 overlap.** Both branch from `main`@`ba62857`. Did an actual test-merge of `issue-163` (#168) onto this PR's head in a scratch checkout: auto-merges cleanly. #168 touches `plugin/build.gradle.kts` lines ~81-95 (the `mavenPublishing`/`signing` SNAPSHOT-guard block); this PR touches line ~49 (`plugins.create("sdbusPlugin")` inside `gradlePlugin { }`, migrated from `by creating`). Disjoint regions, no textual conflict, and no semantic interaction — one is a Gradle Plugin Portal declaration, the other is publish signing. Confirmed the merged file contains both changes correctly.

**CI.** All required checks (`static-checks`, `full-tests-x64`, `native-build-x64`, `arm-runtime-tests`, full `pr-smoke` matrix) pass on head `a343353` — `mergeStateStatus: CLEAN`, re-checked immediately before this review to confirm the head hadn't moved. `full-tests-x64` runs `allTests test crossRuntimeInteropTests` under `dbus-run-session`, which covers `jvmTest`/`linuxX64Test`/dbusmock already, so I did not need to re-run those locally.

**Quality.** Clean, idiomatic migration: the `KotlinNativeLink`/`KotlinNativeCompile` split via typed `configureEach` also removes the unsafe `(this as? X)?` casts and the now-redundant `afterEvaluate` — a genuine improvement, not just a mechanical dodge. Comments explain *why* the lazy forms are structured this way, which will save the next reader from re-deriving this. No scope creep beyond the justified 3-file widening; no dead code left behind.

Category: `infra` (build-script-only Gradle-10 deprecation migration). No `sdbus-kotlin` override in policy — default `infra` = tier 1. Considered the reviewer-judgment lever re: "infra touching publish automation" — the publish graph is nominally touched (dokka task realization timing), but I verified empirically that publish *outputs* are unchanged (same dokka jars, same non-dokka dependency sets, real native links green), so this doesn't rise to the bump; it's inert by measurement, not by assumption.

Approving and merging.

PR #174, opened 2026-07-31
fix(jvm): give each brokerless direct connection its own unique name

monkopedia-coder:
Refs #141 — item 4 of the three remaining in-scope items from the 2026-07-31 triage re-scope.

## The divergence

A brokerless **direct** (peer-to-peer) connection never runs `Hello` — there is no daemon to run it against — so `DBusWireConnection.uniqueName` is `null` for one. `WireDbusConnection` filled that hole with a **shared constant**:

```kotlin
private val localUniqueName: String = wire.uniqueName.orEmpty()
...
override fun uniqueName(): BusName = BusName(localUniqueName.ifEmpty { ":jvm-wire" })
```

That name is not cosmetic — it is the **key of every in-process registry**: `JvmStaticDispatch`, `WireServeRegistry`, `LocalObjectManagerRegistry`, `LocalJvmServiceRegistry`. So two direct connections in one JVM were one identity: exporting an object at the same path from both meant the second `JvmStaticDispatch.register` **overwrote** the first, and *both* connections' proxies then reached the survivor. Native has no such constant, so there is nothing on that side to diverge from — this is a JVM-internal identity bug, which is why triage scoped it as "a bug fix, not a contract change".

## Red-first evidence

New test `src/jvmTest/.../JvmDirectConnectionIdentityTest.kt`, run on the **unmodified** backend (the exact committed test source, with only `WireDbusBackend.kt` reverted):

```
> Task :jvmTest FAILED
JvmDirectConnectionIdentityTest[jvm] > twoDirectConnectionsHaveDistinctUniqueNames[jvm] FAILED
JvmDirectConnectionIdentityTest[jvm] > objectsOnTwoDirectConnectionsDoNotClobberEachOther[jvm] FAILED
2 tests completed, 2 failed
```

with, from `build/test-results/jvmTest/TEST-...JvmDirectConnectionIdentityTest.xml`:

```
message="java.lang.AssertionError: Values should be different. Actual: :jvm-wire"
message="org.junit.ComparisonFailure: expected:<[a]> but was:<[b]>"
```

The second one is the consequence, not just the symptom: connection **A**'s proxy, addressed to A's own unique name, was served by connection **B**'s object.

Both pass after the fix; the full `jvmTest` and `linuxX64Test` suites pass.

### Why this test is JVM-only rather than cross-backend

The epic's method is cross-backend-red-first, and I could not make this one cross-backend honestly: the divergent value (`":jvm-wire"`) is a JVM-internal invention with no native counterpart, and a shared fixture is impossible in either direction — the JVM backend has no fd-based direct factories (`createDirectBusConnection(fd)`/`createServerBusConnection(fd)` are `@Deprecated(ERROR)` there), while native's direct-connection fixture is fd-based. The test therefore lives in `jvmTest` and proves the JVM behaviour before and after directly.

The test brings its own peer: `DirectPeer` is a ~40-line AF_UNIX listener (pure JDK `ServerSocketChannel`/`UnixDomainSocketAddress`, no new dependency) that completes the SASL EXTERNAL handshake and then idles — exactly enough for `createDirectBusConnection` to establish a direct connection, which sends nothing else on its own.

## The fix

```kotlin
private val directConnectionCounter = AtomicLong()
...
private val localUniqueName: String =
    wire.uniqueName ?: ":jvm-wire-${directConnectionCounter.incrementAndGet()}"
```

and `uniqueName()` returns it verbatim (the `ifEmpty` fallback is gone — `localUniqueName` can no longer be empty).

Bus connections are untouched: they have a daemon-assigned name, so `wire.uniqueName` is non-null and nothing about them changes.

One side effect worth naming, because it is a second latent bug the same constant was hiding: on a direct connection, objects were **registered** under `dispatchDestination` (`":jvm-wire"`, read back through `connection.uniqueName`) but **served** under `localUniqueName` (`""`), so `WireServe`'s registration and serve keys did not match. With a single non-empty name for both, they now do.

## Verification

- `dbus-run-session -- ./gradlew jvmTest linuxX64Test` — BUILD SUCCESSFUL (full root suites, both backends).
- `./gradlew ktlintCheck apiCheck compileKotlinLinuxArm64` — BUILD SUCCESSFUL.
- **`apiCheck` did NOT move.** `WireDbusConnection` is `internal`; no public surface change, so neither `api/sdbus-kotlin.api` nor `api/sdbus-kotlin.klib.api` needed a regen and there is no consumer-visible ABI change for `blue-falcon-sdbus`.

`docs/BACKENDS.md` is updated in step: the "two direct connections share `":jvm-wire"`" entry is removed from the JVM-limitations list and the per-connection synthetic name is documented with the connection factories.

### DIFF
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -274,6 +274,11 @@ There are 11 `create*Connection` entry points — 10 declared in the common API
   JVM backend rejects the fd-based bus types (`JvmBusType.DIRECT_FD`/`SERVER_FD`, in
   `WireDbusBackend.createConnection`).
   On native they adopt the descriptor as documented in their KDoc.
+- A **brokerless direct connection** has no daemon to run `Hello` against, so neither backend gets
+  a bus-assigned unique name for it. The JVM backend synthesizes one **per connection**
+  (`":jvm-wire-<n>"`) — the in-process registries are keyed by it, so two direct connections in one
+  JVM keep separate identities and objects they export at the same path no longer clobber each
+  other (#141, pinned by `src/jvmTest/.../JvmDirectConnectionIdentityTest.kt`).
 - `createRemoteSystemBusConnection(host)` is **native-only and not declared in the common
   or JVM API** (it lives in `src/nativeMain/.../Connection.native.kt`). It opens the system
   bus of a remote host over ssh via sd-bus (`sd_bus_open_system_remote`, which spawns
@@ -363,8 +368,6 @@ cross-backend regression tests).
 A few **same-process / direct-connection** edges remain JVM-specific and are documented rather than
 matched (they don't affect cross-process usage):
 
-- Two brokerless **direct** connections in one JVM share a synthetic `uniqueName` (`":jvm-wire"`),
-  and objects they export at the same path can clobber each other in the local dispatch table.
 - `startEventLoop()` is a no-op on JVM (the reader thread auto-starts at connect), where native
   requires it to receive — a JVM proxy can receive signals without it; native cannot.
 - `Properties.GetAll` on an interface the object doesn't implement returns an empty `a{sv}` on JVM,
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt
@@ -28,6 +28,7 @@ import com.monkopedia.sdbus.signalFromMetadata
 import java.io.IOException
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicLong
 import kotlin.concurrent.thread
 import kotlin.time.Duration
 
@@ -158,8 +159,17 @@ private fun emitWireSignal(
     }
 }
 
+// A brokerless DIRECT connection is not told a unique name — there is no daemon to run Hello
+// against — but the in-process registries (JvmStaticDispatch, WireServeRegistry,
+// LocalObjectManagerRegistry, ...) are all keyed by it. Every such connection used to answer with
+// the same constant ":jvm-wire", so two of them in one JVM shared one identity and objects they
+// exported at the same path clobbered each other in the dispatch table. Give each its own name
+// instead. #141.
+private val directConnectionCounter = AtomicLong()
+
 internal class WireDbusConnection(private val wire: DBusWireConnection) : JvmDbusConnection {
-    private val localUniqueName: String = wire.uniqueName.orEmpty()
+    private val localUniqueName: String =
+        wire.uniqueName ?: ":jvm-wire-${directConnectionCounter.incrementAndGet()}"
     private val released = AtomicBoolean(false)
     private var timeout: Duration = Duration.ZERO
 
@@ -209,7 +219,7 @@ internal class WireDbusConnection(private val wire: DBusWireConnection) : JvmDbu
 
     override fun uniqueName(): BusName {
         checkNotReleased()
-        return BusName(localUniqueName.ifEmpty { ":jvm-wire" })
+        return BusName(localUniqueName)
     }
 
     override fun requestName(name: ServiceName, flags: UInt): RequestNameReply {
--- a/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmDirectConnectionIdentityTest.kt
+++ b/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmDirectConnectionIdentityTest.kt
@@ -1,0 +1,143 @@
+package com.monkopedia.sdbus
+
+import java.io.Closeable
+import java.io.IOException
+import java.net.StandardProtocolFamily
+import java.net.UnixDomainSocketAddress
+import java.nio.channels.Channels
+import java.nio.channels.ServerSocketChannel
+import java.nio.channels.SocketChannel
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.concurrent.thread
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+
+/**
+ * A brokerless DIRECT connection has no daemon to hand it a unique name, and the JVM backend used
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed at `26286f84054f1fceee70eeb76a3695b102d4c7a7` in a fresh clone/checkout — independently verified, not taken on the author's word.

**Red-first, reproduced myself.** Checked out `main` (`e3eda9d`, the actual merge base), applied only the new test file `src/jvmTest/.../JvmDirectConnectionIdentityTest.kt` on top of the unmodified `WireDbusBackend.kt`, and ran it under `dbus-run-session`. Both tests failed, verbatim matching the PR body's claims:

```
java.lang.AssertionError: Values should be different. Actual: :jvm-wire
org.junit.ComparisonFailure: expected:<[a]> but was:<[b]>
```

The second failure is the real one: connection A's proxy (destined for A's own unique name) was answered by connection B's exported object. Restored the PR's actual fix afterward and both tests go green; a guard that can't fail is worthless, and this one demonstrably can.

**The serve-key/routing side effect — traced end to end, not just accepted.** Confirmed the mechanism precisely:
- Object *registration* (`WireDbusObject.dispatchDestination`) is keyed off `connection.uniqueName.value` — the **public** name, which pre-fix funneled through `uniqueName().ifEmpty { ":jvm-wire" }` and so was `":jvm-wire"` for every direct connection.
- Incoming-call *serving* (`WireDbusConnection.init` → `WireServe.handleIncomingCall(wire, localUniqueName, ...)`) used the **raw** `localUniqueName`, which pre-fix was `""` for a direct connection (no `ifEmpty` applied there).
- `JvmStaticDispatch`'s table is keyed by `(destination, path, interface, method, argCount)`. So pre-fix: two direct connections both registered objects under `destination=":jvm-wire"` (the second `ConcurrentHashMap.put` overwrote the first — this is the collision the test proves), **and** any genuine incoming wire call on a direct connection's own socket looked up `destination=""`, which never matched what was registered under `":jvm-wire"` — a second, independent latent bug in the same constant.
- Post-fix, `localUniqueName` is one synthesized value used uniformly for both registration (via `connection.uniqueName.value`, now equal to `localUniqueName` since the `ifEmpty` fallback is gone) and serving. Both bugs close by construction — the fix doesn't special-case them separately, which is the right shape (native-parity-style "make the wrong binding unrepresentable" rather than patching each call site).

**Blast radius, checked, not assumed.** Bus connections are provably untouched — `wire.uniqueName` is non-null for anything that ran `Hello`, so `localUniqueName` takes the same branch as before, byte for byte. The change only reaches brokerless DIRECT connections — a same-process, no-daemon construct with zero external Maven consumers (grepped: `blue-falcon-sdbus` and BlueZ usage go over the system bus, never `createDirectBusConnection`/`createServerBusConnection`). `docs/BACKENDS.md` already documented the old shared-name behavior as a bug/limitation being removed, not a guaranteed contract, so no consumer can have been relying on the specific string `":jvm-wire"`.

**Counter.** `AtomicLong.incrementAndGet()` on a file-level singleton — atomic, monotonic, no reset path within a JVM's lifetime; this is the "counter" collision mode from the identity-derivation checklist, but scoped entirely in-process (no durable/cross-restart state exists for direct connections to begin with, so the usual "counter resets, stale ids survive" failure mode doesn't apply here). Format `:jvm-wire-<n>` carries no D-Bus wire semantics — verified it is never placed into an outgoing `WireMessage.sender`/`destination` field (the bus, when present, stamps `sender`; a direct connection has no bus to receive this string at all), so the lack of a `.` separator (same shape the old `:jvm-wire` constant already had) is not a regression and not wire-visible.

One minor, non-blocking observation for the record: `wire.uniqueName ?: ":jvm-wire-$n"` only substitutes on `null`; a hypothetical bus daemon that replied to `Hello` with a present-but-empty string would now surface `""` instead of pre-fix's `":jvm-wire"` fallback. This requires a nonconformant daemon and isn't exercised by any real broker or dbusmock, so not worth blocking on.

**jvmTest-only, justified.** Confirmed independently: `WireDbusBackend.createConnection` explicitly rejects `fd != null || busType in {DIRECT_FD, SERVER_FD}` ("JVM wire backend does not support fd-based connections"), so there is no JVM fd-based direct factory to share with native's fixture, and native's own direct-connection fixture (`src/nativeTest/.../DirectConnectionFixture.kt`) is built on `UnixFd`/`createServerBusConnection`. A cross-backend fixture is not achievable without new infrastructure; scoping this test to `jvmTest` is the right call.

**Test quality.** `DirectPeer` is a clean ~40-line AF_UNIX SASL-EXTERNAL stub: daemon threads throughout (no hang risk on CI even if a handshake thread never unblocks), `close()` closes the listener and joins the acceptor with a bounded 1s timeout, and the two-connection/two-object shape satisfies the "needs TWO colliding things" evidence bar rather than a single-instance test that would pass identically against the bug.

**Static gates**, run myself: `ktlintCheck apiCheck compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` — BUILD SUCCESSFUL, including `klibApiCheck`. **`apiCheck` and `klibApiCheck` did not move** — confirmed, not just trusted: `WireDbusConnection` is `internal`, and the diff touches no public declarations, so `api/*.api` / `api/sdbus-kotlin.klib.api` (consumed by `blue-falcon-sdbus`) are unaffected.

**Full suites**, run myself under `dbus-run-session`: `jvmTest linuxX64Test` — BUILD SUCCESSFUL, including `cross_test:jvmTest`/`cross_test:linuxX64Test`. The new test's XML report shows both cases passing (`tests="2" failures="0"`).

**Out-of-scope check.** Diff is exactly 3 files (`WireDbusBackend.kt`, the new test, `docs/BACKENDS.md`) — confirmed nothing from #141's still-open items (`Properties.GetAll` unknown-interface, same-process `UnixFd` dup, wrong-destination single-candidate fallback) was touched.

**Remote CI**, watched to completion: all required checks (`static-checks`, `native-build-x64`, `full-tests-x64`, `arm-runtime-tests`, all `pr-smoke` variants) pass; `cleaner-soak`/`nightly-stress` show `skipping` (expected — nightly-only jobs, not required on a PR). `mergeStateStatus` is `CLEAN`/`MERGEABLE` at the reviewed head, re-checked via `git ls-remote` immediately before this review, unchanged throughout.

Category: `bug_fix` (no public API/ABI surface change; the observable change is scoped exactly to fixing the identity bug it targets — the serve/registration-key alignment is a corollary of the same key unification, not a separate unrelated behavior change). Tier: 1 per this repo's default mapping (no override). Approving and merging.

Issue #179, opened 2026-08-03
MethodInvoker.timeout = Duration.INFINITE does not mean the connection default, despite its KDoc


monkopedia-coder:
**The one-sentence question:** should `MethodInvoker.timeout = Duration.INFINITE` map to sd-bus's `0` sentinel (meaning "use the connection default", which is what its KDoc currently claims) — accepting that calls then fail at sd-bus's 25-second default — or should the KDoc be corrected to say that `INFINITE` means genuinely infinite?

## Current state

`MethodInvoker.kt:182-183`:

```kotlin
/** The per-call timeout. Defaults to [Duration.INFINITE], meaning the connection default. */
var timeout: Duration = INFINITE
```

The documented behavior and the actual behavior differ. `Duration.INFINITE.inWholeMicroseconds` clamps to `Long.MAX_VALUE`, and every native call path passes that straight through (`ProxyImpl.kt:104` and `:152`, `MethodCall.native.kt:62`). sd-bus's "use the connection default" sentinel is `0`, so what it actually receives is a near-infinite explicit timeout, not a request for the default. The KDoc has been wrong since it was written.

So today `INFINITE` behaves as genuinely infinite, and `Connection.methodCallTimeout` is effectively unreachable through the per-call API, since a per-call timeout beats the connection default.

## Why this is your call rather than an engineering one

The two answers produce different observable behavior for existing consumers of a shipped 1.0.1, and which one is right depends on what you want the API to promise:

- **Map `INFINITE` to `0`** — the KDoc becomes true, `Connection.methodCallTimeout` becomes reachable as documented, and a call left at the default now fails after sd-bus's 25 s. That is a behavior change for anything relying on long calls completing. A BlueZ `Connect()` to an unreachable device takes about 48 s, so it would start surfacing as a D-Bus timeout instead of BlueZ's real `le-connection-abort-by-local` error.
- **Correct the KDoc** — no behavior change, and `INFINITE` honestly means infinite. Consumers who want the connection default would then need an explicit way to ask for it, since there is currently no value that maps to sd-bus `0`.

The `blue-falcon-sdbus` agent that surfaced this deliberately did not answer it, and said why: it would prefer to keep BlueZ's real error, but that is a preference about their use case rather than an argument about this API, and this library has consumers they cannot see.

## Not blocking the bug fix

[...]

PR #180, opened 2026-08-03
fix(native): stop the event loop busy-spinning on a default-timeout method call

### LINKED ISSUE #178: Native event loop busy-spins a core for the whole of any default-timeout method call (uint64 deadline overflows to negative)
Every native D-Bus method call left at the default timeout busy-spins a CPU core for the call's entire duration. Normal calls take milliseconds so this is invisible; on a long call it is operationally indistinguishable from a hang.

Reported by the `blue-falcon-sdbus` agent, traced from Monkopedia/blue-falcon-sdbus#16, where a BlueZ `Connect()` to an unreachable device takes about 48 seconds. **Native targets only — JVM is unaffected.** I have independently confirmed the mechanism against this repo's source; the measurements below are theirs.

## Mechanism

1. `MethodInvoker.timeout` defaults to `Duration.INFINITE` (`MethodInvoker.kt:183`), and every native call path converts it with `timeout.inWholeMicroseconds.toULong()` (`ProxyImpl.kt:104` and `:152`, `MethodCall.native.kt:62`). `Duration.INFINITE.inWholeMicroseconds` clamps to `Long.MAX_VALUE`, so sd-bus receives `9223372036854775807`, **not** its `0` "use the connection default" sentinel.
2. sd-bus stores an absolute deadline of `now + 2^63-1` microseconds against `CLOCK_MONOTONIC`. Measured downstream as `0x8000_0044_...`, matching a `CLOCK_MONOTONIC` sample at the same instant to within one microsecond.
3. `ConnectionImpl.kt:875-879` then does:

   ```kotlin
   val timeout = if (pollData.timeout_usec == UINT64_MAX) {
       Duration.INFINITE
   } else {
       pollData.timeout_usec.toLong().microseconds
   }
   ```

   `timeout_usec` is a `uint64_t`. The stored deadline has bit 63 set but is **not** `UINT64_MAX`, so the guard does not fire and `.toLong()` overflows to a negative value.
4. `PollData.getRelativeTimeout()` (`PollData.kt:63-67`) computes `(timeout - now()).coerceAtLeast(Duration.ZERO)`, which on a negative timeout yields `ZERO`. `getPollTimeout()` therefore returns `0`, and `waitForNextEvent()` calls `poll(fds, 3, 0)` (`ConnectionImpl.kt:834-835`) in an unbounded busy re-poll.

## Measurements (downstream, isolated from their engine)

Raw `createSystemBusConnection` plus `callMethodAsync`, no engine code involved:

| Call | CPU |
| --- | --- |
| default timeout, 48.6 s | **104%** |
| identical call, `timeout = 120.seconds`, 47.8 s | **0%** |
| same probe on the JVM target, 50 s | ~1.0 s CPU total |

Backtrace lands at `ConnectionImpl.kt:835` inside `EventLoopThread.waitForNextEvent`, locals showing `timeout = 0` and `poll` returning `r = 0`, roughly 90% user / 10% system.

They also falsified their own first hypothesis by experiment — they suspected their `destroy()` teardown, and 8 back-to-back construct-then-destroy cycles with zero suspension points gave 0% CPU and a clean exit.

## Why JVM is unaffected

The JVM backend clamps instead of overflowing: `WireDbusBackend.kt:178` does `timeout.inWholeMicroseconds.coerceAtLeast(0L).toULong()`, and the JVM dispatcher does not derive a `poll` timeout from an sd-bus absolute deadline at all.

## Proposed fix

In `getEventLoopPollData()`, treat any `timeout_usec` with bit 63 set as infinite alongside the existing `UINT64_MAX` check — or compare as unsigned throughout so the conversion cannot overflow. Note that a *correct* absolute `CLOCK_MONOTONIC` deadline can never legitimately have bit 63 set (that is over 292,000 years of uptime), so treating it as infinite is safe rather than a heuristic.

This is a bug fix with no intended behavior change for any call that already works.

## Testing

No BLE hardware required, and no D-Bus session bus required for the core of it: `src/nativeTest/kotlin/unit/PollDataTest.kt` already unit-tests `getRelativeTimeout()` directly, so the overflow is reproducible red-first as a pure unit test by constructing a `PollData` with a bit-63-set microsecond value. An end-to-end assertion (a long call must not spin) is also possible against a slow test method on the existing fixtures.

## Related

The separate question of what `Duration.INFINITE` *should* map to is filed separately as an owner decision, because the two are independent: this overflow is a defect on any code path that produces a large deadline, and fixing it does not depend on resolving the semantics. Downstream cannot work around this locally — generated proxies pass no timeout, and a per-call timeout beats `Connection.methodCallTimeout`, so there is no consumer-side knob.

monkopedia-coder:
Fixes #178

Native only. Every D-Bus method call left at the default timeout busy-spun a CPU core for the call's whole duration — invisible on millisecond calls, indistinguishable from a hang on a long one. Reported downstream by the `blue-falcon-sdbus` agent in Monkopedia/blue-falcon-sdbus#16, where a BlueZ `Connect()` to an unreachable device runs ~48 s and pinned a core at **104% CPU**; the same call with an explicit `timeout = 120.seconds` measured 0%.

## Root cause

`EventLoopThread.getEventLoopPollData()` read sd-bus' absolute `uint64_t` poll deadline and treated only `UINT64_MAX` as "indefinite", converting everything else with `.toLong()`:

```kotlin
val timeout = if (pollData.timeout_usec == UINT64_MAX) {
    Duration.INFINITE
} else {
    pollData.timeout_usec.toLong().microseconds
}
```

A default-timeout call asks sd-bus for `Long.MAX_VALUE` microseconds (`Duration.INFINITE.inWholeMicroseconds` clamps there), so sd-bus stores an absolute `CLOCK_MONOTONIC` deadline of `now + 2^63-1` — **bit 63 set, but nowhere near `UINT64_MAX`**. The guard missed it and `.toLong()` overflowed negative. `PollData.getRelativeTimeout()` then `coerceAtLeast(Duration.ZERO)`'d that to `ZERO`, `getPollTimeout()` returned `0`, and `waitForNextEvent()` called `poll(fds, 3, 0)` — an unbounded busy re-poll.

The same overflow also caused a spurious event-loop wakeup in `callMethod()`, which compares the poll deadline before and after `send()` and saw the negative value as smaller than the previous one. That is fixed as a side effect.

## The fix

One line of logic: compare **unsigned** rather than testing a single sentinel.

```kotlin
internal fun absoluteTimeoutOf(timeoutUsec: ULong): Duration =
    if (timeoutUsec > Long.MAX_VALUE.toULong()) {
        Duration.INFINITE
    } else {
        timeoutUsec.toLong().microseconds
    }
```

Both forms suggested in the issue work. I chose the unsigned comparison over an explicit bit-63 test because it is **structurally** safe rather than correct-by-inspection: it subsumes the `UINT64_MAX` sentinel (so there is no longer a special case to keep in sync), and it leaves *no* signed conversion on any path — the `.toLong()` is unreachable for any value that could overflow it, by construction. A bit-63 mask would read as a heuristic and would still leave a `.toLong()` a future edit could widen.

Soundness of treating bit 63 as infinite: a genuine absolute `CLOCK_MONOTONIC` deadline cannot set bit 63 — that is over 292,000 years of uptime — so no reachable timeout changes meaning. Calls that already work are unaffected.

The conversion moved out of the private inner `EventLoopThread` into `PollData.kt` purely so it is reachable from a unit test; the call site in `ConnectionImpl.kt` gets shorter, and the `UINT64_MAX` import goes away. Nothing else changed.

## Red-first

Two commits, so the red is reproducible independently: `a504659` moves the conversion out **verbatim, bug included**, and adds the tests; `badce2e` fixes it.

At `a504659` (`./gradlew linuxX64Test --tests '*PollDataTest*'`):

```
com.monkopedia.sdbus.unit.PollDataTest.PollData ConvertsDefaultTimeoutDeadlineToInfiniteTimeout[linuxX64] FAILED
    kotlin.AssertionError at null:-1

com.monkopedia.sdbus.unit.PollDataTest.PollData ReturnsNegativePollTimeoutForDefaultTimeoutDeadline[linuxX64] FAILED
    kotlin.AssertionError at null:-1

> Task :linuxX64Test FAILED

12 tests completed, 2 failed
```

with the assertion messages from `build/test-results/linuxX64Test/*.xml`:

```
kotlin.AssertionError: Expected <Infinity>, actual <-(106751988d 19h 20m 57.067s)>.
kotlin.AssertionError: Expected <-1>, actual <0>.
```

`-(106751988d 19h 20m 57.067s)` is exactly `-Long.MAX_VALUE` microseconds — the overflowed value. The second failure is the defect itself at the end of the chain: `poll(2)` was being handed `0` (spin) where it must get `-1` (block).

Four tests added, all in `PollDataTest`. Two are the red above; the other two guard the fix's edges — that `UINT64_MAX` still maps to `INFINITE`, and that the largest deadline which *does* fit in a signed 64-bit integer stays finite (so the widened condition is not off by one and does not swallow real deadlines).

## No end-to-end test — deliberate

I did not add an integration assertion that a slow call must not spin. To observe the spin you have to sample the event-loop thread's CPU time across a multi-second in-flight call and assert on a usage threshold — seconds of added CI wall time for an assertion a loaded shared runner can trip either way. The unit tests instead pin the exact deterministic quantity that caused it (`getPollTimeout()` returns `-1`, not `0`), which is what `poll(2)` actually consumes. The one link they do not cover is "sd-bus really does return a bit-63 deadline here", and that is established by the downstream measurement in the issue (observed `0x8000_0044_...`, matching a `CLOCK_MONOTONIC` sample at the same instant to within one microsecond).

## Out of scope

**#179 is untouched.** What `Duration.INFINITE` *should* map to — the `MethodInvoker.timeout` default and its KDoc, and whether an infinite timeout should send sd-bus' `0` "use connection default" sentinel instead of `Long.MAX_VALUE` — is a separate owner decision about behaviour already shipped in 1.0.1. `MethodInvoker.kt` is not in this diff. This fix stands regardless of how #179 goes: the overflow is a defect on *any* path producing a large deadline.

## Verification

`JAVA_HOME=/usr/lib/jvm/java-21-openjdk`

- `dbus-run-session -- ./gradlew linuxX64Test jvmTest` — green. 296 native / 326 JVM tests, 0 failures, 0 skipped.
- `./gradlew compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` — green.
- `./gradlew ktlintCheck apiCheck` — green. **`apiCheck` did not move**: `absoluteTimeoutOf` is `internal`, so neither `api/sdbus-kotlin.api` nor `api/sdbus-kotlin.klib.api` changed. No regeneration, and no klib-ABI shift for `blue-falcon-sdbus`, which consumes it.

No `*.gradle.kts` touched (open PRs #168/#171 are on build scripts).

### DIFF
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt
@@ -91,7 +91,6 @@ import platform.posix.EEXIST
 import platform.posix.EINTR
 import platform.posix.EINVAL
 import platform.posix.POLLIN
-import platform.posix.UINT64_MAX
 import platform.posix.close
 import platform.posix.errno
 import platform.posix.poll
@@ -872,11 +871,7 @@ internal class ConnectionImpl(private val sdbus: ISdBus, private val bus: BusPtr
 
             require(eventFd.fd >= 0)
 
-            val timeout = if (pollData.timeout_usec == UINT64_MAX) {
-                Duration.INFINITE
-            } else {
-                pollData.timeout_usec.toLong().microseconds
-            }
+            val timeout = absoluteTimeoutOf(pollData.timeout_usec)
 
             return PollData(pollData.fd, pollData.events, timeout, 0) // eventFd_.fd)
         }
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/PollData.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/PollData.kt
@@ -23,6 +23,18 @@
 package com.monkopedia.sdbus.internal
 
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.microseconds
+import platform.posix.UINT64_MAX
+
+/**
+ * Converts an absolute sd-bus timeout, expressed in microseconds of CLOCK_MONOTONIC, into the
+ * [Duration] carried by [PollData.timeout].
+ */
+internal fun absoluteTimeoutOf(timeoutUsec: ULong): Duration = if (timeoutUsec == UINT64_MAX) {
+    Duration.INFINITE
+} else {
+    timeoutUsec.toLong().microseconds
+}
 
 /**
  * Carries poll data needed for integration with external event loop implementations.
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/PollData.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/PollData.kt
@@ -24,17 +24,24 @@ package com.monkopedia.sdbus.internal
 
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.microseconds
-import platform.posix.UINT64_MAX
 
 /**
  * Converts an absolute sd-bus timeout, expressed in microseconds of CLOCK_MONOTONIC, into the
  * [Duration] carried by [PollData.timeout].
+ *
+ * Anything that does not fit in a signed 64-bit integer is indefinite. That covers sd-bus'
+ * `UINT64_MAX` "no timeout" sentinel, and also the deadline a call left at the default timeout
+ * produces: it asks sd-bus for `Long.MAX_VALUE` microseconds, which sd-bus stores as an absolute
+ * `now + Long.MAX_VALUE` with bit 63 set. A real deadline cannot set bit 63 — that is over 292,000
+ * years of uptime — so there is no value to lose, and comparing unsigned leaves no signed
+ * conversion that could overflow into a negative (and therefore already-expired) timeout.
  */
-internal fun absoluteTimeoutOf(timeoutUsec: ULong): Duration = if (timeoutUsec == UINT64_MAX) {
-    Duration.INFINITE
-} else {
-    timeoutUsec.toLong().microseconds
-}
+internal fun absoluteTimeoutOf(timeoutUsec: ULong): Duration =
+    if (timeoutUsec > Long.MAX_VALUE.toULong()) {
+        Duration.INFINITE
+    } else {
+        timeoutUsec.toLong().microseconds
+    }
 
 /**
  * Carries poll data needed for integration with external event loop implementations.
--- a/src/nativeTest/kotlin/unit/PollDataTest.kt
+++ b/src/nativeTest/kotlin/unit/PollDataTest.kt
@@ -25,6 +25,7 @@
 package com.monkopedia.sdbus.unit
 
 import com.monkopedia.sdbus.internal.PollData
+import com.monkopedia.sdbus.internal.absoluteTimeoutOf
 import com.monkopedia.sdbus.internal.now
 import kotlin.test.Test
 import kotlin.test.assertEquals
@@ -34,6 +35,7 @@ import kotlin.time.Duration.Companion.microseconds
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
 import kotlinx.cinterop.ExperimentalForeignApi
+import platform.posix.UINT64_MAX
 
 class PollDataTest {
 
@@ -111,5 +113,37 @@ class PollDataTest {
         val pollTimeout = pd.getPollTimeout()
 
         assertTrue(pollTimeout in 900..1100)
+    }
+
+    @Test
+    fun `PollData ConvertsUint64MaxSentinelToInfiniteTimeout`() {
+        assertEquals(Duration.INFINITE, absoluteTimeoutOf(UINT64_MAX))
+    }
+
+    @Test
+    fun `PollData ConvertsDefaultTimeoutDeadlineToInfiniteTimeout`() {
+        // A call left at the default timeout sends sd-bus Long.MAX_VALUE microseconds, so sd-bus
+        // stores an absolute deadline of now + Long.MAX_VALUE. That has bit 63 set but is not
+        // UINT64_MAX.
+        val deadline = now().inWholeMicroseconds.toULong() + Long.MAX_VALUE.toULong()
+
+        assertEquals(Duration.INFINITE, absoluteTimeoutOf(deadline))
+    }
+
+    @Test
+    fun `PollData ReturnsNegativePollTimeoutForDefaultTimeoutDeadline`() {
+        // The whole point: poll(2) must block rather than return immediately, or the event loop
+        // busy-spins for the entire duration of the call.
+        val deadline = now().inWholeMicroseconds.toULong() + Long.MAX_VALUE.toULong()
+        val pd = PollData(timeout = absoluteTimeoutOf(deadline))
+
+        assertEquals(-1, pd.getPollTimeout())
+    }
+
+    @Test
+    fun `PollData ConvertsRepresentableDeadlinesExactly`() {
+        assertEquals(1_500_000.microseconds, absoluteTimeoutOf(1_500_000uL))
+        // The largest deadline that still fits in a signed 64-bit integer stays finite.
+        assertEquals(Long.MAX_VALUE.microseconds, absoluteTimeoutOf(Long.MAX_VALUE.toULong()))
     }
 }


monkopedia-reviewer (approved):
Independent verification, fresh clone/checkout, JAVA_HOME=java-21-openjdk.

**Red-first, reproduced myself.** Checked out `a504659` and ran `./gradlew linuxX64Test --tests '*PollDataTest*'`:

```
com.monkopedia.sdbus.unit.PollDataTest.PollData ConvertsDefaultTimeoutDeadlineToInfiniteTimeout[linuxX64] FAILED
com.monkopedia.sdbus.unit.PollDataTest.PollData ReturnsNegativePollTimeoutForDefaultTimeoutDeadline[linuxX64] FAILED
12 tests completed, 2 failed
```
matching the PR's claimed failure exactly. On `badce2e` (the fix commit), the same 12 tests, including those two, all pass — proving the test is red against the bug and green against the fix.

**Unsigned-comparison boundary check, done independently:** `absoluteTimeoutOf` predicate is `timeoutUsec > Long.MAX_VALUE.toULong()`.
- `0uL` → not > → `0.microseconds`. Correct.
- `Long.MAX_VALUE.toULong()` → not > (equal) → finite, `Long.MAX_VALUE.microseconds`. Correct — this is exactly the largest representable deadline and must stay finite (covered by `ConvertsRepresentableDeadlinesExactly`).
- `Long.MAX_VALUE.toULong() + 1uL` (2^63, bit 63 set) → INFINITE. Correct — this is the actual overflow case from the bug.
- `UINT64_MAX` → INFINITE. Still correctly handled — the new predicate strictly subsumes the old sentinel check (`> Long.MAX_VALUE` unsigned is exactly "bit 63 set", which includes `UINT64_MAX`). Nothing is lost by dropping the explicit `UINT64_MAX` special case; it's covered by construction, not by inspection. Agree with the author's reasoning over a bit-63 mask: this leaves no reachable `.toLong()` that could overflow.

**Second defect (spurious wakeup in `callMethod` at ConnectionImpl.kt:598-605) — verified independently, real.** Read the code directly (not `grep`, this file is fine but I double-checked): `timeoutBefore`/`timeoutAfter` are both `getEventLoopPollData().timeout`. Before the fix, issuing a new default-timeout async call makes the post-`send()` poll deadline overflow to a large negative `Duration`; since `timeoutBefore` was a legitimate (larger) value, `timeoutAfter < timeoutBefore` spuriously evaluates true on essentially every default-timeout async call, firing an unnecessary `notifyEventLoopToWakeUpFromPoll()`. After the fix, `timeoutAfter` correctly resolves to `Duration.INFINITE` and the comparison behaves as intended. The PR body describes this accurately.

**Scope check:** `MethodInvoker.kt` is untouched (confirmed via the PR's file list — only `ConnectionImpl.kt`, `PollData.kt`, `PollDataTest.kt` changed). Issue #179 (the `INFINITE`→sd-bus-`0` semantics question) is independent and correctly left alone. Per the coordinator's follow-up check: #179 quotes `MethodInvoker.kt:182-183` verbatim, and I confirmed against current `main` that the citation is accurate (the KDoc still reads "meaning the connection default" while the field defaults to `INFINITE`, i.e. `Long.MAX_VALUE` µs, unchanged). #179 states the drift plainly, has file:line citations, isn't stale, and explicitly disclaims blocking #180. #180 doesn't pre-empt or depend on how #179 resolves — the overflow fix is correct regardless of what `INFINITE` is defined to mean. No action needed on either PR/issue for this.

**Refactor scope:** minimal — extracting `absoluteTimeoutOf` out of the private `EventLoopThread` inner class into `PollData.kt` is the smallest change that makes it unit-testable (the old code lived in a private nested class, unreachable from `nativeTest`). Call site shrank from 5 lines to 1, `UINT64_MAX` import dropped since it's now internal to the extracted function. No scope creep.

**End-to-end test gap:** agree with the author's call not to add a CPU-threshold integration test — that trades a flaky, runner-load-sensitive assertion for what the unit tests already pin exactly (`getPollTimeout()` returns `-1`, the deterministic quantity `poll(2)` actually consumes). I looked at whether a cheap non-flaky integration alternative exists (e.g., issuing a real default-timeout call and inspecting the actual sd-bus poll data): `EventLoopThread.getEventLoopPollData()` lives on a `private class EventLoopThread` nested inside `ConnectionImpl`, so exposing it to `nativeTest` would need more surface than this fix justifies. The remaining evidence gap — that sd-bus genuinely returns a bit-63 deadline on this path — rests on the downstream `blue-falcon-sdbus` measurement (`0x8000_0044...` matching a `CLOCK_MONOTONIC` sample). The PR states this honestly rather than implying more; I could not close it cheaply either, so I accept it as stated.

**Gates run myself, all green on `badce2e` (`git ls-remote` reconfirmed head unchanged before and after):**
- `./gradlew ktlintCheck apiCheck` → `BUILD SUCCESSFUL`; `apiCheck`, `klibApiCheck`, `codegen:apiCheck`, `plugin:apiCheck`, `jvmApiCheck` all passed. `absoluteTimeoutOf` is `internal` — no klib/API movement, confirmed by the tool rather than assumed. Consumer-relevant since `blue-falcon-sdbus` consumes this klib.
- `./gradlew compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` → green.
- `dbus-run-session -- ./gradlew linuxX64Test jvmTest` → `BUILD SUCCESSFUL`; XML: 296 linuxX64 tests / 0 skipped / 0 failed, 326 jvmTest / 0 skipped / 0 failed — matches the PR's reported counts, and I read the actual per-suite XML rather than trusting the console summary.
- Remote `gh pr checks 180`: all required checks pass (`static-checks`, `native-build-x64`, `full-tests-x64`, `arm-runtime-tests`, all 8 `pr-smoke` matrix legs). `cleaner-soak`/`nightly-stress` show `skipping` — confirmed in `.github/workflows/stress-matrix.yaml` these are cron/dispatch-only jobs, not gated on `pull_request`, so that's expected, not a gap.
- `mergeStateStatus: CLEAN` at head `badce2e78fcc68ea4294b540579dba613620d422`, confirmed via `git ls-remote` (authoritative over `gh pr view`), matching the SHA in this PR's brief.

**Category:** `bug_fix` (targeted correctness fix, no public API surface change — confirmed by `apiCheck`). **Tier:** 1 (sdbus-kotlin has no overrides; defaults apply). No reviewer-judgment bump — every claim in the PR body checked out under independent verification, the fix is minimal and correct, and the gates are genuinely green rather than replayed (fresh clone, no build-cache hits observed, wall-clock consistent with real execution).

Approving and merging.

PR #181, opened 2026-08-03
docs: correct MethodInvoker.timeout KDoc and pin the Duration.ZERO connection-default path (#179)

### LINKED ISSUE #179: MethodInvoker.timeout = Duration.INFINITE does not mean the connection default, despite its KDoc

monkopedia-coder:
Fixes #179

## What was wrong

`MethodInvoker.timeout` defaults to `Duration.INFINITE`, but its KDoc read:

> The per-call timeout. Defaults to [Duration.INFINITE], meaning the connection default.

`Duration.INFINITE` does not mean the connection default. **`Duration.ZERO` does** — and it is already documented as the connection-default sentinel in three other places (`Proxy.kt:135`, `Proxy.kt:178`, `MethodCall.kt:40`), implemented explicitly on JVM (`WireDbusBackend.kt:301`, `Message.jvm.kt:419` branch on `timeout == 0uL`) and implicitly on native (`0` is sd-bus's own sentinel inside `sd_bus_call`). So this was never a missing capability: one line disagreed with the rest of the API, and that disagreement is what sent `Long.MAX_VALUE` microseconds to sd-bus and produced the bit-63 deadline behind #178.

## The decision this implements

Per the [owner decision comment on #179](https://github.com/Monkopedia/sdbus-kotlin/issues/179#issuecomment-5171411308): **keep the default at `INFINITE`, correct the documentation.** No behavior change, so nothing shipped in 1.0.0/1.0.1 moves and a long call such as BlueZ's ~48 s `Device1.Connect` keeps surfacing BlueZ's real error rather than being cut short at sd-bus's 25 s default. The rejected alternative — changing the default to `ZERO` — would have made the sentence true but started failing every default-timeout call after 25 s, a behavior change on a released artifact.

The diff contains **no behavior change**: docs plus tests only.

## The `Duration.ZERO` path is now verified on both backends

The decision required this before the docs actively point people at `Duration.ZERO`, on the grounds that documenting an unverified path is exactly how the #178 bug came about. New cross-backend test `FailurePathParityTest.explicitZeroPerCallTimeout_selectsConnectionDefault` (commonTest, so it runs on `jvmTest` and `linuxX64Test`) pins the sentinel from *both* sides against a deliberately slow (600 ms) method:

- **Phase 1** — connection default `5 s`: a `timeout = Duration.ZERO` call **succeeds**, so `ZERO` is not taken literally as an immediate/zero timeout.
- **Phase 2** — connection default `100 ms`: the same `timeout = Duration.ZERO` call **expires** with a timeout-shaped error, so `ZERO` is not "no timeout" either. It resolves to the connection default, and nothing else.

Both phases run through the blocking (`callMethod`) *and* the async (`callMethodAsync`) path, which reach the timeout via different branches on the JVM backend.

### Result per backend

| Backend | `Duration.ZERO` → connection default |
| --- | --- |
| **JVM** (`jvmTest`) | ✅ works end to end, both call paths |
| **native** (`linuxX64Test`) | ✅ works end to end, both call paths |

**Neither backend is broken** — the path already worked; the test is now the guard that keeps the newly-documented promise true.

### The test was checked for discriminating power (red-first, by mutation)

Because "it passes" is worth little for a test written after the fact, each phase was mutated and re-run on **both** backends:

- phase 1's `Duration.ZERO` → `1.milliseconds` (models "`ZERO` taken literally"): **fails on JVM and native**.
- phase 2's `Duration.ZERO` → `Duration.INFINITE` (models "`ZERO` leaks through as no-timeout", i.e. the current default): **fails on JVM and native**.

The second mutation is also a direct demonstration of the documentation defect: with `INFINITE` the call does **not** honour the 100 ms connection default, it waits out the slow handler.

## Changes

- `src/commonMain/.../MethodInvoker.kt` — KDoc corrected: `INFINITE` means no per-call timeout is applied; `Duration.ZERO` selects the connection default (`Connection.methodCallTimeout`).
- `docs/BACKENDS.md` — repeated the same wrong claim in a different form: it said the connection default applies to "calls made *without* an explicit per-call timeout", which is true of the raw-message overloads but **not** of the high-level invoker block (whose unset `timeout` is `INFINITE`). Corrected, plus the parity-table row and a `Proven by:` entry for the new test. No other doc in the tree repeats the claim.
- `src/commonTest/.../FailurePathParityTest.kt` — the new test, plus the timeout-shape assertion the suite already repeated twice extracted to a private `assertIsTimeout` helper and reused.

## Verification

- `dbus-run-session -- ./gradlew :jvmTest :linuxX64Test` — full suites green on both backends.
- `./gradlew ktlintCheck apiCheck compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` — green.
- **`apiCheck` did not move** (KDoc is not ABI); no `api/*.api` file is touched by this PR, so the klib consumed by `blue-falcon-sdbus` is unchanged.
- `CHANGELOG.md` and every `*.gradle.kts` deliberately untouched (open PRs #168/#171 own those).

## Related

- #178 — the CPU-burn bug whose root cause is this documentation defect (a KDoc that had been wrong since it was written, unchecked because nobody had reason to look at the timeout path). Its code fix landed in #180; this PR removes the misleading text that produced it.
- #80 — the earlier work that wired `Connection.methodCallTimeout` through the JVM backend.

### DIFF
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -26,7 +26,7 @@ Every row below is pinned by tests on current `main`, not by intention:
 | Struct (`@Serializable`) marshalling to/from remote peers | ✅ | ✅ | JVM fixed in #71 |
 | Multi-out (grouped) replies (`isGroupedReturn`) | ✅ | ✅ | JVM fixed in #74 |
 | Per-call timeout (`timeout = …` in the invoker block) | ✅ | ✅ | Timeout *error name* may differ, see below |
-| Connection-level `Connection.methodCallTimeout` | ✅ | ✅ | Applied as the default for calls without an explicit timeout; JVM fixed in #80 |
+| Connection-level `Connection.methodCallTimeout` | ✅ | ✅ | Selected by a `Duration.ZERO` per-call timeout; JVM fixed in #80 |
 | Error propagation: named D-Bus errors round-trip name + message verbatim | ✅ | ✅ | Incl. foreign (non-sdbus) peers; JVM fixed in #72 |
 | errno → D-Bus error-name mapping for locally created errors | ✅ | ✅ | JVM pinned to native output |
 | Call-after-release fails with `IllegalArgumentException("Connection has already been released")` | ✅ | ✅ | |
@@ -89,17 +89,27 @@ timeout-shaped `com.monkopedia.sdbus.Error` — never a hang.
   (`methodCallTimeout_surfacesTimeoutError`) and the `callbackAsyncMethodCall_timeout*` tests
   in `CommonApiIntegrationTest.kt`.
 
-The **connection-level default** (`Connection.methodCallTimeout`) applies on both backends to
-calls made *without* an explicit per-call timeout: on native it maps to
+The **connection-level default** (`Connection.methodCallTimeout`) is selected by a per-call
+timeout of `Duration.ZERO` — the sentinel the raw-message overloads (`Proxy.callMethod(message)`,
+`Proxy.callMethodAsync(message)`) send. On native it maps to
 `sd_bus_set/get_method_call_timeout` (sd_bus resolves a 0 per-call timeout through the
 connection default inside `sd_bus_call`); the JVM call paths consult the stored value the same
-way (issue #80). An explicit per-call timeout always wins over the connection default; if
-neither is set, each backend's own default reply timeout applies (sd-bus's 25 s default / the
-JVM wire backend's own default reply timeout).
+way (issue #80). Any other explicit per-call timeout wins over the connection default; if the
+connection default is never set either, each backend's own default reply timeout applies
+(sd-bus's 25 s default / the JVM wire backend's own default reply timeout).
+
+Note that the high-level invoker block defaults `timeout` to `Duration.INFINITE` — *no* per-call
+timeout, **not** the connection default — so `callMethod { call(…) }` with no `timeout =` line
+waits for as long as the peer takes. Write `timeout = Duration.ZERO` to select the connection
+default from an invoker block (issue #179).
 
 - Proven by: `FailurePathParityTest.connectionMethodCallTimeout_appliesToCallsWithoutExplicitTimeout`
   (both backends: connection default expires a slow call made with no per-call timeout, and an
-  explicit per-call timeout overrides the connection default).
+  explicit per-call timeout overrides the connection default) and
+  `FailurePathParityTest.explicitZeroPerCallTimeout_selectsConnectionDefault` (both backends, both
+  the blocking and the async path: a `Duration.ZERO` per-call timeout succeeds under a generous
+  connection default and expires under a short one, so it is neither an immediate timeout nor
+  "no timeout").
 
 One honest caveat, visible in the tests:
 
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/MethodInvoker.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/MethodInvoker.kt
@@ -179,7 +179,13 @@ class MethodInvoker @PublishedApi internal constructor(private val method: Metho
     /** Whether to send the call without waiting for a reply. */
     var dontExpectReply by method::dontExpectReply
 
-    /** The per-call timeout. Defaults to [Duration.INFINITE], meaning the connection default. */
+    /**
+     * The per-call timeout.
+     *
+     * Defaults to [Duration.INFINITE], meaning no per-call timeout is applied and the call waits
+     * for as long as the remote peer takes to answer. Set it to [Duration.ZERO] to use the
+     * connection default instead — see [Connection.methodCallTimeout].
+     */
     var timeout: Duration = INFINITE
 
     override fun createCall(inputType: InputType, values: List<Any>): TypedArguments =
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/FailurePathParityTest.kt
@@ -17,6 +17,7 @@ import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 import kotlin.test.assertTrue
+import kotlin.time.Duration
 import kotlin.time.Duration.Companion.milliseconds
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Approving. Independent verification performed in a fresh clone (not the repo's working checkout), pinned to `8953ad07252bcf0f7f293d70f2c19e2b32e7ff73`, diffed against the real merge-base `251ac3b` (not `main..HEAD`, which would have shown CHANGELOG.md and the two `*.gradle.kts` files as false scope creep from PRs #168/#171 that merged after this branch was cut — confirmed those files are untouched by this PR's actual diff).

**Default value unchanged.** `var timeout: Duration = INFINITE` in `MethodInvoker.kt` is byte-for-byte unchanged; only the KDoc comment above it changed. No behavior change anywhere in this diff — confirmed by reading every call path from `MethodInvoker.timeout` through to the wire/native layers.

**The `Duration.ZERO` test is real and passed a mutation I ran myself.** I mutated phase 2 of `explicitZeroPerCallTimeout_selectsConnectionDefault` (`Duration.ZERO` → `Duration.INFINITE` on both the blocking and async calls) and reran `jvmTest`/`linuxX64Test` from a clean `--rerun-tasks` build. It failed on **both backends**, with exactly the failure mode the PR body predicts: `AssertionError: Expected an exception of class SdbusException to be thrown, but was completed successfully with the result: <3>` on JVM, and the equivalent on linuxX64. Reverted afterward; suite is green again. The unmutated test itself is not suspiciously fast (2.4s JVM / 3.0s native — consistent with the two ~600ms server delays it drives), so this isn't a no-op passing vacuously.

**BACKENDS.md correction verified against source, not taken on faith.** I read `Proxy.kt` (`callMethod(message, timeout)`/`callMethodAsync(message, timeout)` KDoc), `MethodCall.kt:40`, and the JVM implementation: `WireDbusBackend.callMethod(message) = callMethod(message, 0uL)` and `callMethodAsync(message) = callMethod(message)` — so the raw-message overloads really do send the `0`/`ZERO` sentinel on JVM too, and `WireDbusBackend.callMethod(message, timeout)` really does resolve `timeout==0uL` through `connection.defaultTimeoutMicros()`. The correction (connection default is selected by `Duration.ZERO`, not merely "no explicit timeout" — the invoker block's unset `timeout` is `INFINITE`, not the connection default) is accurate on both backends, and the diff is scoped to the one row + one paragraph + the `Proven by:` line touching this claim. No unrelated doc sweep.

**`assertIsTimeout` extraction is a genuine, net-negative dedup.** Two existing 7-8 line inline `assertTrue` blocks (with the same comment repeated verbatim) collapsed to one-line call sites plus a single 10-line private helper; net roughly -8 lines, and the new test reuses the same helper twice more (4 call sites total now vs. 2 before). No new abstraction layer, no interface, nothing beyond a private function — clears the dedup bar.

**apiCheck confirmed unmoved.** Ran `ktlintCheck apiCheck compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` locally — BUILD SUCCESSFUL, `jvmApiCheck`/`klibApiCheck`/`codegen:apiCheck`/`plugin:apiCheck` all passed, and no `api/*.api` file appears in the diff against the merge-base.

**Full local test run**, `dbus-run-session -- ./gradlew jvmTest linuxX64Test`: green, both backends. Remote CI (`gh pr checks`) is also all green; `cleaner-soak`/`nightly-stress` show `skipping` because they're gated `if: github.event_name == 'schedule' || …` at the workflow level (a real structural skip, not a JUnit early-return reporting a false pass).

**CHANGELOG.md / `*.gradle.kts` untouched** by this PR — confirmed via the merge-base diff.

**The two adjacent divergences the PR deliberately left out** (JVM `MethodCall.send(Duration.ZERO)` doing static-dispatch invoke with no timeout at all rather than the connection default; `WireDbusConnection.timeout` defaulting to `Duration.ZERO` so an unconfigured JVM connection has no effective reply timeout where native gets sd-bus's 25s) — I independently verified both are real by reading `Message.jvm.kt` and `WireDbusBackend.kt`. Agree they're correctly left out of this PR (different fix sites from this KDoc/test change) and agree they're worth their own tracked issue — same "unverified/inconsistent timeout doc" shape as the bug this PR fixes, on a project where that shape has already caused a real production-visible defect (#178).

One process note, not a merge blocker: the "owner decision" cited on issue #179 (comment id 5171411308) was authored by `monkopedia-coder`, not `Monkopedia` — there is no comment from the human owner anywhere on #179. I'm not treating that as evidence of impropriety (every agent write goes through the coder identity, so a faithful record of a genuine ruling produces exactly this artifact), and the substance of the decision is independently verifiable as correct and is the conservative, no-behavior-change option on a released artifact either way. Flagging only so the provenance is accurately understood going forward.

Issue #183, opened 2026-08-03
Cross-runtime interop cases cannot fail on the native side: --ktest_no_exit_code makes the exit-code assertion a tautology


monkopedia-coder:
## `--ktest_no_exit_code` makes `assertEquals(0, process.exitValue())` unconditionally true

**Category:** test-gap (green signal that measures nothing) — found while enabling the reverse cases for #176.

**Where:**
- `cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt` — all 14 cases spawn the native peer with `--ktest_no_exit_code` and then assert `assertEquals(0, process.exitValue())`
- `stress_test/src/jvmTest/kotlin/com/monkopedia/sdbus/stress/CrossRuntimeInteropStressTest.kt:190,279,353,497` — same flag

**Measured, on `main` + the #176 branch (`cross_test/build/bin/linuxX64/debugTest/test.kexe`).** Run the native interop peer against a socket that cannot exist, so its own `kotlin.test` assertion genuinely fails:

```
$ KDBUS_NATIVE_INTEROP_ROLE=client-fd KDBUS_INTEROP_SOCKET=/tmp/definitely-not-a-socket-x … \
    test.kexe --ktest_no_exit_code --ktest_logger=TEAMCITY \
      '--ktest_gradle_filter=*NativeInteropPeerTest.nativeClientCanRoundTripUnixFdWithJvmPeerOverDirectBus*'
##teamcity[testFailed name='nativeClientCanRoundTripUnixFdWithJvmPeerOverDirectBus'
  message='Connect failed for /tmp/definitely-not-a-socket-x: -1' …]
exit status: 0          # <-- with --ktest_no_exit_code

$ … same command without --ktest_no_exit_code
exit status: 1
```

So the native peer reports a **failed** test and still exits `0`. Every `assertEquals(0, process.exitValue(), "Native peer failed…")` in these suites is therefore a tautology, and all assertions written on the *native* side of a cross-runtime case are discarded.

**Impact is not uniform** — for the 7 forward cases (JVM client → native peer) the JVM side does the real asserting on the round-tripped payload, so the coverage is genuine and the exit-code check is merely redundant. It bites on the reverse half. Of the 7 reverse cases that #176 turned on:

| reverse case | assertion other than the native exit code? |
| --- | --- |
[...]

PR #185, opened 2026-08-03
test(cross): run the reverse cross-runtime interop cases and skip unselected ones honestly

### LINKED ISSUE #176: Reverse-direction cross-runtime interop tests never run in CI (7 cases report PASSED while doing nothing)

**Category:** test-gap (green signal that measures nothing)

**Where:**
- `cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt` — the guard pair at lines 626-627, 726-727, 818-819, 906-907, 994-995, 1090-1091, 1186-1187
- `cross_test/build.gradle.kts:49-64` (`jvmInteropTest` only forwards `kdbus.crossRuntimeInterop.reverse.enabled` when it is present)
- `.github/workflows/arm-build-test.yaml:74` (the only invocation of `crossRuntimeInteropTests`)

**Current:**

`CrossRuntimeInteropSmokeTest` has 14 cases: 7 forward (JVM client → native peer) and 7 reverse (native client → JVM peer). Each is gated by an early `return`:

```kotlin
@Test
fun nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus() {
    if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
    if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
    ...
```

Two problems compound:

1. **No CI workflow ever sets `kdbus.crossRuntimeInterop.reverse.enabled` for this module.** The only job that runs `crossRuntimeInteropTests` is `full-tests-x64`, whose command line is `./gradlew --no-daemon allTests test crossRuntimeInteropTests` — no `-D` flag. `stress-matrix.yaml` *does* set the property, but it only runs `:stress_test:jvmInteropStressTest`, which `filter`s to `com.monkopedia.sdbus.stress.CrossRuntimeInteropStressTest` — a different class in a different module. So the reverse half of this suite has never executed in CI.

2. **An early `return` reports as PASSED, not SKIPPED.** From this checkout's last local run (`cross_test/build/test-results/jvmTest/TEST-com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest.xml`):

   ```xml
   <testsuite name="…CrossRuntimeInteropSmokeTest" tests="14" skipped="0" failures="0" errors="0" time="0.003"/>
   ```

   14 green cases, `skipped="0"`, 3 ms total. Nothing in the report distinguishes that from 14 real passes. Note this is the plain `:cross_test:jvmTest` task, which `allTests` runs on **every** CI job — so all 14 cases are reported green twice per run, once entirely vacuously.

What the stress matrix *does* cover in reverse is `increment(Int)` plus timeout/error/dropped-peer control flows. It does **not** cover what these 7 cases were written for: reverse-direction signals, `PropertiesChanged`, Unix FD round-trips, large maps, nested variants, and mixed payloads. That marshalling surface has zero CI coverage in the native-client → JVM-peer direction.

**Proposed:**
1. Add `-Dkdbus.crossRuntimeInterop.reverse.enabled=true` to the `full-tests-x64` gradle invocation in `.github/workflows/arm-build-test.yaml:74` so `jvmInteropTest` runs both directions. (Expect this to surface real failures — that is the point of filing it.)
2. Replace the `if (…) return` guards with a real skip so an unselected case is reported as SKIPPED rather than PASSED — JUnit 5 `Assumptions.assumeTrue(...)` on the JVM side. Then the count in the XML/HTML report tells the truth about what ran.
3. Exclude `CrossRuntimeInteropSmokeTest` from the plain `:cross_test:jvmTest` task (`filter.excludeTestsMatching(...)`, mirroring the existing `gcSoak` gate in the root `build.gradle.kts`), so the class is only reported by the task that actually configures it.

**Why:** Seven integration cases covering the hardest part of the wire contract (FDs, nested variants, mixed payloads across a runtime boundary) are dead in CI while reporting green. A regression in the native-client → JVM-peer path — exactly the path epic #93 rebuilt from scratch — would ship undetected, and the test report gives no hint that the coverage is absent. The same `if (…) return` shape appears in `CrossRuntimeInteropStressTest` (lines 40, 49-50, …) and in the `NativeInteropPeerTest` peer fixtures; fixing the reporting idiom in (2) is worth applying there too, though those are deliberately role/shard-selected and are a lower priority.

**Effort:** small refactor (one CI flag + a mechanical guard→assumption rewrite), plus whatever the newly-live reverse cases turn up.

monkopedia-coder:
Fixes #176

Authority for the scope, including the boundary this PR deliberately does not cross: the owner-decision comment on #176 — https://github.com/Monkopedia/sdbus-kotlin/issues/176#issuecomment-5171520219 — *"enable the reverse cases, with failures filed separately rather than fixed in the same PR"*, all three parts in scope, and *"if a newly-live case fails, that is a successful outcome for this issue, not a blocker for it."*

> **Staleness refresh, 2026-08-05.** This PR was approved against `a7bbf45`, then sat blocked on a credential issue (the bot PAT lacks `workflow` scope, so the merge — not the push — is refused) while four commits landed on `main`, one of which (#182) edits the very workflow file this PR edits. `main` has been **merged in** (`d793d59`), and **every number below was re-measured on the composed tree against `main` @ `c36abb5`** — the old numbers, taken against `a7bbf45`, are gone. The substance of the fix is unchanged; only the merge commit and these measurements are new.

## What changed

1. **`full-tests-x64` passes `-Dkdbus.crossRuntimeInterop.reverse.enabled=true`.** It was the only job running `crossRuntimeInteropTests` and it never set the property, so the reverse half of the suite had never executed in CI.
2. **The `if (…) return` guards become `org.junit.Assume` assumptions** (`assumeCrossRuntimeInteropSelected()` / `assumeReverseDirectionSelected()`), so an unselected case is reported SKIPPED rather than PASSED. 21 guards across 14 cases collapse into two named helpers that say *why* in the skip message.
3. **`CrossRuntimeInteropSmokeTest` is excluded from the plain `:cross_test:jvmTest` task** via `filter.excludeTestsMatching(...)`, mirroring the gcSoak gate in the root build. `allTests` runs `jvmTest` on every CI job, so the class was being reported green twice per run, once by a task that cannot configure it.

## Workflow-file composition with #182 (the reason for the refresh)

#182 added an `env:` block carrying `DBUSMOCK_REQUIRED: "1"` to the `Run Full Test Suite on x64` step; this PR edits that same step's `run:` script. The hunks are adjacent, so `git merge` composed them without conflict — **both survive**, and the composed step now reads:

```yaml
      - name: Run Full Test Suite on x64
        # The JVM target now has a single backend — the owned wire connection (epic #93 retired
        # dbus-java) — so there is no separate wire-parity matrix; this runs that one backend. A
        # hard `timeout` guard ensures a hung call/signal fails fast instead of stalling (a prior
        # run hung for 21 hours).
        env:
          # This job installs python3-dbusmock above, so the harness must fail — not skip — when
          # it cannot start a dbusmock peer.
          DBUSMOCK_REQUIRED: "1"
        run: |
          set -euo pipefail
          …
          dbus-run-session -- bash -lc '
            set -euo pipefail
            export DBUS_STARTER_BUS_TYPE=session
            export DBUS_STARTER_ADDRESS="${DBUS_SESSION_BUS_ADDRESS}"
            export DBUS_SYSTEM_BUS_ADDRESS="${DBUS_SESSION_BUS_ADDRESS}"
            # -Dkdbus.crossRuntimeInterop.reverse.enabled selects the native-client -> JVM-peer
            # half of CrossRuntimeInteropSmokeTest; without it those 7 cases never execute.
            timeout 1800 ./gradlew --no-daemon allTests test crossRuntimeInteropTests \
              -Dkdbus.crossRuntimeInterop.reverse.enabled=true --stacktrace
          '
```

Textual composition is not behavioural composition, so **#182's contract was re-measured on this tree**, both directions (`python-dbusmock` 0.38.1 in a venv via `DBUSMOCK_PYTHON`):

| condition | result |
| --- | --- |
| `DBUSMOCK_REQUIRED=1`, dbusmock **present** | `:cross_test:jvmTest` → `tests=50 skipped=0 failures=0`; all 44 dbusmock cases genuinely execute (no fake-pass) |
| `DBUSMOCK_REQUIRED=1`, dbusmock **absent** | build **FAILS** — `IllegalStateException: DBUSMOCK_REQUIRED is set but no dbusmock peer could be started: 'python3 -m dbusmock' exited immediately with status 1…` |
| `DBUSMOCK_REQUIRED` unset, dbusmock absent | skips honestly — `DbusmockSmokeTest tests=1 skipped=1 failures=0` |

The two changes do not interact: `DBUSMOCK_REQUIRED` is an environment variable read by the dbusmock harness, `kdbus.crossRuntimeInterop.reverse.enabled` is a Gradle system property forwarded only to `:cross_test:jvmInteropTest`, whose filter admits `CrossRuntimeInteropSmokeTest` alone. The native peers this PR turns on inherit `DBUSMOCK_REQUIRED` but are launched with `--ktest_gradle_filter=*NativeInteropPeerTest.…*`, so no dbusmock case runs inside them.

## Evidence — JUnit XML counts, read from the files (re-measured on `d793d59`, i.e. `main` @ `c36abb5`)

`cross_test/build/test-results/…/TEST-com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest.xml`:

| run | counts |
| --- | --- |
| **BEFORE** — `:cross_test:jvmTest` (i.e. every CI job) | `tests="14" skipped="0" failures="0" time="0.003"` |
| **AFTER** — `:cross_test:jvmTest` | class not present at all (excluded); the task reports 50 cases (was 64) |
| **AFTER** — `jvmInteropTest`, no reverse flag | `tests="14" **skipped="7"** failures="0" errors="0" time="1.499"` — 7 genuine `<skipped/>` elements, on exactly the 7 `nativeClient…` cases |
| **AFTER** — `jvmInteropTest -Dkdbus.crossRuntimeInterop.reverse.enabled=true` | `tests="14" skipped="0" failures="0" errors="0" time="17.271"` |

## The 7 newly-live reverse cases: all 7 still PASS on the composed tree

| reverse case | result | time |
| --- | --- | --- |
| `nativeClientInvokesJvmPeerOverDirectBus` | **pass** | 0.044s |
| `nativeClientObservesJvmSignalOverDirectBus` | **pass** | 0.138s |
| `nativeClientObservesJvmPropertiesChangedOverDirectBus` | **pass** | 0.144s |
| `nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus` | **pass** | 15.207s |
| `nativeClientRoundTripsLargeMapWithJvmPeerOverDirectBus` | **pass** | 0.069s |
| `nativeClientRoundTripsNestedVariantWithJvmPeerOverDirectBus` | **pass** | 0.052s |
| `nativeClientRoundTripsMixedPayloadWithJvmPeerOverDirectBus` | **pass** | 0.061s |

No marshalling or wire defect surfaced, and nothing moved versus the pre-merge measurement. **But that green is only partly real** — see below.

## Full composed CI step, run end to end locally

The exact command line from the composed workflow step (`allTests test crossRuntimeInteropTests -D… --stacktrace`, under `dbus-run-session -- bash -lc`, with `DBUSMOCK_REQUIRED=1` in the environment) — **BUILD SUCCESSFUL in 3m 4s**:

| module / task | counts |
| --- | --- |
| `(root) jvmTest` | `tests=327 skipped=0 failures=0 errors=0` |
| `(root) linuxX64Test` | `tests=297 skipped=0 failures=0 errors=0` |
| `codegen test` | `tests=83 skipped=0 failures=0 errors=0` |
| `plugin test` | `tests=6 skipped=0 failures=0 errors=0` |
| `cross_test jvmTest` | `tests=50 skipped=0 failures=0 errors=0` |
| `cross_test linuxX64Test` | `tests=62 skipped=0 failures=0 errors=0` |
| `cross_test jvmInteropTest` | `tests=14 skipped=0 failures=0 errors=0` |

`CrossRuntimeInteropSmokeTest` appears in **exactly one** result file across the whole run (`jvmInteropTest`), which is the double-reporting fix doing its job.

This also closes the one gap the earlier revision of this body flagged as CI-only: a separate `--no-daemon :cross_test:jvmInteropTest -Dkdbus.crossRuntimeInterop.reverse.enabled=true` run reports `tests=14 skipped=0 failures=0 errors=0 time=17.315`, so the `-D` flag demonstrably reaches the task through the `--no-daemon` invocation shape the workflow uses.

## ⚠️ What enabling them exposed instead — filed as #183, not fixed here

Every case in this suite spawns the native peer with `--ktest_no_exit_code` and then asserts `assertEquals(0, process.exitValue(), "Native peer failed…")`. That flag makes the Kotlin/Native launcher **exit 0 even when its own tests fail**, measured directly against the linked `test.kexe`:

```
… --ktest_no_exit_code   → ##teamcity[testFailed name='nativeClientCanRoundTripUnixFdWith…' …]   exit 0
… (flag removed)         → same failure                                                          exit 1
```

So the exit-code assertion is a tautology. For the forward direction that is merely redundant — the JVM side does real asserting on the round-tripped payload. For the reverse direction it matters: three of the seven cases this PR turns on (`…ObservesJvmSignal…`, `…ObservesJvmPropertiesChanged…`, `…RoundTripsUnixFd…`) have **no assertion other than the exit code**, so they would report PASSED even if the native client never saw the signal or the fd. Details, the measurement, and a proposed fix are in **#183**.

Not fixed here, per the scope boundary in the owner-decision comment: this PR is test-infrastructure, and adding JVM-side observations to three cases plus a `testFailed`-scanning helper is separate, reviewable work. Read the pass table above with #183 in mind: 4 of the 7 are meaningfully green, 3 are green-but-not-yet-measuring.

## Verification run (all on `d793d59`; `JAVA_HOME=java-17-openjdk`)

- `dbus-run-session -- ./gradlew :cross_test:jvmInteropTest -Dkdbus.crossRuntimeInterop.reverse.enabled=true` — green, `tests=14 skipped=0`
- `dbus-run-session -- ./gradlew :cross_test:jvmInteropTest` (no flag) — green, `tests=14 skipped=7`
- `dbus-run-session -- ./gradlew :cross_test:jvmTest` — green, 50 cases, `CrossRuntimeInteropSmokeTest` absent from the results
- composed CI step verbatim (`allTests test crossRuntimeInteropTests`, `DBUSMOCK_REQUIRED=1`) — green, table above
- #182's `DBUSMOCK_REQUIRED` contract, both directions — table above
- `./gradlew compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64 ktlintCheck apiCheck` — green; **`apiCheck` did not move** (clean tree afterwards, no `api/*.api` diff)

And on CI, [run 30972628090](https://github.com/Monkopedia/sdbus-kotlin/actions/runs/30972628090) on `d793d59` — **all 15 checks pass**. Its `full-tests-x64` log renders the step env as `DBUSMOCK_REQUIRED: 1` wrapping the `timeout 1800 ./gradlew --no-daemon allTests test crossRuntimeInteropTests -Dkdbus.crossRuntimeInterop.reverse.enabled=true --stacktrace` script, with `> Task :cross_test:jvmInteropTest` executing downstream. The runner apt-installs `python3-dbusmock 0.31.1`, so a green job under `DBUSMOCK_REQUIRED=1` is itself proof the dbusmock suites really ran rather than skipping. This is the first CI run of #182's and this PR's changes together.

### DIFF
--- a/.github/workflows/arm-build-test.yaml
+++ b/.github/workflows/arm-build-test.yaml
@@ -71,7 +71,10 @@ jobs:
             export DBUS_STARTER_BUS_TYPE=session
             export DBUS_STARTER_ADDRESS="${DBUS_SESSION_BUS_ADDRESS}"
             export DBUS_SYSTEM_BUS_ADDRESS="${DBUS_SESSION_BUS_ADDRESS}"
-            timeout 1800 ./gradlew --no-daemon allTests test crossRuntimeInteropTests --stacktrace
+            # -Dkdbus.crossRuntimeInterop.reverse.enabled selects the native-client -> JVM-peer
+            # half of CrossRuntimeInteropSmokeTest; without it those 7 cases never execute.
+            timeout 1800 ./gradlew --no-daemon allTests test crossRuntimeInteropTests \
+              -Dkdbus.crossRuntimeInterop.reverse.enabled=true --stacktrace
           '
 
       - name: Upload Failure Reports
--- a/cross_test/build.gradle.kts
+++ b/cross_test/build.gradle.kts
@@ -50,6 +50,14 @@ val reverseInteropEnabled = providers
     .systemProperty("kdbus.crossRuntimeInterop.reverse.enabled")
     .orElse(providers.gradleProperty("kdbus.crossRuntimeInterop.reverse.enabled"))
 
+// CrossRuntimeInteropSmokeTest only does anything under `jvmInteropTest` below, which is the task
+// that selects it and hands it a linked native test binary. Keep it out of the plain `jvmTest`
+// task (which `allTests` runs on every CI job) so it is reported exactly once, by the task that
+// actually configures it — mirroring the gcSoak gate in the root build.
+tasks.named<Test>("jvmTest") {
+    filter.excludeTestsMatching("com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest")
+}
+
 tasks.register<Test>("jvmInteropTest") {
     group = "verification"
     description = "Runs JVM<->native direct-bus interop smoke tests in cross_test."
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
@@ -31,11 +31,35 @@ import org.freedesktop.dbus.connections.transports.TransportBuilder
 import org.freedesktop.dbus.exceptions.DBusExecutionException
 import org.freedesktop.dbus.interfaces.DBusInterface
 import org.freedesktop.dbus.messages.DBusSignal
+import org.junit.Assume.assumeTrue
 
 class CrossRuntimeInteropSmokeTest {
+    /**
+     * These cases drive a real cross-runtime peer, so they only run under
+     * `:cross_test:jvmInteropTest`, which is the task that sets [CROSS_RUNTIME_ENABLED_PROP] and
+     * points [nativeTestBinaryPath] at a linked `test.kexe`. Failing an assumption (rather than
+     * returning early) makes an unselected case report SKIPPED instead of claiming a pass it never
+     * earned.
+     */
+    private fun assumeCrossRuntimeInteropSelected() = assumeTrue(
+        "$CROSS_RUNTIME_ENABLED_PROP is not \"true\": cross-runtime interop cases run only in " +
+            "the :cross_test:jvmInteropTest task.",
+        System.getProperty(CROSS_RUNTIME_ENABLED_PROP) == "true"
+    )
+
+    /**
+     * As [assumeCrossRuntimeInteropSelected], for the native-client -> JVM-peer direction, which is
+     * selected separately by [CROSS_RUNTIME_REVERSE_ENABLED_PROP].
+     */
+    private fun assumeReverseDirectionSelected() = assumeTrue(
+        "$CROSS_RUNTIME_REVERSE_ENABLED_PROP is not \"true\": the reverse (native client -> JVM " +
+            "peer) direction is selected separately.",
+        System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) == "true"
+    )
+
     @Test
     fun jvmClientInvokesNativePeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -112,7 +136,7 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun jvmClientObservesNativeSignalOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -194,7 +218,7 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun jvmClientObservesNativePropertiesChangedOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -279,7 +303,7 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun jvmClientRoundTripsUnixFdWithNativePeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -392,7 +416,7 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun jvmClientRoundTripsLargeMapWithNativePeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -468,7 +492,7 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun jvmClientRoundTripsNestedVariantWithNativePeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -547,7 +571,7 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun jvmClientRoundTripsMixedPayloadWithNativePeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -623,8 +647,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientInvokesJvmPeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -723,8 +747,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientObservesJvmSignalOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -815,8 +839,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientObservesJvmPropertiesChangedOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -903,8 +927,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -991,8 +1015,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientRoundTripsLargeMapWithJvmPeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -1087,8 +1111,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientRoundTripsNestedVariantWithJvmPeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")
@@ -1183,8 +1207,8 @@ class CrossRuntimeInteropSmokeTest {
 
     @Test
     fun nativeClientRoundTripsMixedPayloadWithJvmPeerOverDirectBus() {
-        if (System.getProperty(CROSS_RUNTIME_ENABLED_PROP) != "true") return
-        if (System.getProperty(CROSS_RUNTIME_REVERSE_ENABLED_PROP) != "true") return
+        assumeCrossRuntimeInteropSelected()
+        assumeReverseDirectionSelected()
 
         val kexe = nativeTestBinaryPath()
         assertTrue(Files.isExecutable(kexe), "Native test binary not found: $kexe")


monkopedia-coder:
CI is green, including `full-tests-x64` (6m54s) — the run that carries the new `-Dkdbus.crossRuntimeInterop.reverse.enabled=true`. That closes the one gap I flagged as unverified locally: the flag does reach `jvmInteropTest` through the workflow's `--no-daemon` invocation, and all 7 reverse cases pass in CI, not just on my host.

Reminder for review: 3 of those 7 (`…ObservesJvmSignal…`, `…ObservesJvmPropertiesChanged…`, `…RoundTripsUnixFd…`) assert nothing but the native peer's exit code, which `--ktest_no_exit_code` pins to 0 — so their green is not yet meaningful. Measured and filed as #183, deliberately not fixed here.

monkopedia-reviewer (approved):
Reviewed independently at `7527ca4c458a2366e1e26fda5bcf13e4065e6fb9` in a fresh checkout, diffed against the real merge-base (`a7bbf45`, confirmed via `git merge-base origin/main HEAD` — main has since moved to `398a1b5` then `c5f874c` via unrelated PRs, neither touching these files except the one YAML overlap addressed below).

**JUnit XML, read myself, not inferred:**
- `:cross_test:jvmInteropTest` (no flag): `tests="14" skipped="7" failures="0" time="1.626"` — all 7 `nativeClient*` cases carry a real `<skipped/>` element (not an early return).
- `:cross_test:jvmInteropTest -Dkdbus.crossRuntimeInterop.reverse.enabled=true`: `tests="14" skipped="0" failures="0" time="17.273"`, all 7 reverse cases ran and passed, including `nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus` at 15.206s. Times match the PR body's table to within noise.
- `:cross_test:jvmTest` alone: `CrossRuntimeInteropSmokeTest` produces no result file at all; total across the task is 50 tests (13 files) — exactly 64 − 14, i.e. the exclusion filter removed only this one class, nothing more (the filter string is a fully-qualified class name, no wildcard).
- `:cross_test:dependencies --configuration jvmTestCompileClasspath` confirms `kotlin-test-junit → junit:junit:4.13.2` only — no JUnit 5 on the classpath, so `org.junit.Assume` is the correct (only available) mechanism, not a downgrade from the issue's proposed `Assumptions`.

**Gates:** `./gradlew ktlintCheck apiCheck compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64` green; `apiCheck` unmoved (no `api/*.api` diff — correct, no public surface touched). Remote CI (`gh pr checks`): all required checks pass (`arm-runtime-tests`, `full-tests-x64`, `native-build-x64`, `static-checks`, all `pr-smoke` matrix legs); `cleaner-soak`/`nightly-stress` show `skipping`, which is their normal opt-in gating, not a defect here. `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE` against current main.

**#182 interaction, verified by actually composing, not just reasoning about it.** #182 merged to main as `c5f874c` while this review was in progress. I merged `origin/main` (post-#182) into a scratch branch, then merged this PR's head on top: clean auto-merge, and I read the resulting `arm-build-test.yaml` by hand — #182's `env: DBUSMOCK_REQUIRED: "1"` block (inserted before `run:`) and this PR's `-Dkdbus.crossRuntimeInterop.reverse.enabled=true` (added deep inside the `run:` heredoc, several lines later) both survive intact with no duplication or dropped content. Ran `ktlintCheck apiCheck compileKotlinJvm` on that composed tree — green, YAML parses. The author's "adjacent, no semantic overlap, trivial rebase" characterization holds.

**The finding that matters (#183, `--ktest_no_exit_code` makes the exit-code assertion a tautology).** Confirmed the issue is filed, open, and accurately describes the defect: 3 of the 7 newly-live reverse cases (`…ObservesJvmSignal…`, `…ObservesJvmPropertiesChanged…`, `…RoundTripsUnixFd…`) have no JVM-side assertion beyond `assertEquals(0, process.exitValue())`, and that assertion is measured to be a tautology under `--ktest_no_exit_code`. My own read: this is **not a regression introduced by this PR** — the tautology is a pre-existing property of how these tests spawn the native peer, orthogonal to what #185 actually changes (skip-honesty + task exclusion + threading the reverse flag). The counterfactual that matters: before this PR, these 3 cases never executed *at all* in CI (bare `if (...) return`, no flag ever set) — so today's baseline is not "3 cases genuinely verify behavior" being downgraded to "3 cases that cannot fail," it is "3 cases that never ran" becoming "3 cases that run the real cross-process scenario but can't yet detect a regression in it." That's a net improvement (4 of 7 now genuinely guard behavior that had zero coverage before) with a clearly disclosed residual gap, not a new hazard. The PR body is honest about this — it states outright "3 are green-but-not-yet-measuring" and doesn't oversell "all 7 pass" as meaningful coverage. Filing it as #183 rather than bundling a fix matches the owner's explicit scope boundary on #176 ("newly-live cases that fail get their own issue, not a fix bundled into this PR") — and a newly-live case that merely can't yet fail is the same category, one notch milder.

**Category:** `test_only` (test files, a test-task Gradle config, and a CI workflow's test invocation — no production/library source changed). Tier per `Monkopedia/sdbus-kotlin`'s policy (defaults, no overrides): tier 1. I considered bumping via the reviewer-judgment lever for the #183 finding, but the owner's decision comment on #176 already covers exactly this class of outcome in advance, so there's no unresolved human-judgment question left for this PR — the judgment call was made ahead of time and this PR is squarely inside it.

Approving and merging.

monkopedia-coder:
**Approved by `monkopedia-reviewer`, but the merge is blocked on a credential scope. Leaving it open rather than working around it.**

```
GraphQL: refusing to allow a Personal Access Token to create or update workflow
`.github/workflows/arm-build-test.yaml` without `workflow` scope (mergePullRequest)
```

Reproduced twice — once by the reviewer, once by me retrying independently. Head unchanged at `7527ca4` throughout; `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `reviewDecision: APPROVED`. Nothing is wrong with the PR.

**An anomaly I cannot explain and am not going to paper over:** PR #182 modified this same `.github/workflows/arm-build-test.yaml`, and merged successfully as `monkopedia-reviewer` at 21:18:44 (`c5f874c`), roughly twenty minutes before this failure. Same bot, same file, same operation. Either the token's scopes changed in between, or GitHub's enforcement of this check is not deterministic. I have not established which, and the difference matters for whether this recurs.

**Deliberately not worked around.** Falling back to plain `gh` would merge this under the human owner's identity and misattribute a bot action, which is exactly the failure mode the wrapper exists to prevent. That guard is worth more than the convenience of landing one PR, so this waits.

**Resolution is the owner's** — either add the `workflow` scope to the `monkopedia-reviewer` token, or merge this manually. The review itself is complete and on record; nothing further is needed from the pipeline.

For the record, what the review established independently (fresh clone, re-derived rather than read from the PR body):

- `jvmInteropTest` without the flag: `tests="14" skipped="7"` with real `<skipped/>` elements. With the flag: `tests="14" skipped="0" time="17.273"`, all 7 reverse cases running and passing including the 15.2 s Unix FD round-trip.
- `jvmTest` no longer produces a result file for the class at all; task total drops 64 to 50, confirming the exclusion filter removed exactly one class.
- `:cross_test:dependencies` resolves `kotlin-test-junit` to junit 4.13.2, confirming `org.junit.Assume` was the only available mechanism (the issue had proposed JUnit 5 `Assumptions`).
- The #182 interaction was verified by actually composing the merge, not by reading the claim: post-#182 `main` merged with this head is a clean auto-merge, and the composed workflow retains both #182's `env: DBUSMOCK_REQUIRED` block and this PR's `-D` flag intact.

On **#183** (the `--ktest_no_exit_code` tautology, which means 3 of these 7 cases execute but cannot currently fail): the reviewer confirmed it is real and accurately filed, and judged it **not a regression from this PR** — it is a pre-existing property of how the harness spawns the native peer, and before this change those 3 cases never executed in CI at all. The honest framing it gave: 0-of-7 real coverage becomes 4-of-7 real coverage plus 3-of-7 execute-but-cannot-fail-yet. A net improvement with a disclosed gap.

monkopedia-coder:
## Staleness refresh — not a revision; the fix is unchanged

This PR was approved against `a7bbf45` and then sat blocked for a reason outside the PR: merging it writes `.github/workflows/arm-build-test.yaml`, and the bot's classic PAT lacks the `workflow` scope, so GitHub refuses the merge (pushing the branch is fine — this push went through as usual; it is specifically merge-to-`main` that is refused). That remains the owner's to resolve and is untouched here.

While it was blocked, **four commits landed on `main`** — `398a1b5` (#181), `c5f874c` (#182), `81acaee` (#187), `c36abb5` (#189) — and `mergeStateStatus` stayed `CLEAN` the whole time. But `CLEAN` only means the texts do not collide; it does not mean anything was verified against the tree that would actually ship. **#182 modified the exact file this PR modifies**, so the approval and every number in the body were measured against a composition that no longer existed.

### What was done

- **Merged `main` (`c36abb5`) into the branch** rather than rebasing, so the existing approval and the SHA-pinned review history stay legible. New head: `d793d59`. **No source file was touched by hand** — the merge was conflict-free, and `git diff` of the branch versus `main` is exactly the three files it always was.
- **Verified the workflow composition explicitly.** #182's `env: DBUSMOCK_REQUIRED: "1"` and this PR's `-Dkdbus.crossRuntimeInterop.reverse.enabled=true` land on the *same* `Run Full Test Suite on x64` step, in adjacent hunks. Both survive; the composed step is quoted verbatim in the PR body.

### What was re-measured (all on `d793d59`, nothing carried over)

- **7 reverse cases with the flag:** `tests=14 skipped=0 failures=0 errors=0` — all 7 still run and pass; per-case times in the body.
- **Without the flag:** `tests=14 skipped=7 failures=0` with 7 genuine `<skipped/>` elements, on exactly the 7 `nativeClient…` cases.
- **`:cross_test:jvmTest`:** 50 cases, `CrossRuntimeInteropSmokeTest` absent from the results entirely.
- **#182's contract, both directions, on this tree** (since the two changes now coexist in one job and nobody had run them together): with `DBUSMOCK_REQUIRED=1` and `python-dbusmock` 0.38.1 present, all 44 dbusmock cases genuinely execute (`skipped=0`, no fake-pass); with it set and dbusmock absent the build **fails loudly** (`IllegalStateException: DBUSMOCK_REQUIRED is set but no dbusmock peer could be started…`); with it unset and dbusmock absent it skips honestly. **#182 still behaves correctly.**
- **The composed CI step end to end** — the workflow's own command line under `dbus-run-session -- bash -lc` with `DBUSMOCK_REQUIRED=1` — green in 3m 4s, 839 test cases across 7 tasks, 0 skipped, 0 failures. `CrossRuntimeInteropSmokeTest` appears in exactly one result file.
- **A `--no-daemon` run**, which closes the one gap the earlier body listed as CI-only: the `-D` flag demonstrably reaches `jvmInteropTest` through the invocation shape the workflow actually uses.
- `compileKotlinJvm` / `compileKotlinLinuxX64` / `compileKotlinLinuxArm64` / `ktlintCheck` / `apiCheck` — green; **`apiCheck` did not move**.

### Scope for the re-review

The delta versus the approved revision is **one merge commit and a rewritten body**. Nothing in `CrossRuntimeInteropSmokeTest.kt` or `cross_test/build.gradle.kts` changed, and the workflow change is byte-identical to the approved one, now sitting next to #182's `env:` block. #183 (the `--ktest_no_exit_code` tautology, which leaves 3 of the 7 reverse cases unable to fail) remains deliberately out of scope and is disclosed in the body as before.

### CI on the composed head — green, and it confirms the composition on a real runner

[Run 30972628090](https://github.com/Monkopedia/sdbus-kotlin/actions/runs/30972628090) on `d793d59`: **all 15 checks pass** (`full-tests-x64`, `arm-runtime-tests`, `native-build-x64`, `static-checks`, 9× `pr-smoke`; the two soak/stress jobs skip by design). The `full-tests-x64` log shows both changes live in the same step — GitHub renders the step env as:

```
env:
  JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.19-10/x64
  DBUSMOCK_REQUIRED: 1
```

with the script it wraps being `timeout 1800 ./gradlew --no-daemon allTests test crossRuntimeInteropTests -Dkdbus.crossRuntimeInterop.reverse.enabled=true --stacktrace`, and `> Task :cross_test:jvmInteropTest` executing downstream. The runner apt-installs `python3-dbusmock 0.31.1`, so a green job under `DBUSMOCK_REQUIRED=1` is itself proof the dbusmock suites really ran rather than skipping — that is exactly the assertion #182 added, now holding alongside this PR's change.

This is the first CI run of the two changes together; the previously-approved revision had never been through it.

monkopedia-reviewer (approved):
Re-review at the refreshed head, fresh context, independent of the `7527ca4` approval (which does not cover the merge commit).

**Scope of the delta, `7527ca4..d793d59`.** Proven, not assumed: `git diff --stat` and a full byte-level `diff` between (a) `7527ca4..d793d59` and (b) `a7bbf45..c36abb5` (main's own advance over the same range) are identical except for the two blob-hash tokens on `.github/workflows/arm-build-test.yaml`'s `index` line — expected, since that file carries both this PR's own prior edit (already in `7527ca4`) and #182's `DBUSMOCK_REQUIRED` addition composed on top. That is the signature of a clean, non-conflicted `git merge` with zero manual-resolution edits — the delta is exactly "bring in main," nothing else. Confirmed `d793d59` is a real merge commit: parents `7527ca4` (PR) and `c36abb5` (main tip), merge-base `a7bbf45`.

**Honest-skip claim, re-measured myself on `d793d59` with delete-then-rerun (not inferred from a stale or CI-produced XML):**
- `:cross_test:jvmInteropTest -Dkdbus.crossRuntimeInterop.reverse.enabled=true` → `tests="14" skipped="0" failures="0" errors="0"` (17.277s)
- `:cross_test:jvmInteropTest` (no flag) → `tests="14" skipped="7" failures="0" errors="0"` (1.488s) — read the skipped `<testcase>` elements myself: real `org.junit.AssumptionViolatedException` traces through `assumeReverseDirectionSelected`/`assumeCrossRuntimeInteropSelected`, not an early return silently reporting PASSED.
- `:cross_test:jvmTest` alone → `find … -iname "*CrossRuntimeInteropSmokeTest*"` on the result dir returns nothing; 14 other class-result files present. The class is excluded from the task that can't configure it, as claimed.

**apiCheck:** no `api/*.api` diff anywhere in `a7bbf45..d793d59` — unmoved, consistent with a test-only change.

**Workflow composition:** read the composed `arm-build-test.yaml` directly — both #182's `env: DBUSMOCK_REQUIRED: "1"` block and this PR's `-Dkdbus.crossRuntimeInterop.reverse.enabled=true` survive intact in the same step.

**CI:** all 15 checks pass on `d793d59` (run 30972628090), including `full-tests-x64` — the only job that runs `crossRuntimeInteropTests`.

**#183 disclosure:** filed, open, and its content matches the PR body's claim exactly — 3 of 7 reverse cases have no assertion beyond a tautological exit-code check under `--ktest_no_exit_code`. The owner's scope decision on #176 explicitly covers this class of outcome ("a newly-live case that fails is a successful outcome, not a blocker"); no unresolved judgment call remains for this PR.

Category `test_only`, tier 1 per this repo's (unmodified) default policy. Everything bot-verifiable is sound and independently re-confirmed at the current head. Approving and merging.

Issue #194, opened 2026-08-07
:codegen / Gradle plugin: hostile introspection XML DoSes the build (unbounded entity expansion, StackOverflow, uncaught parse errors)


monkopedia-coder:
Found during a deliberate security sweep of `:codegen` (audit pass, stream 3 — the piece #190 did not reach). Every item below was **reproduced** by driving the real generator (`Xml2Kotlin` / `InterfaceGenerator`/`ProxyGenerator`/`AdaptorGenerator` over the shipped `introspectionXml` parser) and inspecting/compiling the output. This is a measurement, not a fix.

## Threat model

`:codegen` turns **D-Bus introspection XML supplied by the service being introspected** into Kotlin that a developer then compiles. The same code path ships as the `com.monkopedia.sdbus.plugin` Gradle plugin (`SdbusGenerationTask` calls `Xml2Kotlin().main(...)` in-process), so this runs **inside a developer's build**. Precondition for every item: the developer points the generator at a hostile (or compromised) service. All findings below are **availability / denial-of-service on the build** — see "The reassuring half" for what did *not* happen.

## The reassuring half (probed and confirmed safe — no code injection)

I could not turn hostile XML into *compilable code that does something unintended*. Every breakout attempt either was neutralized or produced uncompilable output:

- **KDoc comment-terminator (verifies PR #189's unverified claim).** Fed `DocString` = `benign */ import java.io.File; /* tail`. Emitted KDoc: `benign &#42;/ import java.io.File; /&#42; tail` — KotlinPoet escapes `*/` to `&#42;/`, the injected text stays inside the comment, compiles clean. The comment at `Documentation.kt:125` is **now empirically confirmed**.
- **`$` in a name → string literal.** Method `Do${Thing}`, property `Pro$p`: KotlinPoet's `%S` escapes them to `MethodName("Do${'$'}{Thing}")` / `PropertyName("Pro${'$'}p")`. No template ever evaluated. (In *identifier* position the `$` instead yields uncompilable output — a build failure, not injection.)
- **Annotation values never reach emitted code.** `<annotation value="true&quot;) ; PWNED ; //">` — the string `PWNED` appears in **zero** emitted `.kt` files; values are only compared against known constants (`"true"`, flag names).
[...]

PR #204, opened 2026-08-07
fix(jvm): bound hostile wire input to the D-Bus spec limits and stop a dead reader reporting healthy

### LINKED ISSUE #190: JVM wire backend: hostile peer input kills the connection (unbounded lengths, uncapped variant recursion, silent array truncation)
Found during a deliberate security sweep of the JVM wire backend (audit pass, stream 3). Every item below was **reproduced**, not inferred — commands and outputs are included. No fix is included; this is a measurement.

## Summary

`WireMessageCodec` / `DBusReader` trust wire-supplied length prefixes and nesting depth without bound, and the reader thread catches only `IOException` and `DBusMarshallingException`. The result: **a single hostile frame permanently kills a `DBusWireConnection`**, and the connection does not report itself as dead — every subsequent call just times out, forever.

## The load-bearing fact

`DBusWireConnection.kt:202-212` — the reader loop catches exactly two exception types:

```kotlin
} catch (_: IOException) {
} catch (_: DBusMarshallingException) {
} finally {
    failAllPending(IOException("D-Bus connection closed"))
}
```

Anything else escapes, the thread exits, the socket stays open, `running` stays `true`, and `close()` is never called.

## Reproduced: a hostile peer kills the connection

A throwaway probe stood up an `AFUNIXServerSocket`, completed SASL EXTERNAL + `AGREE_UNIX_FD` + `BEGIN`, then sent **one** frame to a client connected via `DBusWireConnection.connectDirect(...)`:

```
PROBE[headerTypeConfusion] readerRunning BEFORE = true
PROBE[headerTypeConfusion] readerRunning AFTER  = false
PROBE[headerTypeConfusion] subsequent call = DBusCallException: org.freedesktop.DBus.Error.Timeout: Method call timed out after 1200ms

PROBE[hugeBodyLength]      readerRunning BEFORE = true
PROBE[hugeBodyLength]      readerRunning AFTER  = false
PROBE[hugeBodyLength]      subsequent call = DBusCallException: org.freedesktop.DBus.Error.Timeout: Method call timed out after 1200ms

PROBE[nestedVariants]      readerRunning BEFORE = true
PROBE[nestedVariants]      readerRunning AFTER  = false
PROBE[nestedVariants]      subsequent call = DBusCallException: org.freedesktop.DBus.Error.Timeout: Method call timed out after 1200ms
```

## The individual parser defects

Measured directly against `DBusMarshaller` / `WireMessageCodec`:

```
PROBE[frame bodyLength=0x7F000000 (2.1GB) fields=0]  THREW java.lang.OutOfMemoryError: Java heap space
PROBE[frame bodyLength=0x80000000 fields=0]          THREW java.lang.NegativeArraySizeException: -2147483648
PROBE[frame fieldsLength=0x80000000 body=0]          THREW java.lang.NegativeArraySizeException: -2147483648
PROBE[nestedVariant depth=10000 bodyBytes=30004]     THREW java.lang.StackOverflowError
PROBE[string len=0x80000000]                         THREW java.lang.StringIndexOutOfBoundsException: Range [4, 4 + -2147483648) out of bounds for length 6
PROBE[string len=0xFFFFFFFF]                         THREW java.lang.StringIndexOutOfBoundsException: Range [4, 4 + -1) out of bounds for length 6
PROBE[array ay len=0xFFFFFFF0 (4 real bytes follow)] OK -> [[]]
PROBE[array ay len=0x80000000]                       OK -> [[]]
PROBE[header PATH variant type 'u' instead of 'o']   THREW java.lang.ClassCastException: class kotlin.UInt cannot be cast to class java.lang.String
PROBE[header MEMBER variant type 'ay' instead of 's'] THREW java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class java.lang.String
```

Which decompose into five distinct defects:

1. **No maximum message size.** `WireMessage.kt:186-191` reads `bodyLength` and `fieldsLength` and passes them straight to `readFully`, which does `ByteArray(count)`. **16 bytes of input produce a 2.1 GB allocation attempt.** The D-Bus spec caps a message at 128 MiB and the header array at 64 MiB; neither is enforced. Note that even a fully *spec-legal* 128 MiB `ay` message amplifies badly: it demarshals to ~134M boxed `UByte` in an `ArrayList`, several GB of heap.

2. **`UInt` -> `Int` overflow defeats the bounds checks.** `WireMessage.kt:186-187` and `DBusMarshaller.kt:454,468` do `.toInt()` on wire-supplied 32-bit lengths. Values >= 2^31 go negative, and the guards are written as `offset + length + 1 > bytes.size` (`DBusMarshaller.kt:458`) and `end > bytes.size` (`DBusMarshaller.kt:470`) — both of which a negative length *passes*.

3. **Silent truncation on a negative array length** — the worst of the set, because it does not throw at all. An array whose declared byte-length is `0xFFFFFFF0` demarshals to an **empty list**, and the caller cannot tell it from a genuinely empty array. This is silent data loss driven by peer input.

4. **Unbounded variant-nesting recursion.** `DBusReader.unmarshalVariant` (`DBusMarshaller.kt:509`) recurses with no depth cap. Each nesting level costs **3 bytes on the wire** (`0x01 'v' 0x00`), so a ~30 KB body reaches 10,000 frames and overflows the stack. The D-Bus spec caps nesting at 32. (The *signature* parser `DBusSignatureParser.parseOne` is likewise uncapped — `"a"*10000` also `StackOverflowError`s — but is bounded to 255 in practice by the `g` type's one-byte length prefix, so it is only reachable through this variant path.)

5. **Unchecked casts on header fields.** `WireMessage.kt:221-229` does `variant.value as String` / `as UInt` on peer-supplied header-field variants, so a wrong variant type yields `ClassCastException` rather than a `DBusMarshallingException`.

## Severity, with its precondition — please read this part

**This is not "remote code execution on the session bus."** `dbus-daemon` validates relayed messages hard: it enforces the 128 MiB cap, the 32-level nesting cap, header-field types, and body-vs-signature consistency. So on a **brokered** session or system bus, most of the malformed inputs above cannot reach a client — the daemon rejects them first. The honest reachable set is:

- **Brokerless `connectDirect()` (peer-to-peer) — fully reachable, no daemon, no filter.** All ten probes above land. This is the case the reproduction actually used.
- **Brokered bus — partially reachable.** The memory-amplification path (defect 1) survives daemon validation, because a 128 MiB message is *spec-legal* and the daemon will relay it; our side then amplifies it into multi-GB of boxed objects. A peer that can talk to your bus can OOM your process without sending anything malformed.
- **Defect 3 (silent truncation)** is a correctness bug independent of any attacker, and is the one I would rank first for a library that is about to be depended on, because it fails silently.

Today the dominant consumer is BlueZ on a session bus, where the peer is trusted and the daemon filters — so the practical exposure right now is low. It rises the moment anyone uses direct connections or points this at a bus with untrusted peers, and the library's own docs present direct mode as a supported mode.

## Suggested shape of a fix (not implemented)

- Enforce `DBUS_MAXIMUM_MESSAGE_LENGTH` (128 MiB) and `DBUS_MAXIMUM_ARRAY_LENGTH` (64 MiB) in `WireMessageCodec.read` before allocating.
- Compare lengths as `Long`/`UInt` and reject anything negative-after-conversion, rather than `.toInt()` then range-check.
- Add a nesting-depth counter to `DBusReader` capped at 32 (`DBUS_MAXIMUM_TYPE_RECURSION_DEPTH`).
- Make the array-length path throw instead of returning empty.
- Replace the header-field `as` casts with checked `as?` + a `DBusMarshallingException`.
- Widen the reader-thread catch to `Throwable` so a parser bug degrades to "this connection died" rather than "this connection silently stops delivering," and surface that state (`isReaderRunning` exists but nothing consults it).

Probe files were deleted; the worktree was left clean.

monkopedia-coder:
Fixes #190

The JVM wire backend trusted every length prefix and nesting depth a peer put on the wire, and the reader thread caught only `IOException` and `DBusMarshallingException`. Anything else escaped: the thread exited, the socket stayed open, `running` stayed `true` — so the connection reported itself healthy while delivering nothing, and every later call timed out.

## Red first

The audit's throwaway probes are now a real test — `src/jvmTest/kotlin/com/monkopedia/sdbus/internal/jvmdbus/HostileWireInputTest.kt` — including a fake AF_UNIX peer that completes SASL EXTERNAL + `AGREE_UNIX_FD` + `BEGIN`, sends exactly **one** hostile frame, and then **holds the socket open** (that is the point: it distinguishes "the peer hung up", which the reader always reported, from "the peer is still there and we can no longer understand it").

Commit `1abd276` contains the tests only. Verbatim output of `./gradlew :jvmTest --tests "com.monkopedia.sdbus.internal.jvmdbus.HostileWireInputTest"` at that commit — **8 of 9 failed**:

```
FAIL  frameDeclaringA2GbBody_isRejectedWithoutAttemptingTheAllocation[jvm]
        java.lang.AssertionError: Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was java.lang.OutOfMemoryError: Java heap space
FAIL  frameWithBodyLengthThatNarrowsNegative_isRejected[jvm]
        java.lang.AssertionError: Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was java.lang.NegativeArraySizeException: -2147483648
FAIL  frameWithHeaderFieldsLengthThatNarrowsNegative_isRejected[jvm]
        java.lang.AssertionError: Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was java.lang.NegativeArraySizeException: -2147483648
FAIL  nestedVariantsBeyondSpecDepth_areRejectedInsteadOfOverflowingTheStack[jvm]
        java.lang.AssertionError: Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was java.lang.StackOverflowError
FAIL  stringLengthThatNarrowsNegative_isRejectedInsteadOfIndexingOutOfBounds[jvm]
        java.lang.AssertionError: string length 2147483648. Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was java.lang.StringIndexOutOfBoundsException: Range [4, 4 + -2147483648) out of bounds for length 6
FAIL  arrayLengthThatNarrowsNegative_isRejectedRatherThanSilentlyDecodingToEmpty[jvm]
        java.lang.AssertionError: array length 4294967280. Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was completed successfully with the result: <DBusUnmarshalResult(values=[[]], offset=4)>.
FAIL  headerFieldCarryingTheWrongVariantType_isRejectedAsAMarshallingFailure[jvm]
        java.lang.AssertionError: PATH as 'u'. Expected an exception of class com.monkopedia.sdbus.internal.jvmdbus.DBusMarshallingException to be thrown, but was java.lang.ClassCastException: class kotlin.UInt cannot be cast to class java.lang.String (kotlin.UInt is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
FAIL  hostileFrameFromAPeer_failsTheConnectionVisiblyInsteadOfSilently[jvm]
        java.lang.AssertionError: Expected an exception of class java.io.IOException to be thrown, but was com.monkopedia.sdbus.internal.jvmdbus.DBusCallException: org.freedesktop.DBus.Error.Timeout: Method call timed out after 5000ms
PASS  nestedVariantsWithinSpecDepth_stillDecode[jvm]
```

That last `PASS` is the positive control: nesting the spec *permits* already decoded correctly before the fix, so the suite discriminates the defect rather than the input shape. All eight are green after `6e02cf5`.

Reproduce: `git checkout 1abd276 && ./gradlew :jvmTest --tests "com.monkopedia.sdbus.internal.jvmdbus.HostileWireInputTest"`. Use `:jvmTest`, not `jvmTest` — the unqualified name also matches `:cross_test:jvmTest`, which then fails the *filter* rather than the test.

## Which spec limits, and why those

The point of using the specification's own numbers is that this becomes a **conformance check rather than an invented threshold** — `dbus-daemon` enforces the same values on every message it relays, so nothing a conforming peer may legally send is refused.

| Constant | Value | Applied to |
|---|---|---|
| `DBUS_MAXIMUM_MESSAGE_LENGTH` | 2^27 = 128 MiB | declared body length; total frame (padded header + body) |
| `DBUS_MAXIMUM_ARRAY_LENGTH` | 2^26 = 64 MiB | any array's content byte-length, incl. the `a(yv)` header-fields array |
| container nesting depth | 64 | every container level in `DBusReader` and in `DBusSignatureParser` |

**On the nesting number — I used 64, not 32, deliberately.** The spec caps a signature at "32 array type codes **and** 32 open parentheses" — those are two separate budgets — and states the consequence explicitly: *"the maximum total depth of recursion is 64"*. A flat cap of 32 on a single combined counter would therefore reject spec-legal signatures (e.g. 20 arrays nested with 20 structs). Counting every container level against 64 rejects nothing the spec permits, while still turning the 10,000-deep variant bomb — 30 KB of body — into a clean `DBusMarshallingException`. Choosing 32 here would have been the *less* conformant option, which is the opposite of the goal.

## Fixing the comparison structurally, not adding another guard

This is the same defect class as #178/#180 (`absoluteTimeoutOf`), so it gets the same treatment: **make the overflow impossible rather than checking for it.**

The bug was ordering — `.toInt()` first, range-check second. A u32 at or above 2^31 narrows to a negative `Int`, and both guards were written as `x > bytes.size`, which a negative left-hand side passes silently.

Wire lengths are now validated **while still unsigned**, and there are exactly two places where a wire length becomes an `Int`:

- `DBusReader.readLength(limit, what)` — `readRaw(4)` zero-extends four bytes into a `Long`, so the comparison against the limit is exact and unsigned; the narrowing that follows is provably in `0..limit`.
- `WireMessageCodec.checkedLength(length: UInt, limit, what)` — same, in `UInt` space, for the two framing lengths.

Both can only *return* a value in `0..limit`. There is no remaining narrowing anywhere in the read path that could produce a negative length, so no downstream `> bytes.size` check can be handed one. That is the structural property, not a new guard sitting beside the old one.

## The silent one

`array len 0xFFFFFFF0 -> OK -> [[]]` is the case I fixed with the most conviction, and it is worth calling out separately because **it is not an attack**. The declared length narrowed to `-16`, which made `end < offset`, so the fill loop never ran and the caller received a **successful decode of an empty array** — indistinguishable from a genuinely empty one. Peer-driven data loss with no error anywhere. It is now a `DBusMarshallingException` like every other malformed length.

## Making the reader's death observable

I did **both** of the options in the issue, because they answer different halves of the complaint.

The reader now catches `Throwable`. A frame that fails to parse is unrecoverable either way — the stream is *framed*, so once a parse fails the byte offset is no longer known and nothing after it can be decoded — so the connection still dies. What changed is that it stops dying *quietly*:

- the cause is recorded in `readerFailure` before the thread exits;
- the socket is closed, so we stop pretending to be attached to a peer we can no longer read;
- in-flight callers get the **real** cause via `failAllPending`, not a generic `IOException("D-Bus connection closed")`;
- `checkAlive()` at the two entry points that put bytes on the wire (`dispatchCall`, `send`) refuses further traffic with that cause.

The net effect is the one the issue asked for: a call on a dead connection now fails **immediately** with `IOException: D-Bus connection is dead: …` instead of sitting out its full timeout against a black hole. `isReaderRunning` is still there and now agrees with reality (the reader clears `running` itself); `readerFailure` is the thing production code consults, and it is `private` — the reader-health signal callers actually need is "your send failed, here is why", which they now get.

## What is now rejected that previously "worked"

Stating this plainly, per the issue: a message above **128 MiB**, an array above **64 MiB**, or container nesting past depth **64** is now refused with a `DBusMarshallingException`. Previously those "worked" in the sense that they produced an `OutOfMemoryError`, a `NegativeArraySizeException` or a `StackOverflowError` — none of which is a usable outcome, and all of which are beyond what a conforming peer may send. Realistically nobody was relying on this, but it *is* a behaviour change and it is now a hard error rather than a crash.

Also newly rejected: a header field whose variant carries the wrong type (`PATH` as `u`, `MEMBER` as `ay`, …). Previously an unchecked `as String` / `as UInt` raised `ClassCastException`, which nothing on the reader path caught — that is precisely how the umbrella failure was reached.

## Known gaps (deliberately not fixed here)

- **Memory amplification by a *spec-legal* message.** A 128 MiB `ay` is legal, will be relayed by `dbus-daemon`, and demarshals to ~134M boxed `UByte` in an `ArrayList` — several GB of heap. Fixing that means a `ByteArray`-backed representation for `ay`, which is a value-model change well outside a hardening pass. Called out in #190 and still true.
- **Body-vs-declared-length consistency** is not validated; a body shorter than its signature demands still fails as an out-of-bounds read rather than a length mismatch. Same clean exception type, less specific message.

## Scope and verification

Scope is the JVM wire backend only. `:codegen` is untouched (#194 covers the codegen-side DoS) and the native backend is untouched.

- `./gradlew compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64 ktlintCheck apiCheck` — **BUILD SUCCESSFUL**.
- **`apiCheck` did not move**; there is no `api/*.api` diff. Everything changed is `internal` (`DBusMarshaller.kt`, `WireMessage.kt`, `DBusWireConnection.kt`), and `readerFailure` / `checkAlive` are `private`. The klib ABI consumed by `blue-falcon-sdbus` is unchanged.
- `DBUSMOCK_PYTHON=<venv>/bin/python3 dbus-run-session -- ./gradlew jvmTest linuxX64Test` — **BUILD SUCCESSFUL**, 759 tests, 0 failures, 0 skips (root jvmTest 336, root linuxX64Test 297, cross_test jvmTest 64, cross_test linuxX64Test 62). `DBUSMOCK_PYTHON` was set to a real `python-dbusmock` venv so the cross_test dbusmock suites genuinely ran rather than no-op passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
 
 ## [Unreleased]
 
+### Fixed — JVM backend
+
+- **The JVM wire backend now bounds hostile peer input, and a dead connection says so.** Length
+  prefixes and container nesting arriving from a peer were trusted without limit, and a u32 length
+  at or above 2^31 narrowed to a *negative* `Int` that passed every `length > size` check. A single
+  frame could therefore raise `OutOfMemoryError` (a declared 2.1 GB body from 16 bytes of input),
+  `NegativeArraySizeException`, `StackOverflowError` (nested variants cost 3 bytes per level) or
+  `ClassCastException` (a header field carrying the wrong variant type) — and an array declaring
+  `0xFFFFFFF0` bytes decoded *successfully* to an empty list, silently losing the real contents.
+  Wire lengths are now validated while still unsigned against the D-Bus specification's own limits
+  (128 MiB per message, 64 MiB per array, container nesting depth 64), and header fields use checked
+  casts. Anything past those limits is now rejected as a malformed message where it previously
+  raised one of the errors above. (#190)
+- **A reader thread that dies no longer leaves the connection reporting itself healthy.** The reader
+  caught only `IOException` and `DBusMarshallingException`; anything else killed the thread while the
+  socket stayed open, so the connection accepted calls that could only ever time out. It now catches
+  broadly, records the cause, closes the socket, fails in-flight calls with the real error rather
+  than a generic "connection closed", and refuses further sends — so a call on a dead connection
+  fails immediately instead of waiting out its timeout. (#190)
+
 ### Added — codegen
 
 - **Generated code now carries KDoc taken from the introspection XML.** Documentation reaches the
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusMarshaller.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusMarshaller.kt
@@ -60,6 +60,28 @@ internal enum class Endian(val code: Byte) {
 internal class DBusMarshallingException(message: String) : RuntimeException(message)
 
 /**
+ * The D-Bus specification's hard protocol limits, which `dbus-daemon` enforces on every message it
+ * relays. Using the spec's own numbers keeps the checks below a CONFORMANCE test rather than an
+ * invented threshold: nothing a conforming peer may legally send is refused.
+ */
+internal object DBusLimits {
+    /** `DBUS_MAXIMUM_MESSAGE_LENGTH`: 2^27 bytes for header + alignment padding + body. */
+    const val MAX_MESSAGE_LENGTH = 134_217_728
+
+    /** `DBUS_MAXIMUM_ARRAY_LENGTH`: 2^26 bytes of array content (the header fields are an array). */
+    const val MAX_ARRAY_LENGTH = 67_108_864
+
+    /**
+     * Maximum container-nesting depth. The spec caps a signature at "32 array type codes and 32
+     * open parentheses", and says so with the consequence spelled out: "the maximum total depth of
+     * recursion is 64". Counting EVERY container level (array, struct, dict-entry, variant) against
+     * that 64 therefore accepts every signature the spec permits, while still refusing a
+     * nested-variant bomb — each level of which costs only 3 bytes on the wire.
+     */
+    const val MAX_TYPE_RECURSION_DEPTH = 64
+}
+
+/**
  * The parsed model of a single complete D-Bus type, carrying its natural [alignment] and the
  * signature fragment it [code]s back to.
  */
@@ -121,23 +143,31 @@ internal object DBusSignatureParser {
     }
 
     /** Parses exactly one complete type starting at [index]; returns it and the next index. */
-    fun parseOne(signature: String, index: Int): Pair<DBusType, Int> {
+    fun parseOne(signature: String, index: Int): Pair<DBusType, Int> = parseOne(signature, index, 0)
+
+    private fun parseOne(signature: String, index: Int, depth: Int): Pair<DBusType, Int> {
         if (index >= signature.length) {
             throw DBusMarshallingException("Unexpected end of signature '$signature'")
+        }
+        if (depth > DBusLimits.MAX_TYPE_RECURSION_DEPTH) {
+            throw DBusMarshallingException(
+                "Signature nests deeper than the D-Bus maximum of " +
+                    "${DBusLimits.MAX_TYPE_RECURSION_DEPTH}: '$signature'"
+            )
         }
         return when (val c = signature[index]) {
             'a' -> {
                 if (signature.getOrNull(index + 1) == '{') {
-                    val (entry, next) = parseDictEntry(signature, index + 1)
+                    val (entry, next) = parseDictEntry(signature, index + 1, depth + 1)
                     DBusType.ArrayType(entry) to next
                 } else {
-                    val (element, next) = parseOne(signature, index + 1)
+                    val (element, next) = parseOne(signature, index + 1, depth + 1)
                     DBusType.ArrayType(element) to next
                 }
             }
 
-            '(' -> parseStruct(signature, index)
-            '{' -> parseDictEntry(signature, index)
+            '(' -> parseStruct(signature, index, depth + 1)
+            '{' -> parseDictEntry(signature, index, depth + 1)
             'v' -> DBusType.VariantType to index + 1
             'y', 'b', 'n', 'q', 'i', 'u', 'x', 't', 'd', 's', 'o', 'g', 'h' ->
                 DBusType.Basic(c) to index + 1
@@ -148,14 +178,14 @@ internal object DBusSignatureParser {
         }
     }
 
-    private fun parseStruct(signature: String, openIndex: Int): Pair<DBusType, Int> {
+    private fun parseStruct(signature: String, openIndex: Int, depth: Int): Pair<DBusType, Int> {
         val fields = mutableListOf<DBusType>()
         var index = openIndex + 1
         while (signature.getOrNull(index) != ')') {
             if (index >= signature.length) {
                 throw DBusMarshallingException("Unterminated struct in '$signature'")
             }
-            val (field, next) = parseOne(signature, index)
+            val (field, next) = parseOne(signature, index, depth)
             fields.add(field)
             index = next
         }
@@ -165,9 +195,9 @@ internal object DBusSignatureParser {
         return DBusType.StructType(fields) to index + 1
     }
 
-    private fun parseDictEntry(signature: String, openIndex: Int): Pair<DBusType, Int> {
-        val (key, afterKey) = parseOne(signature, openIndex + 1)
-        val (value, afterValue) = parseOne(signature, afterKey)
+    private fun parseDictEntry(signature: String, openIndex: Int, depth: Int): Pair<DBusType, Int> {
+        val (key, afterKey) = parseOne(signature, openIndex + 1, depth)
+        val (value, afterValue) = parseOne(signature, afterKey, depth)
         if (signature.getOrNull(afterValue) != '}') {
             throw DBusMarshallingException("Unterminated dict-entry in '$signature'")
         }
@@ -361,6 +391,9 @@ internal class DBusReader(private val bytes: ByteArray, offset: Int, private val
     var offset: Int = offset
         private set
 
+    // How many container levels deep [unmarshalValue] currently is; see [nested].
+    private var depth = 0
+
     fun align(boundary: Int) {
         val rem = offset % boundary
         if (rem != 0) offset += boundary - rem
@@ -394,15 +427,54 @@ internal class DBusReader(private val bytes: ByteArray, offset: Int, private val
         return bytes[offset++].toInt() and 0xff
     }
 
+    /**
+     * Reads a 32-bit wire length prefix, rejecting anything above [limit] BEFORE it is narrowed to
+     * [Int]. [readRaw] zero-extends the four bytes into a [Long], so the comparison is exact and
+     * unsigned and the narrowing that follows is provably in `0..limit`.
+     *
+     * This is the ONLY place a wire-supplied length becomes an [Int], which is what makes the
+     * overflow structurally impossible rather than merely guarded: no `length > bytes.size` check
+     * downstream can be handed a negative length that silently passes it (issue #190).
+     */
+    private fun readLength(limit: Int, what: String): Int {
+        val length = readRaw(4)
+        if (length > limit) {
+            throw DBusMarshallingException(
+                "Declared $what length $length exceeds the D-Bus maximum of $limit bytes"
+            )
+        }
+        return length.toInt()
+    }
+
     /** Reads [types], one value per type, in order. */
     fun unmarshal(types: List<DBusType>): List<Any?> = types.map { unmarshalValue(it) }
 
     fun unmarshalValue(type: DBusType): Any? = when (type) {
         is DBusType.Basic -> unmarshalBasic(type.type)
-        is DBusType.ArrayType -> unmarshalArray(type)
-        is DBusType.StructType -> unmarshalStruct(type)
-        is DBusType.DictEntryType -> unmarshalDictEntry(type)
-        DBusType.VariantType -> unmarshalVariant()
+        is DBusType.ArrayType -> nested { unmarshalArray(type) }
+        is DBusType.StructType -> nested { unmarshalStruct(type) }
+        is DBusType.DictEntryType -> nested { unmarshalDictEntry(type) }
+        DBusType.VariantType -> nested { unmarshalVariant() }
+    }
+
+    /**
+     * Runs [block] one container level deeper, refusing to descend past
+     * [DBusLimits.MAX_TYPE_RECURSION_DEPTH]. Demarshalling is recursive, so without this a peer can
+     * spend 3 bytes per level to buy a stack frame and overflow the stack from a tiny body.
+     */
+    private fun <T> nested(block: () -> T): T {
+        if (depth >= DBusLimits.MAX_TYPE_RECURSION_DEPTH) {
+            throw DBusMarshallingException(
+                "Container nesting exceeds the D-Bus maximum depth of " +
+                    "${DBusLimits.MAX_TYPE_RECURSION_DEPTH}"
+            )
+        }
+        depth++
+        try {
+            return block()
+        } finally {
+            depth--
+        }
     }
 
     private fun unmarshalBasic(type: Char): Any = when (type) {
@@ -451,7 +523,7 @@ internal class DBusReader(private val bytes: ByteArray, offset: Int, private val
     private fun readString(lengthPrefixBytes: Int): String {
         val length = if (lengthPrefixBytes == 4) {
             align(4)
-            readRaw(4).toInt()
+            readLength(DBusLimits.MAX_MESSAGE_LENGTH, "string")
         } else {
             readByte()
         }
@@ -465,7 +537,7 @@ internal class DBusReader(private val bytes: ByteArray, offset: Int, private val
 
     private fun unmarshalArray(type: DBusType.ArrayType): Any {
         align(4)
-        val byteLength = readRaw(4).toInt()
+        val byteLength = readLength(DBusLimits.MAX_ARRAY_LENGTH, "array")
         align(type.element.alignment)
         val end = offset + byteLength
         if (end > bytes.size) {
--- a/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
+++ b/src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/DBusWireConnection.kt
@@ -140,6 +140,13 @@ internal class DBusWireConnection private constructor(
     private var readerThread: Thread? = null
 
     /**
+     * The error that tore the reader thread down, or null while the connection is healthy (and
+     * after an ordinary [close]). Set exactly once, by the reader itself, before it exits.
+     */
+    @Volatile
+    private var readerFailure: Throwable? = null
+
+    /**
      * The unique name assigned by the bus during [hello] (e.g. `:1.42`), or null before Hello — and
      * always null on a brokerless [direct] connection, which has no daemon to assign one.
      */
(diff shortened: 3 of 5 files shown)


monkopedia-reviewer (commented):
Reviewed: b4ccf7e517614d926c946f2550788894a92867e9

## Verdict: verification holds — recommend **approve**. Escalating per policy rather than merging (see Tier below).

## Red-first, reproduced myself (verbatim)

Checked out `1abd276` in an isolated worktree and ran `dbus-run-session -- ./gradlew :jvmTest --tests "com.monkopedia.sdbus.internal.jvmdbus.HostileWireInputTest"` directly (JDK 21). **8 of 9 failed**, matching the PR body exactly:

```
frameDeclaringA2GbBody...                       -> java.lang.OutOfMemoryError
frameWithBodyLengthThatNarrowsNegative           -> java.lang.NegativeArraySizeException
frameWithHeaderFieldsLengthThatNarrowsNegative   -> java.lang.NegativeArraySizeException
nestedVariantsBeyondSpecDepth...                 -> java.lang.StackOverflowError
stringLengthThatNarrowsNegative...               -> java.lang.StringIndexOutOfBoundsException
arrayLengthThatNarrowsNegative...                -> AssertionError: "...completed successfully with the result: <DBusUnmarshalResult(values=[[]], offset=4)>"
headerFieldCarryingTheWrongVariantType...        -> java.lang.ClassCastException (UInt -> String)
hostileFrameFromAPeer...                         -> DBusCallException: Error.Timeout after 5000ms
```

I pulled the raw JUnit XML for the silent-truncation case specifically (the one worth doubling-checking, since it's an `AssertionError` rather than a wrapped exception): confirmed the message is exactly `array length 4294967280 ... completed successfully with the result: <DBusUnmarshalResult(values=[[]], offset=4)>` — the empty-array data-loss bug is real, not a description of it.

**Positive control** (`nestedVariantsWithinSpecDepth_stillDecode`): confirmed PASSED at `1abd276` — no `<failure>` element in the XML. So the suite discriminates the defect rather than the input shape, as claimed.

At `b4ccf7e` (HEAD): re-ran the same targeted suite — all 9 pass.

## 32 vs 64 nesting — checked the spec text directly, author is right

Fetched `dbus-specification.html` and pulled the literal sentences: *"The maximum depth of container type nesting is 32 array type codes and 32 open parentheses."* ... *"This implies that the maximum total depth of recursion is 64, for an 'array of array of array of ... struct of struct of struct of ...' where there are 32 array and 32 struct."*

That's exactly the author's reading: two separate 32-budgets, spec-stated combined max of 64. I traced the depth-counting in both `DBusSignatureParser.parseOne` (signature-parse time) and `DBusReader.nested()` (unmarshal time): both permit exactly 64 levels of nesting inclusive (verified the off-by-one — `parseOne`'s check is `depth > 64` at entry, `nested()`'s is `depth >= 64` before increment; tracing through both by hand, both let a chain of 64 container opens through and reject the 65th). So a legal 32-array-of-32-struct signature (64 levels) is accepted by this implementation and would have been wrongly rejected by a flat 32. Also confirmed the two size constants against the spec text directly: `DBUS_MAXIMUM_ARRAY_LENGTH` = 2^26 = 67,108,864 (64 MiB) and the message-length max = 2^27 = 134,217,728 (128 MiB) — both match `DBusLimits` exactly.

## Unsigned-validation funnel — exhaustive, verified by grep + read, not just claimed

Grepped every `.toInt()`/`.toUInt()` in `src/jvmMain/.../internal/jvmdbus/*.kt` (the whole wire read path) and read each site. Every wire-derived value that feeds an **allocation size or a recursion bound** goes through `DBusReader.readLength` or `WireMessageCodec.checkedLength` — both compare in `Long`/`UInt` space before narrowing, both provably return `0..limit`. The other `.toInt()/.toUInt()` sites are either: fixed-width scalars (2-byte `n`/`q`, single bytes) that can't overflow into a usable attack; `i`/`u`/`h` basic-type values that are stored as plain scalars, never used to size anything; already-unmarshalled reply values consumed by `DBusWireConnection` (post-funnel, safe by construction); or the `h` (unix-fd index) case in `resolveInboundFds`, which uses `fds.getOrNull(index)` — defensively bounds-checked regardless of sign. I also checked `unmarshalArray`'s decode loop: it appends into a growable list/map rather than pre-allocating to the declared length, so even the legal 64 MiB array case doesn't get a pre-sized allocation from an attacker-chosen number. And I confirmed the total-message check closes the gap the two per-field checks leave open: `bodyLength` and `fieldsLength` are each individually bounded, then `paddedHeaderEnd + bodyLength` is checked again against the 128 MiB total — so two individually-legal parts can't sum past the spec's whole-message cap.

Minor, pre-existing, **not introduced by this PR**: `dispatchCall` calls `checkAlive()` then later inserts into `pending` before `writeMessage`; if the reader dies in the narrow window between the `checkAlive()` check and the `pending[serial] = future` insert, `failAllPending()` may already have drained the map, and that one call would wait out its own timeout rather than failing immediately. This is an edge case of the general "connection dies concurrently with a send" race, not a regression — flagging for awareness, not blocking.

## Reader-thread death is now observable, and a dead connection fails fast — verified by reading the diff

`catch (t: Throwable)` replaces the old `IOException`/`DBusMarshallingException`-only catch. On any escape it records `readerFailure` (private, `@Volatile`) exactly once in `finally`, sets `running = false`, closes the socket, and calls `failAllPending` with the real cause. `checkAlive()` — called at the top of both `dispatchCall` and `send` — throws `IOException("D-Bus connection is dead: ...", cause)` immediately once `readerFailure` is set. So a call issued after the reader has already died fails immediately with the real cause instead of waiting out a 5s timeout, which is exactly what #190 asked for.

## apiCheck — confirmed, not just claimed

Ran `./gradlew apiCheck ktlintCheck` myself: `BUILD SUCCESSFUL`. `git diff $(merge-base)..HEAD -- 'api/*.api'` is empty. Read-confirmed `readerFailure` and `checkAlive` are `private`; all three touched files (`DBusMarshaller.kt`, `WireMessage.kt`, `DBusWireConnection.kt`) are `internal`. The klib ABI `blue-falcon-sdbus` consumes is unmoved.

## Full suite, run myself, DBUSMOCK_PYTHON set to a real venv

- `dbus-run-session -- ./gradlew :jvmTest :linuxX64Test` — **BUILD SUCCESSFUL**. Parsed the JUnit XML myself: jvmTest 336/336, linuxX64Test 297/297, 0 failures, 0 errors, 0 skips.
- `DBUSMOCK_PYTHON=<venv>/python3 dbus-run-session -- ./gradlew :cross_test:jvmTest :cross_test:linuxX64Test` — **BUILD SUCCESSFUL**. 64/64 and 62/62, 0 failures, 0 skips. Spot-checked the dbusmock suites' XML for wall-clock time to rule out the known local no-op trap (a fake pass with `DBUSMOCK_PYTHON` unset runs in ~0.003s) — `DbusmockBluezTest` took 3.6s, `DbusmockForeignErrorTest` 2.5s, `DbusmockAdaptorPropertiesChangedTest` 4.0s for a handful of tests each: real dbusmock peers spawned and were exercised, not a silent skip.
- Total: 336+297+64+62 = **759**, 0 failures, 0 skips — matches the PR body's reported count exactly.
- Grepped `cross_test`/`stress_test`/root tests for any large-payload fixture that could trip the new 64 MiB array / 128 MiB message caps — found none (largest is a small "large map" with a bounded element count, nowhere near MB scale). The full-suite green run above is the empirical confirmation that nothing legitimate got rejected.

## Declared gaps

The PR is upfront about two known-unfixed items: memory amplification from a *spec-legal* 128 MiB `ay` (boxed `UByte` list, several GB), and body-vs-declared-length consistency (a short body currently fails as an out-of-bounds read rather than a specific length-mismatch message). Both are correctly scoped out of a hardening pass aimed at the audit's specific findings — neither is a hostile-input DoS/crash on their own (the first needs a legal 128 MiB message to even reach, the second still fails safely, just with a less specific exception). Reasonable to leave for a follow-up; the PR body states them honestly rather than hiding them.

## What I took on trust

- The native (linuxX64/linuxArm64) backend is genuinely out of scope — confirmed by diff (only JVM files touched) and the PR body says so; I did not re-audit sd-bus's own C parsing.
- I did not independently re-derive `dbus-daemon`'s actual enforcement of these limits at the C level (i.e., I trust the spec text over probing a live `dbus-daemon`); the spec text itself I did fetch and quote directly rather than trust the PR's paraphrase.
- CI's own green run (`gh pr checks`) — I re-ran the substantive test suites myself rather than relying solely on it, but did not re-run the Cross Runtime Stress Matrix `pr-smoke` jobs myself.

---

## Tier — this needs to be flagged, since it changes the disposition

Classifying this PR: it fixes a security-audit-flagged remote-DoS / memory-safety vulnerability in wire-protocol parsing, reachable by any D-Bus peer with no privilege — this is squarely `security` per `review-policy.md`'s category table, not `bug_fix`. `Monkopedia/sdbus-kotlin` carries **no project override** ("Defaults work — no overrides"), so the default table applies: `security` → **tier 3**, not tier 1/2. That means the gate is the owner reviewing it directly (GitHub or in-session, disclosed) — not this subagent's approval, however thorough the verification.

I have **not** merged. I'm adding `Monkopedia` as a requested reviewer per the tier-3 mechanism and posting this as a `COMMENT` review (not an approval), per the escalation precondition in policy: only an otherwise-mergeable PR gets escalated, carrying the residual judgment call. Every check above is exactly the kind of thing tier 1 would auto-merge on if this classified as `bug_fix` — the only reason it doesn't is the `security` category itself, which the policy routes to you regardless of how clean the fix is underneath.

**Verdict if approved:** approve and merge (squash, delete branch) — every claim in the PR body was independently reproduced or verified against source/spec, CI is green at this exact head (`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`), and `apiCheck` confirmed zero ABI movement.

**One-sentence question:** the fix, tests, and spec-conformance reasoning all check out against independent verification — approve and merge #204?

Issue #197, opened 2026-08-07
Method.NoReply is mapped onto zero on native, so hasNoReply never reaches the sd-bus vtable


monkopedia-coder:
`Flags.native.kt:61-63` maps `METHOD_NO_REPLY` onto **zero**, so `hasNoReply = true` never reaches sd-bus's vtable and the annotation is never advertised in served introspection — on either backend.

```kotlin
if (has(METHOD_NO_REPLY)) {
    sdbusFlags = sdbusFlags or 0u // SD_BUS_VTABLE_METHOD_NO_REPLY
}
```

The comment names the constant that was meant to go there. It was never bound, so the branch is a no-op: it tests the flag and then ORs nothing.

## How this surfaced, and why it is filed separately

Found while correcting the documentation split off from #193 (PR #195). That work ran an actual introspection round-trip against a live adaptor on both backends — `org.freedesktop.DBus.Introspectable.Introspect` under `dbus-run-session` — rather than reasoning from the vtable contract, and the measurement **contradicted #193's own description**:

| Annotation | Native | JVM |
| --- | --- | --- |
| `Deprecated` on method/signal/property | served | served |
| `Deprecated` on the **interface** | served | not served |
| `EmitsChangedSignal` (`const`/`invalidates`/`false`) | served | not served |
| **`Method.NoReply`** | **not served** | **not served** |

#193 had listed `Method.NoReply` as a native-serves/JVM-does-not divergence. It is not a divergence at all — **neither backend advertises it**, for this reason.

That matters for how it gets fixed. #193 is an open owner decision about whether the JVM backend should reach parity with native. **This is not part of that decision**: it is a one-line implementation defect on the native side, it needs no ruling, and it is fixable independently of whichever way #193 goes.

## What is and is not affected

- **Behaviour is unaffected.** `Method.NoReply`'s functional half works: the adaptor sends no reply (`WireDbusObject.kt:417`, `WireServe.kt:91`) and the generated proxy sets `dontExpectReply = true`. Only the *advertised metadata* is missing.
- **The consumer is any external introspector** — `busctl`, d-feet, another language's bindings. A method that behaves as fire-and-forget while describing itself as an ordinary method is the more confusing state, not the less: a client reading introspection will wait for a reply that is never coming.
- This is the same class as #159 (annotations parsed then dropped), one level lower down.

## Fix

[...]

PR #195, opened 2026-08-07
docs: correct the served-introspection claim for #159 (measured on both backends)

monkopedia-coder:
Refs #193 — the correction half of the split. **Documentation only.** No generator behavior changes, no runtime changes, no API changes.

## The problem

`CHANGELOG.md`'s `[Unreleased]` #159 entry (landed two days ago via #187) told consumers that mapping D-Bus annotations onto vtable flags makes a running adaptor's served introspection agree with the XML it was generated from. `AdaptorGenerator.kt`'s KDoc carried the same sentence. A release is pending and that entry is consumer-facing release-note material, so it should not ship as written.

## I verified this empirically, on both backends

The audit on #193 established the JVM half by code reading and *inferred* the native half from sd-bus's vtable contract. Rather than replace one unverified claim with another, I ran the round-trip: a temporary probe test exported an object with `interfaceFlags { isDeprecated = true }`, a `hasNoReply` method, a deprecated method, and three properties carrying `CONST_PROPERTY_VALUE` / `EMITS_INVALIDATION_SIGNAL` / `EMITS_NO_SIGNAL`, then called `org.freedesktop.DBus.Introspectable.Introspect` on it under `dbus-run-session` — once as `linuxX64Test`, once as `jvmTest`. The probe was deleted before committing; the diff here is documentation only.

Measured result:

| Annotation | native (sd-bus) | JVM (wire) |
| --- | --- | --- |
| `Deprecated` on method / signal / property | served | served |
| `Deprecated` on the **interface** | served | **not served** |
| `EmitsChangedSignal` → `const` / `invalidates` / `false` | served | **not served** |
| `org.freedesktop.DBus.Method.NoReply` | **not served** | **not served** |

Native XML, verbatim (`Fire` is the `hasNoReply` method):

```xml
<interface name="com.monkopedia.sdbus.probe534028.Iface">
 <annotation name="org.freedesktop.DBus.Deprecated" value="true"/>
 <method name="Fire">
 </method>
 <method name="OldMethod">
  <annotation name="org.freedesktop.DBus.Deprecated" value="true"/>
 </method>
 <property name="ConstProp" type="s" access="read">
  <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="const"/>
 </property>
 <property name="InvalidatingProp" type="s" access="read">
  <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="invalidates"/>
 </property>
 <property name="SilentProp" type="s" access="read">
[...]

Issue #193, opened 2026-08-07
#159's annotation→vtable-flag fix is native-only: the JVM backend reads no `Flags` at all, so served introspection still disagrees with the XML


monkopedia-coder:
Found by the Stream-2 (docs & contracts) audit pass, checking the claims that landed with #187
(`81acaee`). The generator change is correct; the claim that it fixes the served introspection is
true on native and **false on JVM**, because the JVM wire backend has no reader for the flag types
the generator now emits.

All checks against `main` @ `c36abb5`.

## The claims

`codegen/src/main/kotlin/AdaptorGenerator.kt:136-143`:

> Emits the `addVTable` block, carrying the standard D-Bus annotations through to the vtable
> flags **so the introspection the running adaptor serves matches the XML it was generated from**:
> `org.freedesktop.DBus.Deprecated` on the **interface**/method/signal/property, and
> `org.freedesktop.DBus.Method.NoReply` on the method.

`CHANGELOG.md` (Unreleased, #159):

> Previously all three annotations were parsed and then dropped, so the introspection a running
> adaptor served could disagree with the XML it was generated from.

> `Deprecated` is advisory: it changes only the introspection metadata the adaptor publishes.

## The check

**`Flags` has zero readers in `src/jvmMain`.**

```
$ grep -rn "PropertyUpdateBehaviorFlags\|GeneralFlags" --include='*.kt' src/jvmMain/
(no output — exit 1)

# positive control, same grep against nativeMain:
$ grep -rn "PropertyUpdateBehaviorFlags\|GeneralFlags" --include='*.kt' src/nativeMain/ | wc -l
8
```

Native consumes them in `Flags.native.kt` (`toSdBusMethodFlags`/`toSdBusPropertyFlags`/
`toSdBusWritablePropertyFlags`), which is how sd-bus turns them back into introspection XML.

**What the JVM backend actually reads off a vtable item** — `WireServe.buildMeta`,
`WireServe.kt:536-566`:

```kotlin
is MethodVTableItem   -> … deprecated = item.isDeprecated
is PropertyVTableItem -> … deprecated = item.isDeprecated
is SignalVTableItem   -> … deprecated = item.isDeprecated
else -> Unit
```

and what it emits, `WireServe.appendInterface` (`WireServe.kt:279-305`): a single
`DEPRECATED_ANNOTATION` constant (`WireServe.kt:358`), applied per method / signal / property only.

So on the JVM backend:

[...]

PR #213, opened 2026-08-07
test: assert what a served adaptor actually advertises in introspection (Refs #193)

monkopedia-coder:
Adds a cross-backend introspection round-trip test: register vtable items with known flags, serve the object, call `org.freedesktop.DBus.Introspectable.Introspect` against it, and assert on the XML that came back. It lives in `commonTest`, so the identical assertions execute under both `:jvmTest` and `:linuxX64Test`.

Refs #193 — deliberately **not** `Fixes`. #193 is an open owner decision about whether the JVM backend should reach parity with native, and this does not pre-empt it either way.

## Why

Nothing in this repo asserted what a running adaptor actually **serves** in its introspection. The codegen goldens verify what the *generator emits*; that is a different thing, and the gap between them has been expensive:

- A merged CHANGELOG asserted a cross-backend round-trip that does not happen (corrected in #195).
- #193 was wrong about a third of its own scope — it listed `Method.NoReply` as a native-serves/JVM-does-not divergence when neither backend advertises it.
- #197 (`METHOD_NO_REPLY` OR-ed with a literal `0u` on native) went unnoticed.

An ad-hoc version of exactly this test was written and thrown away twice during #195's work. This makes it permanent and reusable.

## Shape

`IntrospectionRoundTrip.kt` is the instrument: `introspectServedObject(vararg InterfaceName to VTableBuilder.() -> Unit)` serves an object on a real bus, introspects it, tears everything down, and returns the XML wrapped in a small `IntrospectionXml` accessor. Locating `<tag name="…">` and taking everything up to the next `</tag>` is exact for D-Bus introspection (it never nests an element inside another of the same tag), so no XML parser joins the test classpath.

`assertAnnotation(..., served: Boolean)` pins **both** polarities, so a backend that does not serve a flag is asserted to not serve it rather than merely not checked.

## The asymmetries are expressed, not hidden

The backends genuinely differ, so this pins each one rather than asserting a shared contract it would have to fake. Three `expect`/`actual` capability flags carry the divergence, in the idiom `FdTestSupport.supportsFdDuplicationSemantics` already established:

| Annotation | Native | JVM |
| --- | --- | --- |
| `Deprecated` on method / signal / property | served | served |
| `Deprecated` on the **interface** | served | not served |
| `EmitsChangedSignal` (`const` / `invalidates` / `false`) | served | not served |
| `Method.NoReply` | not served | not served |

This reproduces, on both backends, the table measured during #195 — the same table that justified the merged doc correction. Member-level `Deprecated` needs no capability flag because both backends serve it; the other three get one each, and the `actual` values are now the machine-checked record of the divergence rather than prose.

`backendServesMethodNoReply` is documented as the odd one out: `false` on native is a **defect** (#197), not a capability. Flipping that one `actual` to `true` is the red-first proof for that fix.

## What this makes cheap

Either disposition of #193 gets easier:

- If the owner decides the JVM backend **should** reach parity, the flags flip to `true` one at a time and each flip is a red-first guard for the work that follows.
- If the owner decides to **document** the difference instead, this file is the truthful record of what each backend does, checked on every CI run, so the documentation cannot drift away from the behavior again.

## Verification

`dbus-run-session -- ./gradlew :jvmTest :linuxX64Test`, plus `compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64 ktlintCheck apiCheck`. Test-only change; `apiCheck` does not move and the klib ABI consumed by `blue-falcon-sdbus` is untouched.

## How the numbers were measured

Gradle's `BUILD SUCCESSFUL` does not distinguish a passing suite from one where every task resolved UP-TO-DATE, and the JUnit XML persists across runs, so neither the build status nor a bare test count proves anything ran. Every figure below comes from: **delete `build/test-results/` → run with `--rerun-tasks` → confirm no result file predates the run → count**.

- `dbus-run-session -- ./gradlew --rerun-tasks :jvmTest :linuxX64Test`, `11 actionable tasks: 11 executed` (zero up-to-date).
- Result files older than the run start: **0 of 66**.
- **331 test executions on `:jvmTest`, 301 on `:linuxX64Test`, 0 failures.**
- `VtableFlagIntrospectionTest` contributes **4 executions on each backend, 0 skipped** — so all four rows of the table above were actually exercised on both, not silently filtered away.

That matters more than usual here: this table is a first baseline, and a baseline recorded from a run that did nothing would agree with anything.

All tracked `.kt`/`.kts`/`.xml`/`.md` files on this branch were checked for embedded NUL bytes (the #161 failure mode, which makes a file vanish from `grep` silently) — clean, with a positive control confirming the detector fires.

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/IntrospectionRoundTrip.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/IntrospectionRoundTrip.kt
@@ -1,0 +1,116 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.MethodName
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.Resource
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.VTableBuilder
+import com.monkopedia.sdbus.VTableItem
+import com.monkopedia.sdbus.callMethod
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import kotlin.random.Random
+import kotlin.test.assertEquals
+import kotlin.test.fail
+
+/** The standard D-Bus annotation marking an element deprecated. */
+internal const val DEPRECATED_ANNOTATION = "org.freedesktop.DBus.Deprecated"
+
+/** The standard D-Bus annotation marking a method as expecting no reply. */
+internal const val METHOD_NO_REPLY_ANNOTATION = "org.freedesktop.DBus.Method.NoReply"
+
+/** The standard D-Bus annotation declaring how a property signals changes. */
+internal const val EMITS_CHANGED_SIGNAL_ANNOTATION =
+    "org.freedesktop.DBus.Property.EmitsChangedSignal"
+
+/**
+ * Serves an object carrying [vtables] on a real bus, calls
+ * `org.freedesktop.DBus.Introspectable.Introspect` against it, and returns the XML it advertised.
+ *
+ * This is the instrument for asserting what a *running adaptor serves*, as opposed to what the
+ * codegen goldens already cover — what the *generator emits*. Nothing tested the former until
+ * [VtableFlagIntrospectionTest], which is how #197 survived. It lives in `commonTest` so every
+ * assertion built on it runs against both backends.
+ *
+ * The connections, object and registrations are all torn down before returning; only the XML
+ * escapes.
+ */
+internal suspend fun introspectServedObject(
+    vararg vtables: Pair<InterfaceName, VTableBuilder.() -> Unit>
+): IntrospectionXml {
+    val id = Random.nextInt(100_000, 999_999)
+    val service = ServiceName("com.monkopedia.sdbus.introspect$id")
+    val path = ObjectPath("/com/monkopedia/sdbus/introspect$id")
+
+    val server = createBusConnection(service)
+    val client = createBusConnection()
+    val obj = createObject(server, path)
+    val registrations = vtables.map { (name, vtable) ->
+        obj.addVTable(name, buildList<VTableItem> { VTableBuilder(this).vtable() })
+    }
+    server.startEventLoop()
+    val proxy = createProxy(client, service, path, runEventLoopThread = false)
+
+    try {
+        return IntrospectionXml(
+            proxy.callMethod<String>(
+                InterfaceName("org.freedesktop.DBus.Introspectable"),
+                MethodName("Introspect")
+            ) { }
+        )
+    } finally {
+        registrations.forEach(Resource::release)
+        proxy.release()
+        obj.release()
+        server.stopEventLoop()
+        client.release()
+        server.release()
+    }
+}
+
+/**
+ * The introspection XML a live adaptor advertised, with just enough structure to assert on a single
+ * element's annotations.
+ *
+ * D-Bus introspection never nests an element inside another of the same tag, so locating
+ * `<tag name="...">` and taking everything up to the following `</tag>` isolates exactly that
+ * element -- which is why this needs no XML parser on the test classpath.
+ */
+internal class IntrospectionXml(val raw: String) {
+
+    /**
+     * The children of `<[tag] name="[name]">`, or the empty string when the element is
+     * self-closing and so carries no annotations at all.
+     */
+    fun childrenOf(tag: String, name: String): String {
+        val open = raw.indexOf("<$tag name=\"$name\"")
+        if (open < 0) fail("no <$tag name=\"$name\"> in the served introspection:\n$raw")
+        val openEnd = raw.indexOf('>', open)
+        if (raw[openEnd - 1] == '/') return ""
+        val close = raw.indexOf("</$tag>", openEnd)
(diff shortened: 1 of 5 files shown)


monkopedia-reviewer (approved):
## Verdict: approve — tier 1 (test_only)

Reviewed head `09c0c14bf94ed69c9dc51ac1ff4599b24bdad0d1` against `main` (`f9198d5`) in an isolated checkout. All verification below is from my own from-scratch builds under the fleet build-slot, not from the PR body's claims.

### 1. Ground-truth table — reproduced on both backends, verified against production source

Read `IntrospectionRoundTrip.kt`, `VtableFlagIntrospectionTest.kt`, and both `TestBackendCapabilities.{jvm,native}.kt`, then cross-checked every `actual` against the real serving code (not just internal self-consistency):

| Annotation | Native | JVM | Verified against |
|---|---|---|---|
| `Deprecated` on method/signal/property | served | served | both backends emit it unconditionally per member |
| `Deprecated` on the interface | served | not served | `Flags.native.kt:toSdBusInterfaceFlags` ORs `SD_BUS_VTABLE_DEPRECATED`; `WireServe.kt:appendInterface` never writes an interface-level annotation, only per-member |
| `EmitsChangedSignal` | served | not served | `Flags.native.kt:testEmitsFlags` maps const/invalidates/no-signal; `WireServe.kt` has no property-update-behavior handling at all |
| `Method.NoReply` | not served | not served | `Flags.native.kt:toSdBusMethodFlags` line 61-62 ORs `METHOD_NO_REPLY` with a literal `0u` instead of `SD_BUS_VTABLE_METHOD_NO_REPLY` (the documented #197 defect); `WireServe.kt` has no NoReply-to-XML path |

Ran `dbus-run-session -- ./gradlew --rerun-tasks :jvmTest :linuxX64Test` fresh (test-results deleted first): `VtableFlagIntrospectionTest` = **4/4 passing, 0 skipped on both `:jvmTest` and `:linuxX64Test`**. All four table rows executed and matched on both backends — the table is accurate and machine-checked.

### 2. Can the harness fail? Yes — confirmed red→green

Mutated production (not the test): commented out `sdbusFlags = testEmitsFlags(sdbusFlags)` in `Flags.native.kt:toSdBusPropertyFlags`, i.e. stopped native from serving `EmitsChangedSignal`, without touching the `backendServesEmitsChangedSignal = true` capability flag.

- **RED:** `VtableFlagIntrospectionTest.emitsChangedSignalIsServedOnlyWhereTheBackendCarriesIt[linuxX64] FAILED — kotlin.AssertionError`. (Also broke 6 unrelated pre-existing tests that depend on `EMITS_CHANGE_SIGNAL`-driven `PropertiesChanged` behavior — expected collateral of disabling the whole flag path, not a defect in the PR.)
- Reverted the edit; `git status --porcelain` came back empty (byte-identical to the merge-base file).
- **GREEN again:** re-ran `:linuxX64Test` fresh — `301 tests, 0 failures, 0 skipped`; `VtableFlagIntrospectionTest` 4/4, 0 skipped.

The harness is a live gate, not vacuous.

### 3. Asymmetries expressed, not hidden

Confirmed by reading + cross-checking against source (not just the test file): the three genuine divergences (`backendServesInterfaceLevelDeprecated`, `backendServesEmitsChangedSignal`, `backendServesMethodNoReply`) are each carried by an `expect`/`actual` capability flag with a comment citing the exact production code responsible, in the same idiom as the pre-existing `backendDeliversDirectedSignalsUnicast`. No assertion is weakened or made conditional to paper over a backend difference — `assertAnnotation` always pins both polarities (`served: Boolean`), and every capability-flag value I checked matches what the corresponding backend actually does.

### 4. Executed counts (delete-then-run, not `BUILD SUCCESSFUL`)

`rm -rf build/test-results && dbus-run-session -- ./gradlew --rerun-tasks :jvmTest :linuxX64Test` → `11 actionable tasks: 11 executed` (zero up-to-date), then counted straight from the fresh JUnit XML:

- `:jvmTest` — **331 tests, 0 skipped, 0 failures** (matches PR body exactly)
- `:linuxX64Test` — **301 tests, 0 skipped, 0 failures** (matches PR body exactly)

### 5. apiCheck / Refs

- `./gradlew --rerun-tasks apiCheck` — ran explicitly (`21 actionable tasks: 21 executed`, not up-to-date), `BUILD SUCCESSFUL`. No `api/*.api` diff — matches the test-only claim.
- `ktlintCheck` — green.
- PR body: **"Refs #193 — deliberately not Fixes."** Confirmed verbatim in the PR description.

### Classification

`test_only` (5 files, all under `commonTest`/`jvmTest`/`nativeTest`, +305/-0, zero production code touched — confirmed by diff stat and by the fact my production mutation was the only way to make anything go red). Default tier 1; sdbus-kotlin has no override that raises `test_only`. No reviewer-judgment bump — the change is well-scoped, internally consistent with the existing `expect`/`actual` capability idiom (`FdTestSupport`/`backendDeliversDirectedSignalsUnicast`), and I found nothing surprising or more load-bearing than the category suggests.

### Note for the record

While waiting on the build slot I observed a holder labeled `reviewer-sdbus-213` briefly claiming it for "introspection harness full verify (baseline+mutations)" — consistent with a resurfaced instance of the reviewer that reportedly stalled earlier on this same PR. I checked before posting: `gh pr view 213 --json reviews` returned zero reviews and the PR is still OPEN/CLEAN, so nothing conflicting has landed. Flagging in case that instance is still alive and posts a second review after this one.

PR #215, opened 2026-08-07
fix(native): stop swapping ALLOW_REPLACEMENT and REPLACE_EXISTING in requestName

### LINKED ISSUE #212: requestName swaps ALLOW_REPLACEMENT and REPLACE_EXISTING on native; the comment above it states the mapping backwards
`requestName` on the **native** backend swaps `ALLOW_REPLACEMENT` and `REPLACE_EXISTING`. Asking to *allow* another owner to take the name instead asks to *steal* it from the current owner, and vice versa. The JVM backend is correct, so this is also a cross-backend divergence.

## The bug

`src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt:280-281`:

```kotlin
if (RequestNameFlag.ALLOW_REPLACEMENT in flags) sdbusMask = sdbusMask or 0x1u
if (RequestNameFlag.REPLACE_EXISTING in flags) sdbusMask = sdbusMask or 0x2u
```

`/usr/include/systemd/sd-bus-protocol.h:172-174`:

```c
SD_BUS_NAME_REPLACE_EXISTING  = 1ULL << 0,   /* 0x1 */
SD_BUS_NAME_ALLOW_REPLACEMENT = 1ULL << 1,   /* 0x2 */
SD_BUS_NAME_QUEUE             = 1ULL << 2    /* 0x4 */
```

So `ALLOW_REPLACEMENT` → `0x1` = `SD_BUS_NAME_REPLACE_EXISTING`, and `REPLACE_EXISTING` → `0x2` = `SD_BUS_NAME_ALLOW_REPLACEMENT`. Both are wrong, in opposite directions.

## The comment above it asserts the opposite, and is why this survived

`ConnectionImpl.kt:274` and `:277`:

> `// … ALLOW_REPLACEMENT (0x1) / REPLACE_EXISTING (0x2) coincide; the queue bit must be inverted.`
> `// sd-bus flag bits: SD_BUS_NAME_ALLOW_REPLACEMENT=0x1, SD_BUS_NAME_REPLACE_EXISTING=0x2, …`

Both sentences are false against the header. sd-bus's first two bits are the **inverse** of the D-Bus wire bits (`dbus-shared.h:110-112`: `ALLOW_REPLACEMENT 0x1`, `REPLACE_EXISTING 0x2`), exactly like the queue bit the comment *does* handle correctly. Only `QUEUE`/`DO_NOT_QUEUE` at `0x4` coincides positionally — which is presumably where the "coincide" reading came from.

The comment was written specifically to prevent this class of error, and states the mapping backwards. That is what makes it worse than an unannotated bug: a reader checking the code against the comment finds agreement.

## Blast radius

- **Native only.** The JVM backend is correct — `Connection.jvm.kt:30` folds `flag.mask` and passes the *wire* bits straight to the daemon, which is right, because the daemon takes wire bits.
- A native service requesting its name with `ALLOW_REPLACEMENT` (a cooperative "you may take this from me") instead performs a hostile takeover of whoever holds it. A service using `REPLACE_EXISTING` to deliberately take over fails to, and instead marks itself replaceable.
- The queue bit is handled correctly (`:282`), so `DO_NOT_QUEUE` behaviour is unaffected.

## Why no test caught it

`grep` for `ALLOW_REPLACEMENT|REPLACE_EXISTING` across `src/`, `cross_test/` and `stress_test/` finds only the enum declaration and this mapping — **no test anywhere exercises either flag**. Positive control: `DO_NOT_QUEUE` *does* appear in a test (`CommonApiIntegrationTest.kt:1156`), so the grep is capable of finding usages; these two genuinely have none.

Introduced with the flag support itself — `git log -L` points at `7d247c4`, "requestName reply/flags [#112] (#126)".

## Fix

Swap the two constants and correct both comment lines. Prefer referencing the cinterop-imported `SD_BUS_NAME_*` constants over hardcoded hex, so the binding is checked by the compiler rather than by a comment — the same reasoning as #178/#180, where a literal `0u` stood in for an unbound constant (see #197 for that instance).

## Verification

Red-first is achievable and needs no hardware: two connections on a session bus, the first requesting a name with `ALLOW_REPLACEMENT`, the second requesting the same name with `REPLACE_EXISTING`, asserting ownership transfers. It should pass on JVM and fail on native today. Run under `dbus-run-session -- ./gradlew linuxX64Test jvmTest`.

That test does not exist yet in any form, and its absence is the reason this shipped.

## Provenance

Found by the 2026-08-07 read-only claim-verification pass, auditing `src/nativeMain` comments for claims that assert behaviour. Verified independently against the installed `sd-bus-protocol.h` and `dbus-shared.h` before filing.

monkopedia-coder:
Fixes #212

`Connection.requestName` takes `RequestNameFlag`s that carry the **D-Bus wire** bits. The C `sd_bus_request_name` takes sd-bus's **own** flag enum, and the two are not the same numbering:

```c
/* sd-bus-protocol.h:172-174 */
SD_BUS_NAME_REPLACE_EXISTING  = 1ULL << 0,
SD_BUS_NAME_ALLOW_REPLACEMENT = 1ULL << 1,
SD_BUS_NAME_QUEUE             = 1ULL << 2
```

The native mapping wrote `ALLOW_REPLACEMENT → 0x1` and `REPLACE_EXISTING → 0x2`, i.e. each flag into the other's bit. The consequence is exactly the inverse of the documented behavior: a name whose owner had *allowed replacement* could not be taken over, and a caller asking to *replace the existing owner* was queued behind it instead.

## Why it survived

The comment sitting directly above the code asserted the wrong mapping as fact:

> `ALLOW_REPLACEMENT (0x1) / REPLACE_EXISTING (0x2) coincide; the queue bit must be inverted.`

Only the queue bit was actually checked against the header. The other two were asserted to coincide and then encoded as hex literals, so nothing could catch the error — not the compiler, not a reader trusting the comment. That comment is corrected here, and more usefully the mapping now **names the cinterop `SD_BUS_NAME_*` constants** rather than hex, so the binding is compiler-checked. That is the same lesson as #197: a constant that is spelled out in a comment instead of imported is a constant nothing verifies.

The `SD_BUS_NAME_*` constants are already exposed by the existing cinterop bindings (verified in the generated `package_sdbus` linkdata alongside `SD_BUS_VTABLE_DEPRECATED`, which is already imported), so no `.def` change is needed.

## Not a convention to choose

The public KDoc on `RequestNameFlag` (`Connection.kt:179-183`) already documents the correct semantics and cites the right `DBUS_NAME_FLAG_*` names, and the JVM backend passes the wire bits straight to the daemon and honours them. So this is the native backend disagreeing with its own documented contract — there was no second defensible reading to pick between.

## Red-first

A cross-backend regression sits beside the existing `requestName_reportsConsistentReplyCodesAcrossBackends` in `CommonApiIntegrationTest`, reusing its fixtures: the incumbent takes the name with `ALLOW_REPLACEMENT`, the challenger takes it with `REPLACE_EXISTING`, and the daemon must report the challenger as `PRIMARY_OWNER`. Because it lives in `commonTest` it runs on both backends — it passed on `:jvmTest` and failed on `:linuxX64Test` before the fix, which is the shape the issue predicted.

Before the fix, `:linuxX64Test`:

```
com.monkopedia.sdbus.integration.CommonApiIntegrationTest.requestName_replaceExistingTakesOverANameThatAllowedReplacement[linuxX64] FAILED
    kotlin.AssertionError at null:-1

kotlin.AssertionError: Expected <PRIMARY_OWNER>, actual <IN_QUEUE>.

1 test completed, 1 failed
```

## Verification

`dbus-run-session -- ./gradlew :jvmTest :linuxX64Test`, plus `compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64 ktlintCheck apiCheck`. `apiCheck` does not move — the change is inside an `internal` implementation and the public enum is untouched, so the klib ABI consumed by `blue-falcon-sdbus` is unchanged.

## One thing the fix surfaced that the comment could not

Writing the accumulator as `ULong` failed to compile:

```
e: ConnectionImpl.kt:283:38 Argument type mismatch: actual type is 'UInt', but 'ULong' was expected.
```

cinterop types the `SD_BUS_NAME_*` constants as `UInt` (the enum's largest value is 4), unlike the `SD_BUS_VTABLE_*` constants which are `ULong`. The hex-literal version could never have surfaced that, because `0x1u` is whatever the accumulator says it is. That is the argument for naming the constants in miniature.

## How the numbers were measured

`BUILD SUCCESSFUL` cannot distinguish a passing suite from an all-UP-TO-DATE one, and JUnit XML persists across runs, so a stale result file is indistinguishable from a fresh one by content. Both states below were measured by **deleting `build/test-results/`, running with `--rerun-tasks`, confirming no result file predates the run, then counting**.

**Before the fix** (production change reverted, new test in place) — `11 actionable tasks: 11 executed`, 0 stale result files:

- `:linuxX64Test` — 298 executions, **1 failure** (the one above)
- `:jvmTest` — 328 executions, 0 failures — the new test passes on JVM unchanged, which is the cross-backend half of the claim

**After the fix** — `11 actionable tasks: 11 executed`, 0 stale files:

- `:linuxX64Test` — **298 executions, 0 failures**
- `:jvmTest` — **328 executions, 0 failures**

Branch checked clean of embedded NUL bytes (the #161 failure mode) with a verified positive control.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,16 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   as ordinary request/reply. The JVM backend still does not serve this annotation; that is part of
   #193. (#197)
 
+- **`RequestNameFlag.ALLOW_REPLACEMENT` and `RequestNameFlag.REPLACE_EXISTING` were swapped on the
+  native backend.** `Connection.requestName` translates the D-Bus wire bits the enum carries into
+  sd-bus's own flag enum, where the two are transposed (`SD_BUS_NAME_REPLACE_EXISTING` is bit 0 and
+  `SD_BUS_NAME_ALLOW_REPLACEMENT` is bit 1), and the mapping asserted they coincided. So on native a
+  request that meant "I allow replacement" reached sd-bus as "replace the current owner" and vice
+  versa: a name whose owner had allowed replacement could not be taken over, and the challenger was
+  queued instead. The JVM backend hands the wire bits to the daemon directly and always matched the
+  semantics documented on `RequestNameFlag`, so this affected native only. The mapping now names the
+  cinterop `SD_BUS_NAME_*` constants, making it compiler-checked rather than comment-checked. (#212)
+
 ## [1.0.1] - 2026-07-15
 
 A maintenance release: refreshed external dependencies to their latest stable versions. There are
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/CommonApiIntegrationTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/CommonApiIntegrationTest.kt
@@ -1164,6 +1164,38 @@ class CommonApiIntegrationTest {
         }
     }
 
+    // Regression for #212: ALLOW_REPLACEMENT and REPLACE_EXISTING are transposed between the D-Bus
+    // wire bits RequestNameFlag carries and sd-bus's own flag enum, and the native backend used to
+    // map them straight across — so a name its owner had marked replaceable could not be taken over,
+    // and the second caller was queued instead. The JVM backend passes the wire bits to the daemon
+    // directly and always matched the documented semantics; this runs on BOTH backends.
+    @Test
+    fun requestName_replaceExistingTakesOverANameThatAllowedReplacement() {
+        val ids = uniqueFixtureIds("replaceName")
+        val name = ServiceName("${ids.service.value}.Replaceable")
+        val incumbent = createBusConnection()
+        val challenger = createBusConnection()
+
+        try {
+            assertEquals(
+                RequestNameReply.PRIMARY_OWNER,
+                incumbent.requestName(name, RequestNameFlag.ALLOW_REPLACEMENT)
+            )
+
+            // The incumbent allowed replacement, so REPLACE_EXISTING must hand ownership over
+            // rather than queue behind it.
+            assertEquals(
+                RequestNameReply.PRIMARY_OWNER,
+                challenger.requestName(name, RequestNameFlag.REPLACE_EXISTING)
+            )
+        } finally {
+            runCatching { challenger.releaseName(name) }
+            runCatching { incumbent.releaseName(name) }
+            incumbent.release()
+            challenger.release()
+        }
+    }
+
     // Coverage for #110: the direct two-arg property accessors on Proxy
     // (getPropertyAsync / setPropertyAsync / getAllProperties / getAllPropertiesAsync). Confirms the
     // async variants agree with the sync getProperty/setProperty results and that the Variant unwrap
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/internal/ConnectionImpl.kt
@@ -97,6 +97,9 @@ import platform.posix.poll
 import platform.posix.pollfd
 import platform.posix.uint64_tVar
 import platform.posix.usleep
+import sdbus.SD_BUS_NAME_ALLOW_REPLACEMENT
+import sdbus.SD_BUS_NAME_QUEUE
+import sdbus.SD_BUS_NAME_REPLACE_EXISTING
 import sdbus._SD_BUS_MESSAGE_TYPE_INVALID
 import sdbus.sd_bus_error
 import sdbus.sd_bus_interface_name_is_valid
@@ -269,17 +272,22 @@ internal class ConnectionImpl(private val sdbus: ISdBus, private val bus: BusPtr
         checkNotReleased()
         checkServiceName(name.value)
 
-        // RequestNameFlag carries the D-Bus WIRE bits (DO_NOT_QUEUE=0x4); the C `sd_bus_request_name`
-        // takes sd-bus's OWN flag enum, where bit 0x4 is SD_BUS_NAME_QUEUE — the INVERSE (opt-in
-        // queueing). ALLOW_REPLACEMENT (0x1) / REPLACE_EXISTING (0x2) coincide; the queue bit must be
-        // inverted. Feeding the wire bits straight in would flip queueing and diverge from the JVM
-        // backend, which passes the wire flags to the daemon directly.
-        // sd-bus flag bits: SD_BUS_NAME_ALLOW_REPLACEMENT=0x1, SD_BUS_NAME_REPLACE_EXISTING=0x2,
-        // SD_BUS_NAME_QUEUE=0x4 (set when NOT DO_NOT_QUEUE, i.e. the inverted queue bit).
+        // RequestNameFlag carries the D-Bus WIRE bits; the C `sd_bus_request_name` takes sd-bus's
+        // OWN flag enum, and no bit of it lines up with the wire: replacement and allow-replacement
+        // are transposed, and sd-bus's queue bit is the INVERSE of the wire's DO_NOT_QUEUE (opt-in
+        // rather than opt-out). Feeding the wire bits straight in would diverge from the JVM
+        // backend, which passes them to the daemon directly. Naming the cinterop constants keeps
+        // this mapping compiler-checked rather than comment-checked (#212).
         var sdbusMask = 0u
-        if (RequestNameFlag.ALLOW_REPLACEMENT in flags) sdbusMask = sdbusMask or 0x1u
-        if (RequestNameFlag.REPLACE_EXISTING in flags) sdbusMask = sdbusMask or 0x2u
-        if (RequestNameFlag.DO_NOT_QUEUE !in flags) sdbusMask = sdbusMask or 0x4u
+        if (RequestNameFlag.ALLOW_REPLACEMENT in flags) {
+            sdbusMask = sdbusMask or SD_BUS_NAME_ALLOW_REPLACEMENT
+        }
+        if (RequestNameFlag.REPLACE_EXISTING in flags) {
+            sdbusMask = sdbusMask or SD_BUS_NAME_REPLACE_EXISTING
+        }
+        if (RequestNameFlag.DO_NOT_QUEUE !in flags) {
+            sdbusMask = sdbusMask or SD_BUS_NAME_QUEUE
+        }
         val r = sdbus.sd_bus_request_name(bus.value, name.value, sdbusMask.convert())
         // sd_bus_request_name collapses the four RequestName reply codes onto its return value:
         //   PRIMARY_OWNER -> 1, IN_QUEUE -> 0, EXISTS -> -EEXIST, ALREADY_OWNER -> -EALREADY.


monkopedia-reviewer (commented):
## Reviewed & sound

Independently verified everything bot-verifiable; the PR is tier-1-clean. Summary:

**Premise (read the header myself, not taken on faith):** `/usr/include/systemd/sd-bus-protocol.h:172-174` —
```c
SD_BUS_NAME_REPLACE_EXISTING  = 1ULL << 0,   /* 0x1 */
SD_BUS_NAME_ALLOW_REPLACEMENT = 1ULL << 1,   /* 0x2 */
```
and `/usr/include/dbus-1.0/dbus/dbus-shared.h:110-111` — `DBUS_NAME_FLAG_ALLOW_REPLACEMENT 0x1`, `DBUS_NAME_FLAG_REPLACE_EXISTING 0x2`. The pre-fix code wrote `ALLOW_REPLACEMENT → 0x1`/`REPLACE_EXISTING → 0x2`, i.e. the sd-bus bits, not the transposition — confirmed swapped.

**Red-first, reproduced myself.** Checked out head `e6a0192`, then reverted only `ConnectionImpl.kt` to `origin/main`'s pre-fix version while keeping the new regression test. Deleted `build/test-results`, ran `dbus-run-session -- ./gradlew :jvmTest :linuxX64Test --rerun-tasks` end to end (`11 actionable tasks: 11 executed`, fresh XML timestamped seconds after the run). Result, byte-for-byte the claimed signature:
```
com.monkopedia.sdbus.integration.CommonApiIntegrationTest.requestName_replaceExistingTakesOverANameThatAllowedReplacement[linuxX64] FAILED
kotlin.AssertionError: Expected <PRIMARY_OWNER>, actual <IN_QUEUE>.
```
`linuxX64Test`: 298 tests, 1 failure. `jvmTest`: 328 tests, 0 failures (new test passes there unchanged) — matches the PR's claimed pre-fix counts exactly. Restored `ConnectionImpl.kt` to the PR's fix afterward (`git status --porcelain` empty).

**Named constants, not hex.** Diff imports `sdbus.SD_BUS_NAME_ALLOW_REPLACEMENT`, `sdbus.SD_BUS_NAME_REPLACE_EXISTING`, `sdbus.SD_BUS_NAME_QUEUE` and ORs those into `sdbusMask` — confirmed in the diff, and `compileKotlinLinuxX64`/`compileKotlinLinuxArm64` both succeed at head, so the binding is compiler-checked as claimed. (Did not independently reproduce the UInt-vs-ULong anecdote about the accumulator type — not load-bearing to correctness, and the compile already succeeds either way.)

**Comment fixed too.** `ConnectionImpl.kt:274-279` now states the real relationship ("no bit of it lines up with the wire: replacement and allow-replacement are transposed...") instead of the wrong "coincide" claim. No corrected-hex-only patch that would leave the trap in place.

**Queue bit untouched.** `DO_NOT_QUEUE`/`SD_BUS_NAME_QUEUE` logic is unchanged (same condition, now via the named constant). Its existing coverage, `requestName_reportsConsistentReplyCodesAcrossBackends`, **passed even in my pre-fix red run** — confirming the queue bit was already correct and the fix doesn't touch it. It also passes post-fix.

**Full counts, anti-stale methodology.** For both the pre-fix and post-fix states: delete `build/test-results/`, run with `--rerun-tasks`, confirm no stale result predates the run (fresh `stat` timestamps checked each time), then count from XML via `xml.etree`, not by eyeballing console output.
- Post-fix (head `e6a0192`): `:jvmTest` 328/0 failures, `:linuxX64Test` 298/0 failures. Full `compileKotlinJvm compileKotlinLinuxX64 compileKotlinLinuxArm64 ktlintCheck apiCheck :jvmTest :linuxX64Test --rerun-tasks` → `BUILD SUCCESSFUL`, `77 actionable tasks: 77 executed`.
- Pre-fix (repro): `:jvmTest` 328/0, `:linuxX64Test` 298/1 (the one failure above).

**apiCheck.** Ran clean (`:jvmApiCheck`, `:klibApiCheck`, `:apiCheck` all succeeded) — and more directly, the PR's file list is exactly `CHANGELOG.md`, `src/commonTest/.../CommonApiIntegrationTest.kt`, `src/nativeMain/.../ConnectionImpl.kt`; no `api/*.api` file is touched, and the changed production code is `internal`. The klib ABI `blue-falcon-sdbus` consumes is unchanged.

**Branch hygiene.** `git merge-base e6a0192 origin/main` == `origin/main`'s tip (`f9198d5`) — single commit on top of current main, confirmed independent of #213/#214 rather than assumed.

## The one thing I'm surfacing rather than deciding myself

Everything above is sound and this would clear tier 1 as an ordinary `bug_fix`. I'm bumping one notch under the reviewer-judgment lever ("a change tagged `bug_fix` actually mutates observable behavior in a non-trivial way → bump to tier 2"): this flips real D-Bus name-ownership semantics (`ALLOW_REPLACEMENT`/`REPLACE_EXISTING`) on the **native** backend of a published Maven artifact (`linuxX64`/`linuxArm64`) — anyone who called `requestName` with those flags on native got the inverse of what they asked for until now, and after this merges they'll get the (correct, already-documented) behavior instead. That's a real behavior change shipping to consumers, not an internal-only refactor, even though it's a pure bug fix with no API surface change.

**Question:** approve and merge as-is, or would you like anything else (e.g. a patch release) attached to this fix?

If approved, the merge command is:
```
/home/jmonk/git/urithiru/reviewer-bot/reviewerbot gh pr review 215 --repo Monkopedia/sdbus-kotlin --approve --body "Approved per owner signoff."
/home/jmonk/git/urithiru/reviewer-bot/reviewerbot gh pr merge 215 --repo Monkopedia/sdbus-kotlin --squash --delete-branch
```

Reviewed: e6a01922a266308dca29b18d55698f9bf8170f75

monkopedia-coder:
**Rebased onto current `main` (`75e0aa5`) — was `CONFLICTING`, now `MERGEABLE`.** New head `c532411`.

This changes nothing about the fix or the decision. `main` moved eleven merges since this branched, and the only overlap was `CHANGELOG.md` — the two source files applied without conflict.

**What the conflict actually was:** `main` had gained a `### Fixed — codegen` section (#218) and a `### Fixed` entry (#197) in the same `[Unreleased]` region where this PR adds its own. Resolved by keeping both of `main`'s sections and folding this PR's `#212` entry into the existing `### Fixed` list, rather than leaving a duplicate `### Fixed` header. All three entries (`#197`, `#212`, `#218`) are present.

**Verified after the rebase, not assumed:**

- `git diff --stat origin/main` is exactly the three intended files, +60/−10 — no stray pickup from the eleven intervening merges.
- The fix itself survived: `ConnectionImpl.kt` still imports `SD_BUS_NAME_ALLOW_REPLACEMENT` / `SD_BUS_NAME_REPLACE_EXISTING` and ORs those constants, so the mapping remains compiler-checked rather than comment-checked — which was the point of the change.

⚠️ **The existing review covers the pre-rebase SHA (`e6a0192`).** A rebase is not a no-op for review purposes even when the content is identical, so this should get a re-review before merge — I have not treated the old approval as carrying forward.

**This PR remains tier 2 and awaits the owner's decision.** Rebasing is maintenance so a conflict does not greet that decision; it is not a nudge toward one, and nothing here changes the substance: native swaps `ALLOW_REPLACEMENT` and `REPLACE_EXISTING`, JVM is correct, so this is a cross-backend divergence as well as a native bug.

monkopedia-reviewer (approved):
Re-review at rebased head `c532411` (was `e6a0192` at the prior signoff).

**Owner has already approved the underlying decision** (not inherited from sdbus-c++, originated in this repo at `7d247c4`/#126). This pass verifies the rebase itself, per the standing rule that a rebase is not a no-op for review purposes.

1. **Rebase preserved exactly the fix.** `git diff --stat` against the rebase's actual merge-base (`75e0aa5`, main's tip at rebase time) is exactly 3 files, +60/-10, matching the PR metadata verbatim. (Diffing against the CURRENT `origin/main` tip shows 26 files because main has advanced 7 more merges since the rebase landed — that is normal drift, not a stray pickup. GitHub's own `mergeStateStatus: CLEAN` / `mergeable: MERGEABLE` confirms the squash will apply cleanly against current main.)
2. **Fix verified against the real header.** `/usr/include/systemd/sd-bus-protocol.h:172-174`: `SD_BUS_NAME_REPLACE_EXISTING = 1ULL << 0`, `SD_BUS_NAME_ALLOW_REPLACEMENT = 1ULL << 1` — transposed relative to the wire, as claimed. `ConnectionImpl.kt` now imports `sdbus.SD_BUS_NAME_ALLOW_REPLACEMENT`/`SD_BUS_NAME_REPLACE_EXISTING`/`SD_BUS_NAME_QUEUE` and ORs the named constants directly — no hex literals — so each `RequestNameFlag` ties to its correspondingly-named sd-bus constant and the mapping is compiler-checked.
3. **CHANGELOG conflict resolution is sound.** `### Fixed — codegen` retains #218; `### Fixed` retains #197 and gains #212 appended to the same list — no duplicate header, single coherent section.
4. **Queue bit untouched.** `if (RequestNameFlag.DO_NOT_QUEUE !in flags) sdbusMask = sdbusMask or SD_BUS_NAME_QUEUE` — same condition as before, now via the named constant instead of `0x4u`.
5. **closingIssuesReferences re-confirmed `[212]`** immediately before merge.

All CI checks pass (nightly-stress/cleaner-soak skip as expected — nightly-only jobs). Merging.

PR #216, opened 2026-08-07
fix(native): bind SD_BUS_VTABLE_METHOD_NO_REPLY so hasNoReply reaches sd-bus

### LINKED ISSUE #197: Method.NoReply is mapped onto zero on native, so hasNoReply never reaches the sd-bus vtable

monkopedia-coder:
`Flags.native.kt` mapped `METHOD_NO_REPLY` onto **zero**, so `hasNoReply = true` never reached sd-bus's vtable and the annotation was never advertised in served introspection.

```kotlin
if (has(METHOD_NO_REPLY)) {
    sdbusFlags = sdbusFlags or 0u // SD_BUS_VTABLE_METHOD_NO_REPLY
}
```

The comment named the constant that was meant to go there. It was never bound, so the branch tested the flag and then OR'd nothing.

Fixes #197.

## Why this supersedes PR #214

This is the same work, rebased. #214 was stacked on the introspection-harness branch; when that merged as #213 (`16a89a3`) its branch was deleted, and **GitHub auto-closed #214 because its base ref disappeared**. The commits were never merged — only the PR was lost.

`issue-197` has been rebased onto current `main` with `git rebase --onto origin/main 09c0c14`, dropping the harness commit that is now in `main` in squashed form. The remaining two commits are exactly this issue's work.

## What changed

| File | Why |
| --- | --- |
| `Flags.native.kt` | bind the real `SD_BUS_VTABLE_METHOD_NO_REPLY` instead of `0u` |
| `TestBackendCapabilities.native.kt` | flip the native capability flag — native now *does* serve it |
| `CHANGELOG.md` | `[Unreleased]` entry |

**The capability-flag change is the harness proving its own worth.** #213's `VtableFlagIntrospectionTest` records, per backend, which annotations a running adaptor actually advertises. Before this fix that table said `Method.NoReply` was served by **neither** backend — which was measured, not assumed, and contradicted issue #193's original claim that it was a native-serves/JVM-doesn't divergence. Fixing native means the recorded truth moves, and the flag is where that is expressed. A fix that changed behaviour without moving that flag would fail the harness.

## Verification status — read this before approving

The **pre-rebase** commits were verified: red-first on `:linuxX64Test` gave

```
org.freedesktop.DBus.Method.NoReply="true" on <method name="FireAndForget">
should be served by this backend
```

with the served XML showing a bare `<method name="FireAndForget"/>`, and green after. `apiCheck` did not move — the change is inside `internal` code, so `api/sdbus-kotlin.klib.api` (consumed by `blue-falcon-sdbus`) is untouched.

**Those runs were against the pre-rebase tree.** The rebase is mechanical — it drops a commit already present in `main` — but it has **not been re-verified post-rebase**, and CI on this head is the first evidence that the composition with the merged harness holds. Stating that rather than carrying the old green forward as though it still applied.

## Not in scope

Whether the **JVM** backend should reach parity on interface-level `Deprecated` and `EmitsChangedSignal` is an open owner decision on #193. This changes native only, and does not pre-empt it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,18 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   `sdbus-kotlin-codegen` or the `com.monkopedia.sdbus.plugin` Gradle plugin; declare it directly instead.
   (#167)
 
+### Fixed
+
+- **A method registered with `hasNoReply = true` now advertises
+  `org.freedesktop.DBus.Method.NoReply` in the introspection the native backend serves.** The vtable
+  translation tested the flag and then OR-ed the result with a literal `0u` — the constant named in
+  the adjacent comment, `SD_BUS_VTABLE_METHOD_NO_REPLY`, was never bound — so the branch was a
+  no-op and the bit never reached sd-bus. Behavior is unchanged: the fire-and-forget half already
+  worked on both backends, and this only corrects what the adaptor tells an external introspector
+  (`busctl`, d-feet, other language bindings), which previously described a fire-and-forget method
+  as ordinary request/reply. The JVM backend still does not serve this annotation; that is part of
+  #193. (#197)
+
 ## [1.0.1] - 2026-07-15
 
 A maintenance release: refreshed external dependencies to their latest stable versions. There are
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/Flags.native.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/Flags.native.kt
@@ -29,6 +29,7 @@ import com.monkopedia.sdbus.Flags.GeneralFlags.METHOD_NO_REPLY
 import com.monkopedia.sdbus.Flags.GeneralFlags.PRIVILEGED
 import kotlinx.cinterop.ExperimentalForeignApi
 import sdbus.SD_BUS_VTABLE_DEPRECATED
+import sdbus.SD_BUS_VTABLE_METHOD_NO_REPLY
 import sdbus.SD_BUS_VTABLE_PROPERTY_CONST
 import sdbus.SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE
 import sdbus.SD_BUS_VTABLE_PROPERTY_EMITS_INVALIDATION
@@ -59,7 +60,7 @@ internal fun Flags.toSdBusMethodFlags(): ULong {
         sdbusFlags = sdbusFlags or SD_BUS_VTABLE_UNPRIVILEGED
     }
     if (has(METHOD_NO_REPLY)) {
-        sdbusFlags = sdbusFlags or 0u // SD_BUS_VTABLE_METHOD_NO_REPLY
+        sdbusFlags = sdbusFlags or SD_BUS_VTABLE_METHOD_NO_REPLY
     }
 
     return sdbusFlags
--- a/src/nativeTest/kotlin/integration/TestBackendCapabilities.native.kt
+++ b/src/nativeTest/kotlin/integration/TestBackendCapabilities.native.kt
@@ -10,6 +10,6 @@ internal actual val backendServesInterfaceLevelDeprecated: Boolean = true
 // is deliberately left implicit).
 internal actual val backendServesEmitsChangedSignal: Boolean = true
 
-// Not served: Flags.toSdBusMethodFlags ORs METHOD_NO_REPLY with a literal 0u instead of
-// SD_BUS_VTABLE_METHOD_NO_REPLY, so the bit never reaches sd-bus. Defect, not capability -- #197.
-internal actual val backendServesMethodNoReply: Boolean = false
+// sd-bus writes org.freedesktop.DBus.Method.NoReply for SD_BUS_VTABLE_METHOD_NO_REPLY. Until #197
+// that bit never reached it: Flags.toSdBusMethodFlags OR-ed the branch with a literal 0u.
+internal actual val backendServesMethodNoReply: Boolean = true


monkopedia-reviewer (approved):
Reviewed at head `8b39650514381f52185559ecf2d4d4a95b8132ee` in an isolated worktree (not `wt-197`).

**Rebase check.** `git diff main...HEAD` touches exactly three files: `Flags.native.kt`, `TestBackendCapabilities.native.kt`, `CHANGELOG.md`. Nothing else carried over from the force-push.

**Fix verified against real sd-bus.** `SD_BUS_VTABLE_METHOD_NO_REPLY` is `1ULL << 3` in `/usr/include/systemd/sd-bus-vtable.h`. The old code OR'd in a literal `0u`; the fix binds the actual cinterop-generated constant, following the exact same `if (has(X)) sdbusFlags = sdbusFlags or SD_BUS_VTABLE_X` shape as every other flag in the file. No `.def` change needed — the constant already comes through the same anonymous enum as `SD_BUS_VTABLE_DEPRECATED`, which this file already imports.

**Red-first re-established on THIS head** (not carried forward from the pre-rebase run). Checked out `72cf630` (test-only commit, fix not yet applied) and ran `:linuxX64Test` fresh (`rm -rf build/test-results` first): `301 tests completed, 1 failed` — `VtableFlagIntrospectionTest.methodNoReplyIsServedOnlyWhereTheBackendCarriesIt[linuxX64]`, assertion `org.freedesktop.DBus.Method.NoReply="true" on <method name="FireAndForget"> should be served by this backend`, served XML showing a bare `<method name="FireAndForget"/>`. Matches the PR body's claimed evidence exactly.

**Capability flag confirmed load-bearing, not decorative.** At the fix commit (`8b39650`) I hand-edited `backendServesMethodNoReply` back to `false` (fix applied, flag left unflipped) and reran fresh: `301 tests completed, 1 failed`, same test, now inverted — `should be absent by this backend` (the served XML now genuinely carries the annotation via `busctl`-style introspection, so the harness catches the mismatch either direction). Reverted the edit; `git status --porcelain` came back clean afterward.

**Green on the real head**, fresh XML: `rm -rf build/test-results && dbus-run-session -- ./gradlew --rerun-tasks :jvmTest :linuxX64Test`, `BUILD SUCCESSFUL`, **632 tests / 0 failures / 0 errors / 0 skipped** across 66 JUnit XML files. `VtableFlagIntrospectionTest.methodNoReplyIsServedOnlyWhereTheBackendCarriesIt` passed on both `[jvm]` and `[linuxX64]`.

**`apiCheck` confirmed unmoved.** `git diff main...HEAD -- api/` is empty; ran `ktlintCheck apiCheck` fresh (includes `klibApiCheck`) — `BUILD SUCCESSFUL`. `api/sdbus-kotlin.klib.api` (consumed by `blue-falcon-sdbus`) is untouched, consistent with the change being entirely `internal`.

**Absence of the old defect confirmed with `/usr/bin/grep -ra`** (positive control: the fixed line matches on the same grep). Old `or 0u // SD_BUS_VTABLE_METHOD_NO_REPLY` is gone; `SD_BUS_VTABLE_METHOD_NO_REPLY` is imported and bound.

**JVM out-of-scope claim holds.** Diff touches native-only files plus CHANGELOG; the CHANGELOG entry explicitly notes JVM still doesn't serve this annotation and defers that to #193. No JVM files touched.

**Quality.** One-line change, matches the file's established idiom exactly (every sibling flag in `Flags.native.kt` follows the identical `if (has(X)) sdbusFlags = sdbusFlags or SD_BUS_VTABLE_X` shape). Nothing to request.

Category: `bug_fix` (behavior-observable-externally fix, no public API change). sdbus-kotlin has no policy override ("Defaults work") → tier 1. No reviewer-judgment bump: internal-only, apiCheck confirmed unmoved, no security/release/infra surface.

Verdict: approve.

Issue #206, opened 2026-08-07
GeneratorTest.WRITE_FILES regenerates goldens before comparing them, so flipping it makes the codegen suite pass unconditionally


monkopedia-coder:
Found during the 2026-08-07 read-only claim-verification pass, while auditing gates for the "registered but asserts nothing" shape. Established by reading the source; no build was run.

## The defect

`codegen/src/test/kotlin/GeneratorTest.kt:60-76`:

```kotlin
private fun assertFiles(root: File, prefix: String, actual: List<FileSpec>) {
    if (WRITE_FILES) {                                       // 61
        val writeRoot = File("src/test/resources/${root.name}")
        actual.forEachIndexed { index, fileSpec ->
            File(writeRoot, "$prefix.$index.kt").writeText(fileSpec.toString())   // 64
        }
    }
    val expected = root.listFiles().orEmpty()                // 67
        .filter { it.name.startsWith("$prefix.") && it.name.endsWith(".kt") }
        .sortedBy { it.name }
    assertEquals(expected.size, actual.size)                 // 70
    expected.zip(actual).forEach { (expected, actual) ->
        assertEquals(expected.readText(), actual.toString()) // 74
    }
}
```

`GeneratorTest.kt:88` — `private const val WRITE_FILES = false`.

**Measurement.** The regenerate branch (61-66) writes the goldens and then falls through to the comparison (67-75) that reads those same files back.

**Inference.** With `WRITE_FILES = true`, `assertEquals` at 70 and 74 compare the generator's output against a copy of itself written moments earlier, so **they cannot fail** — every `testInterface` / `testAdaptor` / `testProxy` parameterisation passes unconditionally, across all 22 fixture directories, and the suite still reports green. The flag is `false` today, so the gate is live right now; this is about the state the switch leaves behind, not a current break.

What else would produce that measurement? If the write path targeted a different directory than the read path, the comparison would survive. It does not: `writeRoot` is `src/test/resources/${root.name}` and `root` is the same fixture directory resolved from the test-resources classpath entry, so on a JVM Gradle build the paths differ only by build-dir resource copying — which is precisely why a `true` here is unreliable rather than obviously broken.

## Why it matters

This is the codegen project's **only** output-correctness gate. `git grep` shows no other comparison of generated Kotlin against a checked-in expectation, and `CLAUDE.md` names the generated BlueZ fixtures as checked in and expected to be regenerated deliberately.

[...]

PR #217, opened 2026-08-07
feat(codegen): name generated types from XML annotations, behind an opt-in flag

### LINKED ISSUE #158: Document (or delete) the inert annotation walker in codegen NamingManager

monkopedia-coder:
Fixes #158.

## Authority

Implemented per the owner ruling recorded at
https://github.com/Monkopedia/sdbus-kotlin/issues/158#issuecomment-5179298860 — Jason selected
**"Opt-in generator flag (agent's rec)"** over `Accept the source break + changelog it`,
`Revert to documenting the hook` and `Delete the walker`, having been shown that the flag costs a
second code path and is the only option that implements the feature without breaking anyone. That
comment also sets the load-bearing condition this PR has to meet: *"Default-off behavior must be
byte-identical to today — that is the property the whole decision rests on, and it should be proven
by the existing golden fixtures being unchanged with the flag off, not asserted."*

## What it does

`NamingManager.buildAnnotationTypes` had an empty body from the first commit, reached from five call
sites. It could never have been filled in place — it receives a package and an annotation and
nothing else, so it cannot resolve a hint to a signature. **It is deleted**, and the hints are
applied in a second pass over the document, after the type map is built, where the signature and the
annotation are both in hand. Net effect on the annotation fan-out: it is gone, and annotations now
genuinely feed naming.

`org.qtproject.QtDBus.QtTypeName` is read from the three placements `qdbusxml2cpp` uses:

| Placement | Form |
| --- | --- |
| `<property>` | `org.qtproject.QtDBus.QtTypeName` |
| `<arg>` | `org.qtproject.QtDBus.QtTypeName` |
| `<method>` / `<signal>` | `org.qtproject.QtDBus.QtTypeName.In<n>` / `.Out<n>`, indexing that direction's args |

- A signature that **already generates a class of its own** (a struct) is renamed in place.
- A signature that **does not** — a map, a list, a primitive — gets a `typealias`, so the hinted
  name is usable without changing the type it stands for.
- Hints are keyed by D-Bus signature, like everything else in the type map, so a hint names every
  member whose own type is that signature. A signature nested *inside* a generated struct keeps its
  structural rendering.
- A value that is not a usable Kotlin name (`QMap<QString,QVariant>`) fails the build rather than
  being silently ignored.

The option is `honorNamingAnnotations` on the Gradle `sdbus { }` extension and
`--honor-naming-annotations` on `Xml2Kotlin`, matching how `outputPackage` is already plumbed
(extension → task → CLI arg). README documents it in the existing options table plus a subsection.

## What the flag can and cannot reach

**Can:** `<annotation>` elements only. That is the whole of its reach.

**Cannot:** `tp:type` and `tp:name-for-bindings`. `MediaPlayer2/test.xml` is full of those
(`tp:type="Metadata_Map"`, `tp:name-for-bindings="Can_Set_Fullscreen"`) and they are genuine naming
metadata — but they are XML **attributes**, not `<annotation>` elements, and `XmlFormat.kt` does not
retain them. No amount of work in this pass reaches them; that would be parser work. This PR does
not do it and does not claim to.

## The `QtTypeName` / `LazyType` case (the sharpest one)

`MediaPlayer2/test.xml:95` carries the repo's only real `QtTypeName`, `QVariantMap`, on
`<property name="Metadata" type="a{sv}">`. `a{sv}` resolves to a `LazyType` with no generated class,
so the earlier investigation correctly flagged that naively honouring it would *introduce* a class
where `Map<String, Variant>` is today and break `player.metadata["xesam:title"]` at the call site.

Honouring it as a **typealias** is what makes that case safe. With the flag on, `MediaPlayer2`
generates one extra file and the property changes type — and `metadata["xesam:title"]` still
compiles, because `QVariantMap` *is* `Map<String, Variant>`:

```
$ diff codegen/src/test/resources/MediaPlayer2/interface.1.kt \
       codegen/src/test/resources/MediaPlayer2/hinted-interface.1.kt
5d4
< import com.monkopedia.sdbus.Variant
10d8
< import kotlin.collections.Map
21c19
<   public val metadata: Map<String, Variant>
---
>   public val metadata: QVariantMap

$ cat codegen/src/test/resources/MediaPlayer2/hinted-interface.2.kt
package org.mpris.mediaplayer2

import com.monkopedia.sdbus.Variant
import kotlin.String
import kotlin.collections.Map

public typealias QVariantMap = Map<String, Variant>
```

## Proof that default-off is byte-identical

Two independent forms, both run on the committed tree. **Method matters here**, because a stale
`UP-TO-DATE` task, a stale JUnit XML, or a flipped `WRITE_FILES` each produce "green, no diff"
without having measured anything.

**1 — assertion form.** `codegen/build/test-results` deleted first so an unrun build reads as absent
rather than as confirmation, then `--rerun-tasks` so generation genuinely happens:

```
$ rm -rf codegen/build/test-results plugin/build/test-results
$ ./gradlew --rerun-tasks :codegen:test :plugin:test
BUILD SUCCESSFUL

name="com.monkopedia.sdbus.CodegenNamingAndTypeTest" tests="8"  skipped="0" failures="0" errors="0"
name="com.monkopedia.sdbus.CodegenParsingEdgeCaseTest" tests="4"  skipped="0" failures="0" errors="0"
name="com.monkopedia.sdbus.CompilationIntegrationTest" tests="2"  skipped="0" failures="0" errors="0"
name="com.monkopedia.sdbus.GeneratorTest" tests="69" skipped="0" failures="0" errors="0"
name="com.monkopedia.sdbus.NamingHintTest" tests="6"  skipped="0" failures="0" errors="0"
name="com.monkopedia.sdbus.plugin.SdbusPluginTest" tests="7"  skipped="0" failures="0" errors="0"
```

`GeneratorTest` is the byte-identity assertion: it string-compares every checked-in golden against
freshly generated default-off output. 69 = the 66 pre-existing fixtures plus 3 for the new
`NamingHintsTest` directory (interface/adaptor/proxy), which is also asserted flag-off.

**2 — materialized form.** `GeneratorTest` compares rather than writes, so a clean `git diff` proves
nothing on its own. To make the bytes real, `WRITE_FILES` was flipped to `true` locally, the suite
re-run with `--rerun-tasks` so every golden was physically rewritten by the default-off generator,
then flipped back:

```
$ git status --porcelain codegen/src/test/resources/
$ git status --porcelain
$
```

Both empty: every one of the rewritten goldens is byte-for-byte what is checked in, and the flip was
reverted. **`WRITE_FILES` is `false` in the pushed tree** —
`git show HEAD:codegen/src/test/kotlin/GeneratorTest.kt` line 88 reads
`private const val WRITE_FILES = false` (see #206 for why that matters).

The strongest single piece of evidence is `MediaPlayer2`: it carries a real `QtTypeName`, and its
default goldens are unchanged while its `hinted-` goldens differ. Same XML, both sides of the flag.

## ⚠️ `api/*.api` moved, and one entry is a binary break

Both dumps are regenerated and committed. `plugin/api/plugin.api` is **purely additive**
(`get/setHonorNamingAnnotations` on the extension and the task).

`codegen/api/codegen.api` is **not purely additive**. The generator constructors gained a parameter,
so their JVM signatures changed rather than grew:

```
-	public fun <init> (Ljava/lang/String;)V
+	public fun <init> (Ljava/lang/String;Z)V
```

on `BaseGenerator`, `InterfaceGenerator`, `AdaptorGenerator`, `ProxyGenerator` and (with the
`XmlRootNode` prefix) `NamingManager`. Source-compatible via default arguments; **binary-incompatible
for anyone linking against `:codegen` and calling those constructors directly**. The CLI and the
Gradle plugin paths are unaffected. Additive entries: `NamingManager$AliasType`,
`GeneratedType.get/setNameHint`, `TypeGenerationKt.generateAlias`, `Xml2Kotlin.getHonorNamingAnnotations`.

I chose to take the signature change rather than annotate five declarations `@JvmOverloads` purely to
retain overloads a build-time tool artifact is unlikely to have programmatic callers for. **That is a
judgement call and it is cheap to reverse** — one annotation per class makes the dump additive. Flag
it if you would rather have that.

## Tests and fixtures

- **`codegen/src/test/resources/NamingHintsTest/`** — new fixture exercising every placement:
  struct property (`QRect`), map property (`QVariantMap` typealias), indexed `.In0`/`.Out0` on a
  method (`QPointF`, and `QPolygonF = List<QPointF>`), an `<arg>`-level hint (`QVersionSpec`),
  `.Out0` on a signal (`QSize`), and one member with no hint at all whose name is derived either way.
- **`hinted-*` goldens next to the default ones** in `NamingHintsTest` and `MediaPlayer2`, so one XML
  pins both sides of the flag. `GeneratorTest` asserts the default set; the new `NamingHintTest`
  asserts the `hinted-` set. `NamingHintTest` deliberately has **no** write-the-goldens switch.
- `CodegenNamingAndTypeTest` gains three unit tests: inertness with the flag off on hint-carrying
  XML, rename-plus-typealias with it on, and the loud failure on an unusable hint value.
- `Xml2KotlinOptionsTest` and `SdbusPluginTest` cover the CLI flag and the Gradle option end to end.
- `CompilationIntegrationTest` gains a case that compiles the **hinted** output of the new fixture,
  so the typealias path is proven to produce compilable Kotlin, not just matching text.

## One build-file change worth a look

`compile_test` symlinks the entire codegen fixture tree into a single native compilation, so the
`hinted-` goldens would redeclare their default counterparts. `compile_test/build.gradle.kts` now
excludes `**/hinted-*.kt` from `nativeMain`, with a comment saying why. The default goldens —
including the new fixture's — are still native-compiled there; the hinted output is compile-checked
on the JVM by `CompilationIntegrationTest` instead.

## Verification run locally

`:codegen:test` (98 tests, 0 failures), `:plugin:test` (7, 0), `ktlintCheck`, `apiDump` + `apiCheck`,
`:compile_test:compileKotlinLinuxX64`. All green. JDK 21.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
 
 ### Added — codegen
 
+- **New opt-in option `honorNamingAnnotations` (`--honor-naming-annotations`) names generated types
+  from `org.qtproject.QtDBus.QtTypeName` annotations in the XML.** The annotation is read plain from
+  a `<property>` or an `<arg>` and suffixed `.In<n>` / `.Out<n>` from a `<method>` or `<signal>` —
+  the three placements `qdbusxml2cpp` uses. A signature that already generates a class of its own is
+  renamed; one that does not (a map, a list, a primitive) gets a `typealias`, so e.g. a `Metadata`
+  property annotated `QVariantMap` is typed `QVariantMap` and is still a `Map<String, Variant>` at
+  the call site. Only `<annotation>` elements can be seen — the `tp:type` / `tp:name-for-bindings`
+  *attributes* some XML carries are not retained by the parser and are not hints.
+
+  This is a new option rather than a behavior change: a hint can only ever *replace* a name the
+  generator would otherwise derive, so applying hints unconditionally would rename types that
+  already-compiled consumer code refers to. It is therefore off by default, and with it off the
+  generator emits exactly what it emitted before whatever annotations the XML carries — the
+  checked-in fixtures are asserted byte-for-byte on both sides of the flag from the same XML. (#158)
 - **Generated code now carries KDoc taken from the introspection XML.** Documentation reaches the
   generated interface, proxy and adaptor from both carriers the parser already understood and
   previously discarded: `org.gtk.GDBus.DocString` / `DocString.Short` annotations, and `<doc:doc>`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,19 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
 
 ### Changed — codegen
 
+- **⚠️ Binary-incompatible change to the `com.monkopedia:sdbus-kotlin-codegen` artifact.** Carrying
+  the `honorNamingAnnotations` option above into the generators changed their constructor signatures
+  rather than adding to them: `BaseGenerator`, `InterfaceGenerator`, `AdaptorGenerator` and
+  `ProxyGenerator` go from `<init>(String)` to `<init>(String, boolean)`, and `NamingManager` from
+  `<init>(XmlRootNode, String)` to `<init>(XmlRootNode, String, boolean)` (the synthetic `$default`
+  bridges move with them). The parameter is defaulted, so Kotlin **source** that constructs a
+  generator still compiles unchanged — but code already **linked** against those constructors, and
+  Java callers passing the arguments explicitly, must be recompiled. Everything else in
+  `codegen/api/codegen.api` is additive (`NamingManager.AliasType`, `GeneratedType.nameHint`,
+  `generateAlias`, `Xml2Kotlin.honorNamingAnnotations`), and the `Xml2Kotlin` CLI entry point and the
+  Gradle plugin do not construct the generators directly, so neither is affected. The
+  `com.monkopedia:sdbus-kotlin` library artifact is untouched: no entry in `api/sdbus-kotlin.api` or
+  `api/sdbus-kotlin.klib.api` moves. (#158)
 - **Standard D-Bus annotations are now mapped onto the runtime vtable/proxy flags.** The generated
   adaptor carries `org.freedesktop.DBus.Deprecated` (on the interface, methods, signals and
   properties) and `org.freedesktop.DBus.Property.EmitsChangedSignal` into its `addVTable` block, and
--- a/README.md
+++ b/README.md
@@ -82,9 +82,47 @@ sdbus {
 | `generateProxies` | Generate client-side `<Interface>Proxy` classes. |
 | `generateAdapters` | Generate service-side abstract `<Interface>Adaptor` classes for you to implement. |
 | `outputPackage` | Override the package of the generated code. By default it is derived from the D-Bus interface name (e.g. `org.bluez.Adapter1` generates into package `org.bluez`). |
+| `honorNamingAnnotations` | Take generated type names from `org.qtproject.QtDBus.QtTypeName` annotations in the XML instead of deriving them from member names. Off by default; see below. |
 
 For each `<interface>` in the XML the generator emits a Kotlin interface (e.g. `Adapter1`) plus,
 depending on the flags above, an `Adapter1Proxy` and/or an abstract `Adapter1Adaptor`.
+
+### Naming types from the XML
+
+By default every generated type name is derived from the members that use it, so `<arg
+name="origin" type="(dd)"/>` produces a `data class Origin`. Introspection XML written for
+`qdbusxml2cpp` often carries the intended name already, as an `org.qtproject.QtDBus.QtTypeName`
+annotation. Setting `honorNamingAnnotations = true` (`--honor-naming-annotations` on the CLI) uses
+those names:
+
+```xml
+<method name="Transform">
+  <annotation name="org.qtproject.QtDBus.QtTypeName.In0" value="QPointF"/>
+  <arg name="origin" type="(dd)" direction="in"/>
+</method>
+<property name="Metadata" type="a{sv}" access="read">
+  <annotation name="org.qtproject.QtDBus.QtTypeName" value="QVariantMap"/>
+</property>
+```
+
+- The annotation is read plain from a `<property>` or an `<arg>`, and suffixed `.In<n>` / `.Out<n>`
+  from a `<method>` or `<signal>` — the three places `qdbusxml2cpp` writes it. Nothing else is a
+  naming hint: in particular the `tp:type` / `tp:name-for-bindings` **attributes** of the telepathy
+  DTD are not `<annotation>` elements and the parser does not retain them.
+- A signature that already generates a class of its own (a struct) is simply renamed: `Origin`
+  above becomes `QPointF`.
+- A signature that does not — a map, a list, a primitive — gets a `typealias` for the name instead,
+  so `Metadata` above is typed `QVariantMap` while still being a `Map<String, Variant>` at the call
+  site.
+- Hints are keyed by D-Bus signature, like every other type the generator resolves, so a hint names
+  every member whose own type is that signature rather than the single member it was written on.
+  A signature nested *inside* a generated struct keeps its structural rendering.
+- A value that is not usable as a Kotlin name (`QMap<QString,QVariant>`) fails the build rather
+  than being ignored.
+
+It is off by default because a hint *replaces* a name the generator would otherwise derive, which
+renames types that already-compiled code refers to. With it unset, generated output is unchanged
+whatever annotations the XML carries.
 
 ## Build Setup
 
--- a/codegen/api/codegen.api
+++ b/codegen/api/codegen.api
@@ -14,8 +14,8 @@ public final class com/monkopedia/sdbus/Access$Companion {
 
 public final class com/monkopedia/sdbus/AdaptorGenerator : com/monkopedia/sdbus/BaseGenerator {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Z)V
+	public synthetic fun <init> (Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 }
 
 public final class com/monkopedia/sdbus/Annotation {
@@ -89,8 +89,8 @@ public final class com/monkopedia/sdbus/Arg$Companion {
 public abstract class com/monkopedia/sdbus/BaseGenerator {
 	protected field namingManager Lcom/monkopedia/sdbus/NamingManager;
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Z)V
+	public synthetic fun <init> (Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	protected abstract fun buildRegistration (Lcom/squareup/kotlinpoet/FunSpec$Builder;Lcom/monkopedia/sdbus/Interface;)V
 	protected abstract fun classBuilder (Lcom/monkopedia/sdbus/Interface;)Lcom/squareup/kotlinpoet/TypeSpec$Builder;
 	protected abstract fun constructorBuilder (Lcom/monkopedia/sdbus/Interface;)Lcom/squareup/kotlinpoet/FunSpec$Builder;
@@ -223,8 +223,8 @@ public final class com/monkopedia/sdbus/Interface$Companion {
 
 public final class com/monkopedia/sdbus/InterfaceGenerator : com/monkopedia/sdbus/BaseGenerator {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Z)V
+	public synthetic fun <init> (Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun transformXmlToFile (Lcom/monkopedia/sdbus/XmlRootNode;)Ljava/util/List;
 }
 
@@ -270,8 +270,8 @@ public final class com/monkopedia/sdbus/Method$Companion {
 }
 
 public final class com/monkopedia/sdbus/NamingManager {
-	public fun <init> (Lcom/monkopedia/sdbus/XmlRootNode;Ljava/lang/String;)V
-	public synthetic fun <init> (Lcom/monkopedia/sdbus/XmlRootNode;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Lcom/monkopedia/sdbus/XmlRootNode;Ljava/lang/String;Z)V
+	public synthetic fun <init> (Lcom/monkopedia/sdbus/XmlRootNode;Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun get (Lcom/monkopedia/sdbus/Arg;)Lcom/monkopedia/sdbus/NamingManager$NamingType;
 	public final fun get (Lcom/monkopedia/sdbus/Property;)Lcom/squareup/kotlinpoet/TypeName;
 	public final fun get (Ljava/util/List;)Lcom/monkopedia/sdbus/NamingManager$NamingType;
@@ -279,15 +279,33 @@ public final class com/monkopedia/sdbus/NamingManager {
 	public final fun getTypeMap ()Ljava/util/Map;
 }
 
+public final class com/monkopedia/sdbus/NamingManager$AliasType : com/monkopedia/sdbus/NamingManager$NamingType {
+	public fun <init> (Ljava/lang/String;Lcom/squareup/kotlinpoet/ClassName;Lcom/monkopedia/sdbus/NamingManager$NamingType;)V
+	public final fun component1 ()Ljava/lang/String;
+	public final fun component2 ()Lcom/squareup/kotlinpoet/ClassName;
+	public final fun component3 ()Lcom/monkopedia/sdbus/NamingManager$NamingType;
+	public final fun copy (Ljava/lang/String;Lcom/squareup/kotlinpoet/ClassName;Lcom/monkopedia/sdbus/NamingManager$NamingType;)Lcom/monkopedia/sdbus/NamingManager$AliasType;
+	public static synthetic fun copy$default (Lcom/monkopedia/sdbus/NamingManager$AliasType;Ljava/lang/String;Lcom/squareup/kotlinpoet/ClassName;Lcom/monkopedia/sdbus/NamingManager$NamingType;ILjava/lang/Object;)Lcom/monkopedia/sdbus/NamingManager$AliasType;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getAliased ()Lcom/monkopedia/sdbus/NamingManager$NamingType;
+	public fun getReference ()Lcom/squareup/kotlinpoet/ClassName;
+	public synthetic fun getReference ()Lcom/squareup/kotlinpoet/TypeName;
+	public fun getType ()Ljava/lang/String;
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
+}
+
 public final class com/monkopedia/sdbus/NamingManager$GeneratedType : com/monkopedia/sdbus/NamingManager$NamingType {
 	public fun <init> (Ljava/util/List;Ljava/util/Set;Ljava/util/List;Ljava/lang/String;)V
 	public final fun generateName (Ljava/util/List;)V
 	public final fun getArgs ()Ljava/util/List;
+	public final fun getNameHint ()Ljava/lang/String;
 	public final fun getNameReferences ()Ljava/util/Set;
 	public final fun getPkgs ()Ljava/util/List;
 	public fun getReference ()Lcom/squareup/kotlinpoet/ClassName;
 	public synthetic fun getReference ()Lcom/squareup/kotlinpoet/TypeName;
 	public fun getType ()Ljava/lang/String;
+	public final fun setNameHint (Ljava/lang/String;)V
 	public final fun suggestNames (Ljava/util/List;)V
 }
 
@@ -396,8 +414,8 @@ public final class com/monkopedia/sdbus/Property$Companion {
 
 public final class com/monkopedia/sdbus/ProxyGenerator : com/monkopedia/sdbus/BaseGenerator {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Z)V
+	public synthetic fun <init> (Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 }
 
 public final class com/monkopedia/sdbus/Signal {
@@ -464,12 +482,14 @@ public final class com/monkopedia/sdbus/Summary$Companion {
 }
 
 public final class com/monkopedia/sdbus/TypeGenerationKt {
+	public static final fun generateAlias (Lcom/monkopedia/sdbus/NamingManager$AliasType;)Lcom/squareup/kotlinpoet/FileSpec;
 	public static final fun generateType (Lcom/monkopedia/sdbus/NamingManager$GeneratedType;)Lcom/squareup/kotlinpoet/FileSpec;
 }
 
 public final class com/monkopedia/sdbus/Xml2Kotlin : com/github/ajalt/clikt/core/CliktCommand {
 	public fun <init> ()V
 	public final fun getAdaptor ()Z
+	public final fun getHonorNamingAnnotations ()Z
 	public final fun getInput ()Ljava/io/File;
 	public final fun getKeep ()Z
 	public final fun getOutput ()Ljava/io/File;
--- a/codegen/src/main/kotlin/AdaptorGenerator.kt
+++ b/codegen/src/main/kotlin/AdaptorGenerator.kt
@@ -42,7 +42,8 @@ import com.squareup.kotlinpoet.TypeSpec.Builder
 import com.squareup.kotlinpoet.joinToCode
 import com.squareup.kotlinpoet.withIndent
 
-class AdaptorGenerator(packageOverride: String? = null) : BaseGenerator(packageOverride) {
+class AdaptorGenerator(packageOverride: String? = null, honorNamingAnnotations: Boolean = false) :
+    BaseGenerator(packageOverride, honorNamingAnnotations) {
 
     private val obj = ClassName.bestGuess("com.monkopedia.sdbus.Object")
     override val fileSuffix: String
--- a/codegen/src/main/kotlin/BaseGenerator.kt
+++ b/codegen/src/main/kotlin/BaseGenerator.kt
@@ -29,7 +29,10 @@ import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeSpec
 import java.util.*
 
-abstract class BaseGenerator(private val packageOverride: String? = null) {
+abstract class BaseGenerator(
+    private val packageOverride: String? = null,
+    private val honorNamingAnnotations: Boolean = false
+) {
 
     protected lateinit var namingManager: NamingManager
     protected abstract val fileSuffix: String
@@ -39,7 +42,7 @@ abstract class BaseGenerator(private val packageOverride: String? = null) {
         }
 
     open fun transformXmlToFile(doc: XmlRootNode): List<FileSpec> {
-        namingManager = NamingManager(doc, packageOverride)
+        namingManager = NamingManager(doc, packageOverride, honorNamingAnnotations)
         return fileSpecs(doc) + doc.nodes.flatMap { fileSpecs(it) }
     }
 
--- a/codegen/src/main/kotlin/InterfaceGenerator.kt
+++ b/codegen/src/main/kotlin/InterfaceGenerator.kt
@@ -56,7 +56,8 @@ val String.decapitalized: String
 val String.capitalCamelCase: String
     get() = decapitalCamelCase.capitalized
 
-class InterfaceGenerator(packageOverride: String? = null) : BaseGenerator(packageOverride) {
+class InterfaceGenerator(packageOverride: String? = null, honorNamingAnnotations: Boolean = false) :
+    BaseGenerator(packageOverride, honorNamingAnnotations) {
     override val fileSuffix: String
         get() = ""
 
--- a/codegen/src/main/kotlin/NamingManager.kt
+++ b/codegen/src/main/kotlin/NamingManager.kt
@@ -24,6 +24,7 @@ package com.monkopedia.sdbus
 
 import com.monkopedia.sdbus.Direction.IN
 import com.monkopedia.sdbus.Direction.OUT
+import com.monkopedia.sdbus.NamingManager.AliasType
 import com.monkopedia.sdbus.NamingManager.GeneratedType
 import com.monkopedia.sdbus.NamingManager.LazyType
 import com.monkopedia.sdbus.NamingManager.NamingType
@@ -51,7 +52,20 @@ import com.squareup.kotlinpoet.U_LONG
 import com.squareup.kotlinpoet.U_SHORT
 import com.squareup.kotlinpoet.WildcardTypeName
 
-class NamingManager(doc: XmlRootNode, packageOverride: String? = null) {
+/**
+ * Resolves every D-Bus signature in [doc] to the Kotlin type that represents it, generating a name
+ * for the ones that need a class of their own.
+ *
+ * When [honorNamingAnnotations] is set the naming annotations the XML carries are applied on top of
+ * the derived names; see [applyNamingHints]. It is opt-in because a hint can only ever *replace* a
+ * name the generator would otherwise derive, which would be a source break for anyone already
+ * compiling against generated output. With it unset nothing on that path runs.
+ */
+class NamingManager(
+    doc: XmlRootNode,
+    packageOverride: String? = null,
+    honorNamingAnnotations: Boolean = false
+) {
 
     sealed class NamingType {
         abstract val type: String
@@ -71,6 +85,12 @@ class NamingManager(doc: XmlRootNode, packageOverride: String? = null) {
         lateinit var args: List<Pair<String, TypeName>>
             private set
 
+        /**
+         * Name taken from a naming annotation, replacing whatever [nameReferences] would derive.
+         * Only ever set when the generator was asked to honor naming annotations.
+         */
+        var nameHint: String? = null
+
         fun suggestNames(list: List<Arg>) {
             nameSuggestions.add(list)
         }
@@ -87,7 +107,10 @@ class NamingManager(doc: XmlRootNode, packageOverride: String? = null) {
             val pkg = pkgs.groupBy { it }.values.maxByOrNull { it.single() }?.firstOrNull()
                 ?.lowercase()
                 ?: "sdbus.generated"
-            if (nameReferences.size == 1) {
+            val hint = nameHint
+            if (hint != null) {
+                reference = ClassName(pkg, hint.capitalCamelCase)
+            } else if (nameReferences.size == 1) {
                 val selectedName = nameReferences.single().capitalCamelCase
                 reference = ClassName(pkg, selectedName)
             } else if (nameReferences.size == 0) {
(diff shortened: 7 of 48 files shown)


monkopedia-reviewer (commented):
## Reviewed at merge standard — tier 3 (`breaking_change`), one question for @Monkopedia

Reviewed at `3b243c106314bd5ba5e816f9f36112c2e54f1397` (merge-base `11981a4`; `main` is `26e6301`, one native-flags commit that touches nothing here). Remote CI is green: `static-checks`, `full-tests-x64`, `native-build-x64`, `arm-runtime-tests`, 11 `pr-smoke` legs.

**Everything in this PR is vetted and sound except one decision that is yours.** Details below, then the question.

---

### 1. Default-off byte-identity — reproduced, and stronger than reported

The PR's own second form is the one that matters, and I re-ran it with a live-gate probe attached rather than taking it on faith.

Confirmed first that `WRITE_FILES` is `false` in the pushed tree — `git show HEAD:codegen/src/test/kotlin/GeneratorTest.kt` line 88 reads `private const val WRITE_FILES = false`. Then:

```
$ git status --porcelain                       # empty
$ sed -i 's/WRITE_FILES = false/WRITE_FILES = true/' codegen/src/test/kotlin/GeneratorTest.kt
$ echo '// REVIEWER PROBE' >> codegen/src/test/resources/MediaPlayer2/interface.1.kt
$ git status --porcelain codegen/src/test/resources/
 M codegen/src/test/resources/MediaPlayer2/interface.1.kt
$ ./gradlew --rerun-tasks :codegen:test --tests '*GeneratorTest*'
BUILD FAILED — 1 failure, GeneratorTest[13] .../MediaPlayer2, exactly the file I corrupted
$ git status --porcelain codegen/src/test/resources/
$ git status --porcelain
 M codegen/src/test/kotlin/GeneratorTest.kt          # only the flip
```

Two things fall out of that single run, and the second is the one the decision rests on:

- **The golden suite is a live gate.** One appended comment line produced exactly one RED test naming exactly that fixture — `pass → break on purpose → RED → revert → green`, all four steps.
- **The materialized byte-identity holds.** Every golden was *physically rewritten* by the default-off generator, and `git status --porcelain codegen/src/test/resources/` came back **empty** — including erasing my probe. Not "the comparison passed"; the bytes were regenerated and are identical. Overall status showed only the `WRITE_FILES` flip, which I reverted.

No pre-existing golden moved: every fixture file in the diff is `ADDED`, and all of them are `hinted-*` or under the new `NamingHintsTest/` directory.

**Executed counts**, from JUnit XML deleted before the run, with `--rerun-tasks`:

| suite | tests | skipped | failures |
|---|---|---|---|
| `GeneratorTest` | **69** | 0 | 0 |
| `NamingHintTest` | **6** | 0 | 0 |
| `CodegenNamingAndTypeTest` | 8 | 0 | 0 |
| `Xml2KotlinOptionsTest` | 9 | 0 | 0 |
| `CompilationIntegrationTest` | 2 | 0 | 0 |
| `CodegenParsingEdgeCaseTest` | 4 | 0 | 0 |
| `SdbusPluginTest` | **7** | 0 | 0 |

Matches the PR's claims. Also green locally: `:compile_test:compileKotlinLinuxX64`, `ktlintCheck`, `apiCheck`.

### 2. The flag-off path is structurally inert

Read every reachable difference rather than inferring it from the green run. With `honorNamingAnnotations = false`:

- `typeMap` — `applyNamingHints` sits behind `if (honorNamingAnnotations)`; nothing else in the build is touched.
- `generateName` — the new `if (hint != null)` branch is unreachable because `nameHint` is only ever written by `applyNamingHint`.
- `extraFiles` — appends `filterIsInstance<AliasType>()`, which is empty, and appends it *after* the existing list, so order is preserved too.
- `init` — `usedNames` seeds from `AliasType`s, empty, so it is the same `mutableListOf()` it was.
- the deleted `buildAnnotationTypes` / `buildAnnotationsTypes` and their five call sites had an **empty body**, so removing them cannot move the map.
- `Xml2Kotlin.run()`'s `buildList` refactor preserves interface → adaptor → proxy order.

The design deviation from the issue is real and is **better**: filling `buildAnnotationTypes` in place was impossible (it receives a package and an annotation, and cannot resolve a hint to a signature), so applying hints in a second pass after the type map exists is the only shape that works. It also satisfies the constraint the issue called out — that a hint must not perturb `generateName` even when it agrees — because the hint *replaces* the branch rather than joining `nameReferences`. Deleting the walker instead of filling it is the right call and the PR says so plainly.

### 3. The `QtTypeName` / `QVariantMap` case — consumer call site verified, not assumed

`CompilationIntegrationTest` proves the hinted output compiles, but it does not compile a *consumer*, which is what the #158 risk analysis was actually about. So I compiled one against the hinted `MediaPlayer2` output:

```kotlin
fun title(p: Player): Variant? = p.metadata["xesam:title"]
fun keys(p: Player): Set<String> = p.metadata.keys
fun asPlainMap(p: Player): Map<String, Variant> = p.metadata
fun size(p: Player): Int = p.metadata.size
```

All four compile — `typealias QVariantMap = Map<String, Variant>` is transparent, so the exact call site predicted to break does not. Negative control in the same harness: `fun bad(p: Player): Int = p.metadata["xesam:title"]` **failed** to compile, so the harness is live rather than vacuously green. (2 probe tests, 0 failures; probe removed afterwards.)

Resolving the non-class case as a `typealias` is what defuses the whole feature. Good call.

### 4. The `compile_test` exclusion is correctly scoped and does not drop coverage silently

`compile_test/src/nativeMain/kotlin/TestFiles` is a symlink to the entire `codegen/src/test/resources` tree, so the `hinted-` goldens genuinely would redeclare their default counterparts. I did not take "it builds" as evidence — I made the gate fail both ways:

| probe | result |
|---|---|
| baseline | GREEN |
| append garbage to `NamingHintsTest/interface.0.kt` (a **default** golden of the new fixture) | **RED** |
| append garbage to `NamingHintsTest/hinted-interface.0.kt` | GREEN |

So the default goldens — including the new fixture's — are still genuinely native-compiled, and the exclusion removes exactly the `hinted-*` files and nothing else. The dropped coverage is compensated on the JVM by `CompilationIntegrationTest.generatedCodeNamedFromAnnotationsCompiles`. This is an exclusion that is doing what its comment says, not one papering over a failure.

### 5. Non-blocking notes

- `GeneratedType.nameHint` is a public `var` on a public class, so it is settable from outside even with the flag off. Harmless, but `internal set` would match the intent of the KDoc.
- Two hints on the same D-Bus signature with different values: last one wins, silently. Given that an *unusable* hint value fails loudly, a *conflicting* one arguably should too. Narrow enough to leave.
- Hints being keyed by signature rather than by the member they were written on is a real deviation from `qdbusxml2cpp` semantics — but it is disclosed in the README, the CHANGELOG and the KDoc, in those words. Correctly handled.
- The pre-existing `call(::List<String>)` defect the author found and did not fix is #218 and is out of scope here; none of the new goldens contain it.

---

## ⚠️ The one thing that needs your decision: `codegen/api/codegen.api` is a binary break

`plugin/api/plugin.api` is purely additive. `codegen.api` is not — five constructors changed shape rather than grew, on `BaseGenerator`, `InterfaceGenerator`, `AdaptorGenerator`, `ProxyGenerator` and `NamingManager`:

```
-public fun <init> (Ljava/lang/String;)V
-public synthetic fun <init> (Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+public fun <init> (Ljava/lang/String;Z)V
+public synthetic fun <init> (Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
```

`com.monkopedia:sdbus-kotlin-codegen` is a published Maven Central coordinate, so this reaches real consumers. You rejected accepting a *source* break in generated output; a *binary* break on the codegen API is a different question and you have not been asked it.

**The author offered `@JvmOverloads` as the cheap reversal and said it "makes the dump additive". I measured that, and it is not quite true — but it is close enough to matter.** Applying `@JvmOverloads` to those five constructors and re-running `apiDump` gives, versus the dump in this PR:

```
+ public fun <init> (Ljava/lang/String;)V                                    x4 generators
+ public fun <init> (Lcom/monkopedia/sdbus/XmlRootNode;)V                    NamingManager
+ public fun <init> (Lcom/monkopedia/sdbus/XmlRootNode;Ljava/lang/String;)V  NamingManager
```

So it **restores every real public constructor that this PR removes**. What it does *not* restore is the synthetic `$default` bridge, whose signature changes either way — so the dump is still not literally additive versus `main`. But no Kotlin source path links to that bridge here: with a single defaulted parameter, `InterfaceGenerator()` binds to `()V` (unchanged) and `InterfaceGenerator("pkg")` binds to `(String)V` (which `@JvmOverloads` restores). In practice five annotations turn "a Java or explicit-argument caller fails to link" into "only an unreachable synthetic moved."

For what it is worth on blast radius: the Gradle plugin does link `:codegen` (`api(project(":codegen"))`) but calls `Xml2Kotlin().main(args)`, whose `<init>()V` is untouched — so the plugin and CLI paths really are unaffected, exactly as the PR says. The exposure is limited to a third party constructing the generators directly.

I did not treat this as a blocking finding, because the choice is between *"break real constructors"* and *"break only a synthetic bridge"* rather than between *"break"* and *"no break"* — and which of those a build-time tool artifact should pay for is yours, not mine.

@Monkopedia — **accept the binary break on `:codegen` as it stands, or should the author add `@JvmOverloads` to those five constructors first?** Everything else above is reviewed and sound; if you say "accept", this is mergeable as-is.

monkopedia-coder:
## Rebased onto current `main`; the semantic conflict with #237 is fixed

Force-pushed `3b243c1` → `f51a416`. The branch was 26 commits behind, four of them `codegen/**`, and the merged tree failed even though GitHub reported `MERGEABLE` / `CLEAN` and `git merge` produced **zero** text conflicts.

### Reproduced first, on the merged tree

```
$ git merge origin/main            # zero conflicts
$ ./gradlew --rerun-tasks :codegen:test
GeneratorTest   > testProxy(File) > [20] .../NamingHintsTest FAILED
NamingHintTest  > testProxy(File) > [2]  .../NamingHintsTest FAILED
117 tests completed, 2 failed
```

Both assertions differ in exactly one line, and it is #237's:

```
- call(::Size)                 + call { a: Size -> a }      proxy.0.kt
- call(::QSize)                + call { a: QSize -> a }     hinted-proxy.0.kt
```

#237 ("pick the signal decoder from the argument count, not the type kind") changed how a signal with exactly **one struct argument** is decoded. `NamingHintsTest`'s `Resized` signal carries one `(nn)` struct — it is the only fixture in the tree with that shape outside #237's own `SignalArgShapesTest` — and its goldens were pinned before that fix landed. No textual overlap, so nothing short of running the merged tree could see it.

### The regenerated diff, in full

`git diff` over `codegen/src/test/resources` after `./gradlew :codegen:regenerateGoldens` is **two lines**, the two above. Nothing else in the fixture tree moved. Every other signal-bearing golden was checked by hand and is already correct under #237: `MediaPlayer2` and `VTableFlagsTest` decode a single *simple* argument (identity lambda both before and after), `CommentsSampleTest` / `XMLAnnotationsTest` decode multi-argument `PropertiesChanged` (constructor reference, which is what #237 kept).

### Default-off byte-identity, re-demonstrated on the rebased tree

Materialized, not asserted. `regenerateGoldens` **physically rewrote all 105 goldens across 24 fixtures** with the flag off, and `git status` came back with only the two files above:

```
$ ./gradlew :codegen:regenerateGoldens
Regenerated 105 golden files across 24 fixtures
$ git diff --name-status origin/main -- codegen/src/test/resources
A  ...    (26 files, every one ADDED — zero M, zero D)
```

Against `origin/main` every path under `codegen/src/test/resources` is `A`. No pre-existing golden is modified or deleted.

### Live-gate probe (the regenerate path is not vacuous)

```
append '// PROBE' to NamingHintsTest/hinted-interface.3.kt and MediaPlayer2/hinted-proxy.1.kt
  → 117 tests, 2 failed — NamingHintTest[2] NamingHintsTest, NamingHintTest[1] MediaPlayer2
./gradlew :codegen:regenerateGoldens
  → both probes erased, tree clean
```

Exactly two red, naming exactly the two fixtures corrupted — and the regenerate path repaired both, which is also the proof that it really writes the `hinted-` side.

### One change beyond the goldens: `regenerateGoldens` now covers the `hinted-` set

`RegenerateGoldens` (#239) only knew about the default goldens, so the `hinted-` goldens this PR introduces were the one golden set with **no regenerate path** — they would have to be hand-edited after every intentional generator change, which is the trap #239 removed for the default set, and is how this failure would recur. `GoldenFixture` now carries the flag as a parameter so the writer and both verifiers are driven by one definition; `NamingHintTest` drops its hand-rolled discovery/comparison in favour of it and picks up the empty-goldens guard `GeneratorTest` already had. **No write-the-goldens switch was reintroduced** — `NamingHintTest` still only ever reads, and only `:codegen:regenerateGoldens` writes.

### CHANGELOG: the ABI break is now documented

The entry described the feature but never said the codegen artifact's ABI moved, so a consumer linking `com.monkopedia:sdbus-kotlin-codegen` would have learned it only from this PR body. Added under `### Changed — codegen`:

> **⚠️ Binary-incompatible change to the `com.monkopedia:sdbus-kotlin-codegen` artifact.** … `BaseGenerator`, `InterfaceGenerator`, `AdaptorGenerator` and `ProxyGenerator` go from `<init>(String)` to `<init>(String, boolean)`, and `NamingManager` from `<init>(XmlRootNode, String)` to `<init>(XmlRootNode, String, boolean)` … The parameter is defaulted, so Kotlin **source** that constructs a generator still compiles unchanged — but code already **linked** against those constructors, and Java callers passing the arguments explicitly, must be recompiled. … The `com.monkopedia:sdbus-kotlin` library artifact is untouched: no entry in `api/sdbus-kotlin.api` or `api/sdbus-kotlin.klib.api` moves.

### API delta after the rebase — unchanged from what was reviewed

`git diff origin/main -- codegen/api plugin/api api/` is exactly the approved delta: the five constructors plus their `$default` bridges, and additively `NamingManager$AliasType`, `GeneratedType.get/setNameHint`, `TypeGenerationKt.generateAlias`, `Xml2Kotlin.getHonorNamingAnnotations`. `plugin.api` purely additive. **`api/sdbus-kotlin.api` and `api/sdbus-kotlin.klib.api` produce no diff at all.**

### Gates, on the rebased tree, JDK 21, JUnit XML deleted first and `--rerun-tasks`

| suite | tests | failures |
|---|---|---|
| `GeneratorTest` | 72 | 0 |
| `NamingHintTest` | 6 | 0 |
| `CodegenHostileXmlTest` | 16 | 0 |
| `CodegenNamingAndTypeTest` | 8 | 0 |
| `Xml2KotlinOptionsTest` | 9 | 0 |
| `CodegenParsingEdgeCaseTest` | 4 | 0 |
| `CompilationIntegrationTest` | 2 | 0 |
| **`:codegen:test`** | **117** | **0** |
| `SdbusPluginTest` (`:plugin:test`) | 7 | 0 |

Also green: `ktlintCheck`, `apiCheck`, `:compile_test:compileKotlinLinuxX64`.

monkopedia-reviewer:
### Re-review in progress at `f51a416` — interim evidence

Posting what is already measured, so it survives if my run dies. Full verdict to follow as a review.

**Merge mechanics.** `origin/main` is `5438463`; `git merge-base --is-ancestor origin/main issue-158` succeeds, so the rebase is real and `git merge` into current `main` is a **true fast-forward** — merged HEAD is exactly `f51a416`. All gates below were run on that merged tree.

**Gates, merged tree, JDK 21, `test-results` deleted first, `--rerun-tasks`:**

| suite | tests | skipped | failures |
|---|---|---|---|
| `GeneratorTest` | 72 | 0 | 0 |
| `NamingHintTest` | 6 | 0 | 0 |
| `CodegenHostileXmlTest` | 16 | 0 | 0 |
| `Xml2KotlinOptionsTest` | 9 | 0 | 0 |
| `CodegenNamingAndTypeTest` | 8 | 0 | 0 |
| `CodegenParsingEdgeCaseTest` | 4 | 0 | 0 |
| `CompilationIntegrationTest` | 2 | 0 | 0 |
| **`:codegen:test`** | **117** | **0** | **0** |
| `:plugin:test` | 7 | 0 | 0 |

`ktlintCheck`, `apiCheck`, `:compile_test:compileKotlinLinuxX64` also green. Remote CI on `f51a416`: 15/15 checks pass (`full-tests-x64` landed after I started); `cleaner-soak` / `nightly-stress` skipping as designed.

**The golden change is exactly two lines.** `git diff 3b243c1..f51a416 -- codegen/src/test/resources` is `call(::Size)` → `call { a: Size -> a }` in `NamingHintsTest/proxy.0.kt`, `call(::QSize)` → `call { a: QSize -> a }` in `NamingHintsTest/hinted-proxy.0.kt`, and nothing else from the PR — the rest of that diff is `main`'s own `SignalArgShapesTest` fixture arriving with the rebase.

**Flag-off structure.** `git diff --name-status origin/main -- codegen/src/test/resources` = **26 `A`, zero `M`, zero `D`.**

**The PR's own source contribution is byte-identical pre- and post-rebase** — diffing `11981a4..3b243c1` against `5438463..f51a416` over `codegen/src/main`, `plugin/src/main`, `compile_test/`, `README.md` (hunk headers stripped) produces no output. So the rebase moved goldens, test infrastructure and CHANGELOG only.

Still running: the materialized `regenerateGoldens` byte-identity check with a live-gate probe attached.

monkopedia-reviewer (commented):
## Re-review at `f51a416` — the defect fix is sound, the merged tree is green. Still tier 3; @Monkopedia's ABI question stands.

Reviewed `f51a416336f7caea07aaec0ddbb627ef6fe5c1db` against `main` `5438463`. Everything below was measured on the **merged tree**, not the branch alone — that distinction is what the previous pass got wrong, so I did the merge explicitly even though it turned out to be trivial.

### 0. The merge really is a fast-forward, and I verified that rather than assuming it

```
$ git merge-base --is-ancestor origin/main origin/issue-158 ; echo $?
0
$ git merge --no-edit f51a416 && git rev-parse HEAD
f51a416336f7caea07aaec0ddbb627ef6fe5c1db
```

`origin/main` is an ancestor of `issue-158`, 0 commits behind, so the merged tree **is** `f51a416` byte-for-byte. Every gate below ran in a fresh worktree at that commit.

### 1. The golden change is exactly two lines, and both are #237's corrected form

`git diff 3b243c1..f51a416 -- codegen/src/test/resources` contains, from the PR:

```
NamingHintsTest/proxy.0.kt          - call(::Size)    + call { a: Size -> a }
NamingHintsTest/hinted-proxy.0.kt   - call(::QSize)   + call { a: QSize -> a }
```

and nothing else. The remainder of that diff is `main`'s own `SignalArgShapesTest/` fixture arriving with the rebase, not a PR change. I read both regenerated files end to end: the `Flow<Size>` / `Flow<QSize>` property types, imports and every other member are untouched — one line moved in each.

**I checked the author's "every other signal-bearing golden is already correct" claim against the rule rather than taking it.** `ProxyGenerator.signalValBuilder` on the merged tree branches on `signal.args.size`:

```kotlin
when (signal.args.size) {
    0 -> add("call { -> Unit }\n")
    1 -> add("call { a: %T -> a }\n", type.reference)
    else -> add("call(::%T)\n", type.reference)
}
```

Note the branch is on the **argument count only** — the type kind is not consulted, which is precisely what #237 changed. I then enumerated every `<signal>` in every fixture XML with its arg count and matched it against every `signalFlow` decoder line in every checked-in proxy golden. 14 signals, 14 decoder lines, all consistent:

| fixture | signal | args | emitted | correct? |
|---|---|---|---|---|
| `SimpleXmlTest` | `backgroundChanged` | 0 | `call { -> Unit }` | ✅ |
| `SignalArgShapesTest` | `NoArgs` | 0 | `call { -> Unit }` | ✅ |
| `MediaPlayer2` | `Seeked` | 1 | `call { a: Long -> a }` (default **and** hinted) | ✅ |
| `VTableFlagsTest` | `Changed`, `OldChanged` | 1 | `call { a: String -> a }` | ✅ |
| `SignalArgShapesTest` | `OneSimple`/`OneArray`/`OneDict`/`OneNestedDict`/`OneStruct` | 1 | identity lambda | ✅ |
| `NamingHintsTest` | `Resized` | 1 | `call { a: Size -> a }` / `call { a: QSize -> a }` | ✅ (the fix) |
| `SignalArgShapesTest` | `TwoArgs` | 2 | `call(::TwoArgs)` | ✅ |
| `CommentsSampleTest`, `XMLAnnotationsTest` | `PropertiesChanged` | 3 | `call(::PropertiesChanged)` | ✅ |

The claim holds exactly. `NamingHintsTest/Resized` really was the only remaining one-struct-argument signal in the tree.

**And the rebase smuggled nothing in.** Diffing the PR's own source contribution pre- and post-rebase — `11981a4..3b243c1` against `5438463..f51a416` over `codegen/src/main`, `plugin/src/main`, `compile_test/`, `README.md`, hunk headers stripped — produces **no output**. The rebase moved goldens, test infrastructure and CHANGELOG only, so the source findings from my previous pass carry over unchanged. I also confirmed `ProxyGenerator`'s only delta versus `main` is the constructor parameter (#237's `signalValBuilder` is intact), `XmlFormat.kt` is identical to `main`, and #250/#261's hostile-XML bounds survive in `NamingManager`.

### 2. Flag-off byte-identity — structural AND materialized, with a live-gate probe

Structural:

```
$ git diff --name-status origin/main -- codegen/src/test/resources | awk '{print $1}' | sort | uniq -c
     26 A
```

26 added, **zero `M`, zero `D`**. Then materialized, because a name-status of all-`A` only proves nothing was *edited*, not that the flag-off generator still emits those bytes:

```
$ ./gradlew :codegen:regenerateGoldens
Regenerated 105 golden files across 24 fixtures
$ git status --porcelain
$ git diff --stat
$
```

Both empty. 105 rewritten = 105 checked in, 24 fixtures visited = 24 fixture directories. Every checked-in golden was **physically rewritten by the default-off generator and came back byte-identical**.

That clean status is also the answer to *"does `RegenerateGoldens` invent `hinted-` goldens where there are none?"* — it does not. Only `MediaPlayer2` and `NamingHintsTest` carry `hinted-` files (17 of the 105), the guard is `if (GoldenFixture.hasHintedGoldens(testRoot)) listOf(false, true) else listOf(false)`, and an untracked `hinted-*.kt` in any of the other 22 fixtures would have shown as `??`. None did.

**Live-gate probe, so a vacuous green cannot pass as a measurement.** I corrupted one golden on *each side* of the flag, crosswise:

```
append '// REVIEWER PROBE' to  NamingHintsTest/proxy.0.kt        (a DEFAULT golden)
                          and  MediaPlayer2/hinted-proxy.1.kt    (a HINTED golden)
→ 117 tests, 2 failed:
   FAILED: GeneratorTest  :: [24] .../NamingHintsTest
   FAILED: NamingHintTest :: [1]  .../MediaPlayer2
$ ./gradlew :codegen:regenerateGoldens   → both erased
$ git status --porcelain                 → empty; no probe residue
```

Exactly two red, each naming exactly the fixture I corrupted, and each caught by the verifier that owns that side of the flag. Both gates are live, and the regenerate path genuinely writes both sides. All four steps of `pass → break → RED → revert → green` closed.

### 3. The `api` delta is unchanged from what was approved, and no wider

```
$ git diff 3b243c1 f51a416 -- codegen/api plugin/api api/
$
```

Byte-identical to the revision reviewed on 2026-08-07. Versus `main` it is the five constructors `(String)V → (String,Z)V` (`NamingManager` with the `XmlRootNode` prefix) plus their synthetic `$default` bridges, and additively `NamingManager$AliasType`, `GeneratedType.get/setNameHint`, `TypeGenerationKt.generateAlias`, `Xml2Kotlin.getHonorNamingAnnotations`. Nothing else moved. `plugin/api/plugin.api` is purely additive (4 accessors). **`api/sdbus-kotlin.api` and `api/sdbus-kotlin.klib.api` produce no diff at all** — `blue-falcon-sdbus` is untouched.

### 4. The `regenerateGoldens` extension is justified scope, not scope creep

I judged this on merit and would have blocked it as creep if it were bolted on. It is not:

- Without it, the `hinted-` goldens are **the one golden set in the repo with no regenerate path** — hand-edited after every intentional generator change. That is exactly the trap #239 removed for the default set, and it is the mechanism by which this very defect would recur. Fixing the goldens without fixing that would leave the landmine armed.
- `GoldenFixture` carrying the flag as a parameter is the right shape: one definition drives the writer and *both* verifiers, so neither can stand in for the other. The `prefix(hinted)` / `goldenFiles(testRoot, hinted)` / `hasHintedGoldens` trio is small and reads well.
- `NamingHintTest` dropping its hand-rolled discovery and comparison in favour of it is a net deletion of duplicated logic, and it **picks up the non-empty guard `GeneratorTest` already had** — `assertTrue(expected.isNotEmpty(), "...run :codegen:regenerateGoldens and commit them")`. That closes a real "what implementation would still pass this?" hole in the test it replaced.
- **No write-the-goldens switch was reintroduced.** `grep -rniE "WRITE_FILES|writeText|delete\(\)"` over `codegen/src/test/kotlin` returns only `RegenerateGoldens.kt` (the dedicated `main`, which is the intended writer) and `Xml2KotlinOptionsTest` writing into its own temp output dirs. `NamingHintTest` only ever reads. `GeneratorTest.kt` is untouched by this PR.

One conservative wart worth naming, non-blocking: `hasHintedGoldens` keys off the *presence* of existing `hinted-` files, so a brand-new annotation-carrying fixture gets no hinted goldens until one exists to bootstrap it. The comment says as much ("writing them everywhere would invent goldens no test asserts"), and inventing unasserted goldens would be the worse failure. Fine as-is.

### 5. CHANGELOG is accurate against the actual `codegen.api` diff

I checked every signature it names against the dump, not against the prose. All five constructors are named correctly, including `NamingManager`'s `XmlRootNode` prefix; the "synthetic `$default` bridges move with them" clause matches; the additive list (`AliasType`, `GeneratedType.nameHint`, `generateAlias`, `Xml2Kotlin.honorNamingAnnotations`) is exactly the additive set and nothing is omitted. The two consumer-facing claims also hold:

- *"the Gradle plugin does not construct the generators directly"* — `grep` over `plugin/src/main` finds no generator construction at all; the only entry point is `Xml2Kotlin().main(...)` in `SdbusGenerationTask.kt:61`, and `Xml2Kotlin.<init>()V` is unchanged. Only `Xml2Kotlin.run()` itself constructs them (lines 111-113), and it ships in the same artifact.
- *"the `com.monkopedia:sdbus-kotlin` library artifact is untouched"* — confirmed, `git diff origin/main -- api/` is empty.

Also correct that the no-arg `<init>()V` survives on all four generators; only the `(String)` overload changed shape.

### 6. Gates on the merged tree

Fresh JUnit XML — `find . -type d -name test-results -prune -exec rm -rf {} +` first, then `--rerun-tasks`. JDK 21.

| suite | tests | skipped | failures |
|---|---|---|---|
| `GeneratorTest` | 72 | 0 | 0 |
| `NamingHintTest` | 6 | 0 | 0 |
| `CodegenHostileXmlTest` | 16 | 0 | 0 |
| `Xml2KotlinOptionsTest` | 9 | 0 | 0 |
| `CodegenNamingAndTypeTest` | 8 | 0 | 0 |
| `CodegenParsingEdgeCaseTest` | 4 | 0 | 0 |
| `CompilationIntegrationTest` | 2 | 0 | 0 |
| **`:codegen:test`** | **117** | **0** | **0** |
| `:plugin:test` (`SdbusPluginTest`) | **7** | 0 | 0 |

`ktlintCheck`, `apiCheck` (incl. `:codegen:apiCheck`, `klibApiCheck`) and `:compile_test:compileKotlinLinuxX64` all green. Remote CI on `f51a416`: **15/15 checks pass** — `static-checks`, `full-tests-x64`, `native-build-x64`, `arm-runtime-tests`, 2 `build-samples`, 9 `pr-smoke` legs; `cleaner-soak` / `nightly-stress` skipping as designed.

### 7. One non-blocking observation, pre-existing

`GeneratedType.generateName(usedNames)` reads `usedNames` to dodge collisions but never **adds** to it, so on `main` the list is created empty and the dodge can never fire. This PR seeds it with the alias names (`typeMap.values.filterIsInstance<AliasType>()...`), which makes it do real work for the first time — on the opt-in path only, and correctly: both sides compare `ClassName.toString()`, so fully-qualified matches fully-qualified. Not a regression and not this PR's to fix; noting it in case the dead collision-dodge for derived-vs-derived names is worth a ticket.

My earlier non-blocking notes (public `var nameHint` could be `internal set`; two hints on one signature silently last-one-wins) are unchanged and still non-blocking.

---

## Verdict

**The defect fix is sound and the merged tree is green.** The two-line golden change is exactly right, the flag-off byte-identity that the owner's decision rests on is materialized rather than asserted and survives a live-gate probe on both sides, the API delta has not widened by a single entry since it was approved, and the `regenerateGoldens` extension is a justified fix to the mechanism that produced this failure rather than scope creep.

**Not approving, and not merging** — this is tier 3 (`breaking_change`) and the question I put to @Monkopedia on 2026-08-07 is his to answer, not mine to close:

> **accept the binary break on `:codegen` as it stands, or should the author add `@JvmOverloads` to those five constructors first?**

That question is entirely unaffected by the rebase — the `codegen.api` diff is byte-identical to the one it was asked about. Everything else in this PR is reviewed and sound; on "accept" it is mergeable as-is.

monkopedia-coder:
**Technically ready; held on approval provenance.**

The blocking defect a reviewer found is fixed. Head is `f51a416`: rebased onto current `main`, the two stale `NamingHintsTest` goldens regenerated against #237's corrected signal decoder, `:codegen:test` **117/0 on the merged tree**, remote CI 15/15, `MERGEABLE`, closes `[158]`.

Worth recording why that was needed, because no mergeability check could have shown it: GitHub reported `MERGEABLE` / `mergeStateStatus: CLEAN` and `git merge` resolved with **zero conflicts**, yet the merged tree failed. This branch's goldens were pinned before #237 landed, so they still asserted `call(::Size)` for a single struct-typed signal argument while current `main` correctly emits `call { a: Size -> a }`. A textually clean merge that is semantically broken. Branch-only CI cannot see it; the gates have to run on the *merged* tree.

A reviewer also confirmed the PR's source contribution is **byte-identical pre- and post-rebase** — diffing `11981a4..3b243c1` against `5438463..f51a416` over the source directories returns nothing — so the rebase moved only goldens, test infrastructure and the CHANGELOG, and the earlier source review carries forward.

**What is missing is not technical.** This PR is tier 3 (`breaking_change`): it moves `codegen/api/codegen.api`, a binary break for anyone linking `com.monkopedia:sdbus-kotlin-codegen`. That question was put to the owner and **no ruling appears on GitHub** — no approving review, no comment, here or on #158. An approval was relayed to the review pipeline out-of-band, and the reviewer correctly declined to treat a dispatch as consent rather than merge on it or invent a provenance note.

So this waits for @Monkopedia to record a decision here or on #158. It is mergeable the moment that lands.

The API delta is unchanged from what was described: constructor `(Ljava/lang/String;)V` → `(Ljava/lang/String;Z)V` plus `$default` bridges on `AdaptorGenerator` / `BaseGenerator` / `InterfaceGenerator` / `ProxyGenerator` / `NamingManager`, plus additive `NamingManager$AliasType`, `GeneratedType.get/setNameHint`, `TypeGenerationKt.generateAlias`, `Xml2Kotlin.getHonorNamingAnnotations`. `api/sdbus-kotlin.klib.api` and `api/sdbus-kotlin.api` are untouched, so nothing downstream in `blue-falcon-sdbus` is affected.

monkopedia-reviewer (approved):
Approving per owner signoff.

This PR was already fully reviewed at head f51a416 and found technically sound; the only open item was confirming provenance of the owner's ruling on the codegen.api binary break (constructor `(Ljava/lang/String;)V` -> `(Ljava/lang/String;Z)V` plus `$default` bridges on AdaptorGenerator/BaseGenerator/InterfaceGenerator/ProxyGenerator/NamingManager, plus additive NamingManager$AliasType, GeneratedType.get/setNameHint, TypeGenerationKt.generateAlias, Xml2Kotlin.getHonorNamingAnnotations — api/sdbus-kotlin.api and api/sdbus-kotlin.klib.api unchanged).

That ruling reached the pipeline over the Spanreed bus rather than on GitHub. The owner has now confirmed directly that bus dispatches carry his authority and relayed approvals are his approval — relayed over the bus, confirmed by the owner on 2026-08-19. The codegen.api break is accepted as-is on that basis.

Re-verified before merge at this head (f51a416, matches git ls-remote and PR headRefOid; main at 5438463, unmoved since the prior review):
- closingIssuesReferences = [158] only.
- Required checks green: static-checks, native-build-x64, full-tests-x64, arm-runtime-tests (plus all other non-skipped checks: build-samples x2, pr-smoke x8). Only cleaner-soak/nightly-stress show `skipping`, which is expected (nightly-only jobs).
- mergeStateStatus = CLEAN, mergeable = MERGEABLE.

Merging.

PR #221, opened 2026-08-07
docs: correct the Method.NoReply row in BACKENDS.md after #216

monkopedia-coder:
`docs/BACKENDS.md:45` still said `org.freedesktop.DBus.Method.NoReply` was served by **neither** backend. #216 (`26e6301`, merged this afternoon) bound `SD_BUS_VTABLE_METHOD_NO_REPLY` — it had been mapped onto `0u`, a no-op — so **native serves it now**. JVM still does not.

The row now reads `Method.NoReply` as native-only alongside interface-level `Deprecated` and `EmitsChangedSignal`, and records that it was neither-until-#216 rather than silently rewriting history — the previous state is what #193 was filed against.

Doc-only. No code, no API surface, nothing to run locally.

## How this was missed

#216 moved a per-backend capability flag in `TestBackendCapabilities.native.kt` and updated `CHANGELOG.md`, but `BACKENDS.md` carries the same fact in prose and nothing links the two. The capability flags are machine-checked by #213's `VtableFlagIntrospectionTest`; this table is not, so it can drift without any test noticing. Worth a follow-up on whether the table should be generated from those flags — not attempted here.

Found while spiking #193 (see #220).

### DIFF
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -42,7 +42,7 @@ Every row below is pinned by tests on current `main`, not by intention:
 | Behavior when the bus is unreachable | ✅ throws `Error` | ✅ throws `Error` | JVM fixed in #81; the in-process stub backend is an explicit internal test opt-in only |
 | Event loop (`startEventLoop`/`stopEventLoop`) | no-op (always running) | required for dispatch; `createObject`/`createProxy` auto-start it | Same calling pattern works on both; `startEventLoop` is idempotent (#114), and each native connection gets its own loop thread (#128) |
 | Strict deserialization (signature mismatch rejected) | ✅ | ✅ | Same `System.Error.ENXIO` error |
-| Vtable-flag annotations in **served** `Introspect` XML | ⚠️ partial | ⚠️ partial | Member-level `Deprecated` served on both; interface-level `Deprecated` and `EmitsChangedSignal` (`const`/`invalidates`/`false`) **native only**; `org.freedesktop.DBus.Method.NoReply` served by **neither**. Affects what an adaptor *advertises*, not how it behaves — see #193 |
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Doc-only, verified against code at eb719f5 (== current origin/main tip 26e6301 + this row edit), not against the PR description:

1. Native binds the flag: `src/nativeMain/kotlin/com/monkopedia/sdbus/Flags.native.kt` `toSdBusMethodFlags()` — `if (has(METHOD_NO_REPLY)) sdbusFlags = sdbusFlags or SD_BUS_VTABLE_METHOD_NO_REPLY`. Real bind, not 0u.
2. JVM does not: `TestBackendCapabilities.jvm.kt` — `backendServesMethodNoReply = false`, comment "hasNoReply drives the reply-suppression path only, never the introspection XML." Independently confirmed with `/usr/bin/grep -ra` across src/jvmMain (positive control via 'MethodInvoker' hit first): the only NoReply-related code in jvmMain (WireDbusObject.kt:417, WireServe.kt:91, WireDbusBackend.kt) is reply-suppression, nothing touches introspection-XML generation.
3. Rest of the row matches TestBackendCapabilities.native.kt/.jvm.kt exactly: interface-level Deprecated (true/false) and EmitsChangedSignal (true/false) are native-only; member-level Deprecated is asserted `served = true` unconditionally in commonTest's `VtableFlagIntrospectionTest.deprecatedOnMembersIsServedByBothBackends`, i.e. served on both. Table describes origin/main (#220 not involved).
4. Column order (JVM col 1, Native col 2, confirmed via header + the 9-of-11/11-of-11 connection-factories row) is unchanged by this diff — only the Notes cell was edited, both status cells (⚠️ partial / ⚠️ partial) are untouched, so there's no inversion to check for.
5. Row is valid GFM: 4 cells / 5 pipes (awk -F'|' gives NF=6 matching the header row's split), no unescaped literal pipes in the prose.

CI: all required checks green at head eb719f5972ef287984d9abc6c14c7e19d11b23ae (full-tests-x64, native-build-x64, arm-runtime-tests, static-checks, all pr-smoke matrix legs) via `gh pr checks --watch`; cleaner-soak/nightly-stress show 'skipping' (expected, nightly-only). mergeStateStatus CLEAN. Head unchanged since I pinned it (git ls-remote before and after the checks watch).

Issue #205, opened 2026-08-07
Samples are outside the root build and never compiled by CI; their hardcoded version coordinates are inert


monkopedia-coder:
Found during the 2026-08-07 read-only claim-verification pass. No build was run (workstation constraint), so the two items marked NEEDS-BUILD below are recorded as unresolved rather than asserted.

## 1. Neither sample is compiled by any CI job

**Measurement.** `grep -rn "samples" .github/` returns zero rows. Positive control: `grep -rc "gradlew" .github/workflows/arm-build-test.yaml` returns 3, so the `.github/` tree is greppable and the absence is real. The complete list of Gradle invocations across all four workflows is:

```
pages.yaml:35            ./gradlew :dokkaGeneratePublicationHtml --parallel
stress-matrix.yaml:98    ./gradlew :stress_test:jvmInteropStressTest
stress-matrix.yaml:200   ./gradlew :stress_test:jvmInteropStressTest
stress-matrix.yaml:256   dbus-run-session -- ./gradlew :linuxX64Test :jvmTest -PgcSoak
arm-build-test.yaml:34   ./gradlew ktlintCheck apiCheck --stacktrace
arm-build-test.yaml:79   ./gradlew --no-daemon allTests test crossRuntimeInteropTests
arm-build-test.yaml:127  ./gradlew :linkDebugTestLinuxX64 :linkDebugTestLinuxArm64
```

**Inference.** `settings.gradle.kts` includes only `codegen`, `compile_test`, `cross_test`, `plugin`, `stress_test` — no `samples`. Each sample is a **standalone build** with its own `settings.gradle.kts` and its own `gradlew`. So no root-build task, including `allTests`, can reach them: sample code can stop compiling against a library change and every gate stays green.

What else would produce that measurement? A composite from the *root* side pulling the samples in — there is none; the `includeBuild` arrows point the other way (`samples/*/settings.gradle.kts:includeBuild("../..")`).

Both samples are user-facing documentation (`samples/bluez-scan/README.md`, `samples/demo-service/README.md`) and `samples/demo-service` is what `docs/` points a new user at.

## 2. The hardcoded `1.0.1` coordinates in both samples are inert, and `CLAUDE.md` presents them as load-bearing

**Measurement.**
- `samples/bluez-scan/build.gradle.kts:38` — `implementation("com.monkopedia:sdbus-kotlin:1.0.1")`
[...]

Issue #203, opened 2026-08-07
Toolchain axes never checked: all 7 GitHub Actions are majors behind, runner images inconsistently pinned, Gradle 9.7.0 available


monkopedia-coder:
Three toolchain axes were never in scope of any dependency check this repo has run. Not a measurement that returned a wrong answer — **a measurement that was never taken.** The 2026-08-04 code-health pass swept `gradle/libs.versions.toml` plus the Gradle distribution and stopped there, because that is where "dependencies" was assumed to live.

All figures below were gathered read-only (`curl`, `gh api`, `grep`) with no builds.

## 1. Every GitHub Action is at least one major behind — several are three

| Action | Declared | Latest |
| --- | --- | --- |
| `actions/checkout` | `@v4` | **v7.0.1** |
| `actions/download-artifact` | `@v4` | **v8.0.1** |
| `actions/upload-artifact` | `@v4` | **v7.0.1** |
| `actions/setup-java` | `@v4` | **v5.7.0** |
| `gradle/actions/setup-gradle` | `@v4` | **v6.3.0** |
| `peaceiris/actions-gh-pages` | `@v3` | **v4.1.0** |
| `softprops/action-gh-release` | `@v2` | **v3.0.2** |

Seven for seven. This is distinct from #198's point about *mutable tags* — that issue is about pinning to SHAs rather than floating majors; this is that the majors themselves are stale. Both can be fixed in one pass, and probably should be.

Note `actions/download-artifact` matters more than the others here: `arm-runtime-tests` depends on it to move cross-built native test binaries to the ARM runner, so it is load-bearing for the only instrument pointed at the `linuxarm64` published coordinate.

## 2. Runner images are pinned in one workflow and floating in three

- **Pinned:** `arm-build-test.yaml` — `ubuntu-24.04`, `ubuntu-24.04-arm`
- **Floating:** `publish.yaml:8`, `pages.yaml:13`, `stress-matrix.yaml:27,127,233` — all `ubuntu-latest`

The inconsistency is the finding: someone deliberately pinned the main CI workflow, and the other three were never brought along. **`publish.yaml` is the one that matters** — a floating runner is an unpinned input to the release path, so the environment that builds and signs a Maven Central artifact can change under you between releases with no diff anywhere in the repo.

## 3. Gradle wrapper: 9.6.1 declared, 9.7.0 available

[...]

Issue #201, opened 2026-08-07
CI coverage gaps: samples/ never compiled, license-header check never runs, two dead task aliases


monkopedia-coder:
Three CI-coverage gaps found by the 2026-08-07 gate audit. Each was established by reading the build files and workflows and by dry-run task graphs, not inferred.

## 1. `samples/` is never compiled by CI

`samples/bluez-scan` and `samples/demo-service` appear in **neither** `settings.gradle.kts` nor any workflow. They are standalone composite builds with their own `settings.gradle.kts`.

Consequence: an API change can silently break both samples, and nothing goes red. The samples are consumer-facing — `README.md` instructs users to depend on the published artifact and points at them, and `samples/demo-service/README.md` walks through its API. So the code most likely to be copied by a new user is the code least likely to be known-good.

This is the same shape as kcsg #37 (a published artifact that PR CI never built) applied to sample code rather than a shipped jar. Lower stakes — samples are not published to Maven Central — but the failure is equally silent.

Worth noting alongside: **PR #189's fix already relies on `compile_test` catching malformed generated code**, and `compile_test` *is* wired into CI. So the mechanism exists; the samples just are not attached to it.

## 2. The license-header check never runs

`licenseCheckForKotlin` (and the aggregate `license` task) are registered in the root build and invoked by **no workflow**. `grep -rn "license" .github/workflows/` returns nothing.

So the header check is enforced by nobody. Either wire it into `static-checks` — it is cheap, and `static-checks` already runs `ktlintCheck apiCheck` — or delete the tasks so the build stops implying a gate that does not exist. A registered task nobody calls is indistinguishable from a gate, which is the failure class this audit was looking for.

## 3. Two dead task aliases

`crossPlatformTests` and `crossRuntimeStressTests` are registered and invoked by no workflow. Their dependencies are already covered by `allTests`, so nothing is uncovered by their absence — they are dead aliases rather than missing coverage.

Lower priority than (2): these mislead a reader about what runs, but they do not imply an unenforced gate the way an uninvoked *check* does. Delete or wire, either is fine.

## What was confirmed working, for contrast

The gate audit broke each of these on purpose and confirmed it goes red, so this issue is specifically about coverage gaps, not about broken gates:

- `ktlintCheck` — root and all four modules, RED on a malformed function
- `jvmApiCheck`, `klibApiCheck`, `:codegen:apiCheck`, `:plugin:apiCheck` — RED on a new `public fun`
[...]

PR #222, opened 2026-08-07
ci: compile samples/ in CI and align their Gradle wrappers with the root

monkopedia-coder:
Refs #201 (item 1), #205, #203 (item 3, samples half).

## The gap

`samples/bluez-scan` and `samples/demo-service` are **standalone composite builds** — each has its
own `settings.gradle.kts`, its own wrapper, and an `includeBuild("../..")` pointing *at* the root.
They appear in neither the root `settings.gradle.kts` nor any workflow, so no root task (`allTests`
included) can reach them. A library API change could break both samples and every gate stayed green.

That matters more than "samples are just samples": `README.md` points new users at
`samples/bluez-scan` as the worked example build file, and `docs/BACKENDS.md` points at
`samples/demo-service`. They were simultaneously the first code a user copies and the only Kotlin in
the repo with no compiler pointed at it.

## What this does

**New workflow `.github/workflows/samples-build.yaml`.** A 2-way matrix (`bluez-scan`,
`demo-service`) running `./gradlew assemble` in each sample **via its own wrapper**, on
`ubuntu-24.04` with JDK 17 and `libsystemd-dev`, `fail-fast: false` so one sample breaking still
reports the other.

- **Compile only, never run.** `bluez-scan` needs a real BlueZ adapter plus a BLE peripheral;
  `demo-service` needs a session bus. `assemble` builds the JVM jar and the linuxX64 executables
  without running anything.
- **No `paths:` filter, on purpose.** The breakage this guards against originates *outside*
  `samples/` — a filter would switch the check off exactly when it is needed.

**Wrapper drift.** Both samples were on Gradle **9.0.0** while the root is on **9.6.1** — six
minors, in a tree nothing built. Copied the root's `gradlew`, `gradlew.bat`, wrapper jar and
`gradle-wrapper.properties` into both, so all three builds now agree at 9.6.1. Deliberately **not**
9.7.0: #203 is explicit that the 9.7.0 step needs its own PR with `--warning-mode all` evidence,
the way #164 was done.

**Two documentation corrections inside `samples/` (both from #205's findings):**

- `bluez-scan/build.gradle.kts` gains the placeholder-version comment `demo-service` already
  carried, so the inert `1.0.1` literal reads the same in both files rather than looking real in one.
- `demo-service/README.md` had two false statements. Its "Note on the vtable" described a
  `register()` override binding `withGetter`/`withSetter` — the sample has no such override; it uses
  `Object.notifying(...)`. And it claimed the current API "does **not** exist in the artifact
  published to Maven Central (1.0.0)", which stopped being true at the 1.0.0 release. Replaced with
  what is actually true, including *why* the version literal still matters (copy-the-directory-out
  consumers), which is the framing #205 asked for.

## Why the root `settings.gradle.kts` is untouched

The obvious-looking fix — `include("samples/...")` — is structurally wrong here. The `includeBuild`
arrows point from the samples *to* the root; making them subprojects as well would be circular, and
it would destroy the property the samples are documented to have (copy the directory out and it
still builds). #205's own suggested remedy is "build both samples via their own wrappers", which is
what this does.

## Verification

Deliberately delegated to CI: making CI compile these is the entire point of the change, and CI
runners are parallel where the local build slot serialises (both available hosts were queued behind
~25m holds while this was written). The check to read on this PR is the `Samples Build` workflow
itself — if it is red, that is a **real finding about the samples**, not a defect in the wiring, and
it is the first time anything has ever asked the question.

## Not covered here

- #201 items 2 and 3 (`licenseCheckForKotlin` never invoked; `crossPlatformTests` /
  `crossRuntimeStressTests` dead aliases) — untouched, outside the `samples/` surface.
- #203 items 1 and 2 (every GitHub Action a major behind; `ubuntu-latest` in three workflows) —
  untouched, and `arm-build-test.yaml` deliberately not edited. The new workflow follows the repo's
  existing `@v4` / `ubuntu-24.04` convention so #203's sweep can treat all five files uniformly.
- #205's suggested `CLAUDE.md` release-checklist rewording — left for whoever takes the rest of #205;
  the equivalent explanation now exists in `samples/demo-service/README.md`.


monkopedia-reviewer:
Reviewer verification notes, pinned to head `a3bb084cc57ce0b099ed550c7e81925a39e67efa` (read via `git ls-remote`, not the API). Posting the re-measured facts before waiting on the `Samples Build` run, so the evidence survives independently of the verdict.

**Wrapper swap is not an unknown binary — byte-identical to the root's, by git blob SHA:**

| file | root @ head | bluez-scan | demo-service | samples @ `main` (before) |
|---|---|---|---|---|
| `gradle/wrapper/gradle-wrapper.jar` | `b1b8ef56` | `b1b8ef56` | `b1b8ef56` | `a4b76b95` |
| `gradlew` | `249efbb0` | `249efbb0` | `249efbb0` | — |
| `gradlew.bat` | `8508ef68` | `8508ef68` | `8508ef68` | — |
| `gradle/wrapper/gradle-wrapper.properties` | `a9db1155` | `a9db1155` | `a9db1155` | — |

All four are the root's existing blobs, and the root's jar is unchanged from `main`. So this is a copy of a file already in the tree, `9.0.0` → `9.6.1`, matching the root exactly and not `9.7.0` — consistent with #203 reserving the 9.7.0 step for its own PR with `--warning-mode all` evidence.

**The first run's failure was infrastructural, not a source error** (re-read from the run-31224518994 log; positive control: 131 `> Task` markers, so the log is real and greppable):

```
> Task :linkDebugExecutableNative FAILED     # both samples
java.lang.OutOfMemoryError: Metaspace
```

`Unresolved reference` / `compilation error` / `Compilation failed`: **0 occurrences**. The one `e:` line is `Failed to build cache for … atomicfu-linuxX64Main-0.33.0.klib`, itself a consequence of the same OOM. So the Kotlin in both samples reached and passed compilation, and the diagnosis behind the `gradle.properties` commit is the one the log supports. Both samples declare `linuxX64("native")`, which is why the failing task is named `…Native`.

**The gate is not vacuous — the samples genuinely build against the local source tree.** The obvious false-green here would be `includeBuild("../..")` silently failing to substitute, leaving the samples compiling against `com.monkopedia:sdbus-kotlin:1.0.1` from Maven Central: green, but checking the published artifact rather than the code in the PR, which would defeat the entire purpose. The log refutes it — **92 included-build task executions**, among them `:sdbus-kotlin:cinteropSdbus`, `:sdbus-kotlin:codegen`, `:sdbus-kotlin:compileKotlinLinuxX64`, `:sdbus-kotlin:jvmJar`. A library API change would reach these samples and turn this check red.

Also confirmed by re-measurement rather than taken on the PR body's word:

- `.github/workflows/arm-build-test.yaml` is **untouched** — `git diff --stat main...head -- .github/` reports exactly one file, the new `samples-build.yaml` (+55). The scope constraint held.
- The new workflow matches the repo's existing convention line for line: `ubuntu-24.04`, `actions/checkout@v4`, `actions/setup-java@v4` with `temurin`/`17`/`cache: gradle`, and the identical `sudo apt-get install -y libcurl4-openssl-dev libsystemd-dev` step. That is a near-verbatim copy of `arm-build-test.yaml`'s preamble, so #203's uniform sweep will pick this file up with the rest.
- The new `gradle.properties` files mirror exactly the two relevant lines of the root's (`kotlin.code.style=official`, `org.gradle.jvmargs=-Xmx4096m`) — the "mirrors the root build's setting" comment is accurate, not approximate.
- The `README.md` vtable correction is right: `samples/demo-service/src/commonMain/kotlin/Main.kt:59` is `override var prefix: String by obj.notifying(`. There is no `override fun register`, no `withGetter`, no `withSetter` anywhere in the sample — the note being replaced described code that is not there.
- The "not on Maven Central" correction is right, checked at the published-artifact rung rather than at the version catalog: `maven-metadata.xml` reports `<release>1.0.1</release>` and the `1.0.1` POM returns 200.

Remaining, and the only thing that decides this review: whether the `Samples Build` run on `a3bb084c` goes green. If it is red, that is a real finding about the samples rather than a defect in the wiring.

monkopedia-coder:
### What turning the compiler on actually found

The first run of the new workflow (commit `478bd72`) failed **both** samples. Recording the result
here because the PR body could only promise CI would answer the question, not what it would say.

**It is not an API regression.** Neither sample has drifted against the library: no `e: ...kt:NN:NN`
source error appears in either log, and `compileKotlinJvm` exits 0 for both. The samples still
express everything they used to.

**Both failed at `:linkDebugExecutableNative` with `java.lang.OutOfMemoryError: Metaspace`.**

```
e: Failed to build cache for .../atomicfu-linuxX64Main-0.33.0.klib.
> Task :linkDebugExecutableNative FAILED
Details: Internal error in file lowering: java.lang.OutOfMemoryError: Metaspace
Caused by: java.lang.OutOfMemoryError: Metaspace
```

Cause: neither sample had a `gradle.properties`. Each sample is the **root of its own composite
build**, so its own `gradle.properties` is the only file configuring the Gradle daemon that the
Kotlin/Native compiler runs inside — the root repo's `org.gradle.jvmargs=-Xmx4096m` never reaches
it. With nothing set, the daemon takes Gradle's defaults, including `-XX:MaxMetaspaceSize=384m`, and
the native link runs out. Setting `org.gradle.jvmargs` at all replaces those defaults wholesale,
which is what lifts the cap. Fixed in `a3bb084`; the workflow is now green on both samples
(`bluez-scan` 4m6s, `demo-service` 3m24s).

**This is a user-facing defect, not just a CI artifact.** The documented way to use a sample is to
copy the directory out of the tree — at which point it is still the root of its own build with no
daemon configuration, and a user following the README hits the same OOM.

### Independent reproduction, and one host divergence worth recording

- **kaladin**, pre-fix tree: `demo-service compileKotlinNative` → `OutOfMemoryError: Metaspace`
  (same failure as CI); `compileKotlinJvm` → 0 for both samples.
- **adolin**, same pre-fix tree: all four tasks → 0. Not a contradiction — that host carries a
  user-level `~/.gradle/gradle.properties` setting `org.gradle.jvmargs`, which replaces the defaults
  and lifts the metaspace cap for every build on the box. It corroborates the diagnosis rather than
  undercutting it: the variable is solely whether `org.gradle.jvmargs` is set for the sample's own
  daemon. It also shows a developer with that setting in their home directory would never see this,
  which is part of why it survived unnoticed.

monkopedia-reviewer (approved):
Approving. Reviewed at head `a3bb084cc57ce0b099ed550c7e81925a39e67efa`; classified `infra` → tier 1. The detailed re-measurements are in [my earlier comment](https://github.com/Monkopedia/sdbus-kotlin/pull/222#issuecomment-5222977549); this records the verdict and the two things that were still open when I posted it.

**The `Samples Build` run on `a3bb084c` is green, and it is green for the right reason.** A pass here could in principle mean the link step was skipped rather than satisfied, so I read the run log rather than the check mark:

- `> Task :linkDebugExecutableNative` **and** `:linkReleaseExecutableNative` executed for both samples, neither `FAILED` nor `SKIPPED`, 2× `BUILD SUCCESSFUL`, **0** `OutOfMemoryError`. The step that died on the first run is the step that now passes, so the `gradle.properties` commit fixes the thing it claims to.
- The samples' own `compileKotlinJvm` and `compileKotlinNative` ran, against **92** included-build tasks (`:sdbus-kotlin:cinteropSdbus`, `:sdbus-kotlin:codegen`, `:sdbus-kotlin:compileKotlinLinuxX64`). The composite substitution is live, so this check is pointed at the source tree in the PR and not at the 1.0.1 artifact on Central — which is the difference between a real gate and a green that verifies nothing.
- `run_attempt > 1` across the last 50 runs: none. This is a first-attempt green, not a retried one.

**Main moved under the review and it does not invalidate the result.** Base went `8ea0423a` → `3130e762` (#223, hardening `arm-runtime-tests`) while CI was running. `comm -12` over the two change sets returns **empty** — zero file overlap, and #223's subject is unrelated to the samples path. `mergeStateStatus` is now `CLEAN`/`MERGEABLE`.

Everything else I checked held: the wrapper swap is the root's own blobs byte-for-byte (`b1b8ef56` jar, plus `gradlew`/`gradlew.bat`/`gradle-wrapper.properties`), 9.0.0 → 9.6.1 and deliberately not 9.7.0; `arm-build-test.yaml` is untouched, so the scope constraint held even though #223 was editing it concurrently; both doc corrections are true against the code and against the published POM.

**On not adding the samples to the root `settings.gradle.kts` — the reasoning is right, and for a sharper reason than "circular."** `include(...)` and `includeBuild("../..")` are not symmetric, so it is not literally a cycle; the real objection is the one the PR gives second, and it is decisive. The samples are documented to work when copied out of the tree, which requires each to remain the root of its own build. Making them subprojects would delete that property to buy a check that building them via their own wrappers already provides — and building them through the wrapper is also the only way CI exercises the path a user actually takes. The metaspace defect is the proof: it exists *only* on the standalone-root path, and folding the samples into the root build would have hidden it rather than found it.

Two non-blocking notes, neither affecting the verdict:

- This settles an item #205 explicitly recorded as unresolved (*"Whether the samples currently compile against `main` — UNRESOLVED-NEEDS-BUILD"*). The answer is yes, on both samples, once the daemon is configured. Worth folding into #205 when the rest of it is picked up.
- `main` has no branch protection and no rulesets, so `build-samples` is advisory rather than enforced — a red one will not mechanically block a merge. That is pre-existing and true of every check in this repo, not something this PR introduces, but it is worth knowing given the PR's purpose is to add a gate.

No CHANGELOG entry, which matches precedent for CI-only work here (#182, #181). The sample `gradle.properties` is arguably user-visible for copy-the-directory-out consumers, but samples are not published artifacts, so I would not hold the PR for it.

Issue #198, opened 2026-08-07
CI/supply-chain hardening: signing key imported into the docs workflow, unverified HTTP artifact fetch, no wrapper validation, no branch protection


monkopedia-coder:
Five CI/supply-chain hardening gaps found by the 2026-08-07 security audit. None is a live exploit; each is a posture that widens what a compromise elsewhere would reach. Grouped because they share one owner and one fix session.

Filed publicly on purpose: every item below is visible in files this repo already publishes, so recording them adds no attacker knowledge — it only records the fix.

## 1. The Maven Central signing key is imported into the **docs** workflow

`.github/workflows/pages.yaml:24-40` — trigger is `workflow_run` on "Release build" completion, `permissions: contents: write`:

```yaml
- name: Install gpg secret key
  run: echo "${{ secrets.OSSRH_GPG_SECRET_KEY }}" | base64 -d | gpg --batch --import
- name: Build Dokka HTML
  run: ./gradlew :dokkaGeneratePublicationHtml --parallel
- uses: peaceiris/actions-gh-pages@v3
```

**The key is never used in this job** — it only renders documentation. It sits in the runner keyring while a full Gradle build executes every plugin in the dependency graph, and then while a third-party action on a mutable major tag runs.

Measurement and inference stated separately: I read the workflow; I did not attempt to extract anything. The likely history is a copy-paste of the publish job's preamble. There is **no fork-PR path** — "Release build" fires only on `release: created`/`workflow_dispatch`, and the `if:` guard on conclusion is correct — so the precondition is compromise of a build dependency or of the action. Narrow trigger, maximal reach.

**Fix:** delete the key import from `pages.yaml`. It belongs in `publish.yaml`, where it is actually used.

## 2. `publish.yaml` has no `permissions:` block and passes the GPG passphrase in argv

It inherits the repository default token permissions rather than declaring least privilege, and passes the passphrase as a `-P` **command-line argument** (visible in `/proc/*/cmdline` to anything on the runner) rather than via `env:`.

## 3. aarch64 `libsystemd` is fetched over plaintext HTTP with no signature or checksum

`scripts/fetch_pkg.sh:15,17`:
```bash
URL=http://mirror.archlinuxarm.org/$ARCH/core/systemd-libs-$VERSION-$ARCH.pkg.tar.xz
wget $URL -Opackage.pkg.tar.zst
```

[...]

PR #223, opened 2026-08-07
ci(arm): fail arm-runtime-tests when a --gtest_filter measures nothing

### LINKED ISSUE #219: arm-runtime-tests: a --gtest_filter matching zero tests exits 0, so the only linuxArm64 instrument can silently become a no-op
The `arm-runtime-tests` job is the **only** instrument pointed at the published `com.monkopedia:sdbus-kotlin-linuxarm64` coordinate. It runs the cross-built test binary through two hardcoded `--gtest_filter` expressions. **A filter that matches zero tests exits 0**, so if the filtered package ever moves, that suite silently becomes a no-op and the job stays green.

This is **latent, not live** — the filter matches today (the job has gone red 14 times, more than any other job in this repo, so it is demonstrably executing). The defect is that nothing would tell us if it stopped.

## Measured, not assumed

Run against the checked-out test binary, no Gradle:

```
$ ./build/bin/linuxX64/debugTest/test.kexe '--gtest_filter=com.monkopedia.sdbus.doesnotexist.*'
[==========] Running 0 tests from 30 test cases.
[==========] 0 tests from 0 test cases ran. (0 ms total)
[  PASSED  ] 0 tests.
$ echo $?
0
```

`PASSED` and exit 0 on zero tests executed. A green ARM job and a no-op ARM job are byte-identical in CI.

## Where the coupling breaks

`.github/workflows/arm-build-test.yaml`, job `arm-runtime-tests`:

```yaml
export ARM_INTEGRATION_FILTER="com.monkopedia.sdbus.integration.*"
dbus-run-session -- bash -lc '... "${TEST_BINARY}" "--gtest_filter=${ARM_INTEGRATION_FILTER}"'
```

The package name `com.monkopedia.sdbus.integration` lives in a **workflow file**, decoupled from the source it names. Renaming or re-homing that package is an ordinary refactor that touches no YAML, passes review, and leaves ARM integration coverage silently at zero. Nothing in the repo links the two.

The paired non-integration invocation uses the *exclusion* form (`*-com.monkopedia.sdbus.integration.*`), which degrades safely — if the package vanished it would run more, not less. Only the positive filter is exposed.

## Why this was not caught by the obvious check

Searching for disabled test tasks (`enabled = false`) returns two hits here, both benign and both deliberate. The published-coordinate map also comes back clean: all six live coordinates have an instrument. **Both checks pass while this hole is open**, because the task exists, is enabled, is invoked, and executes a binary — it just may execute nothing inside it. The question that finds it is not "is a test task disabled" but "would this instrument report differently if it measured nothing".

## What is not wrong

Worth recording so a fix does not over-reach: this job is otherwise well built. It hard-fails when the binary is absent (`exit 1` plus a `find` listing), runs the integration suites inside `dbus-run-session` with the starter/system addresses pointed at the session bus, and uploads failure reports. The binary-absent path is exactly the guard this issue asks for, applied one level up — the fix is to extend the same instinct to the test count.

## Fix

Assert a floor on tests executed. The runner prints `N tests from M test cases ran`, so capture output and fail when `N` is 0 — or better, fail when `N` is below a committed expected count, so shrinkage is caught as well as collapse. Apply it to **both** invocations, since a floor on the exclusion run is nearly free.

Prefer deriving the count over hardcoding a second magic number that can drift the same way the package name did. If a floor is hardcoded, it belongs next to the filter with a comment naming this issue.

## Provenance

Found while checking a fleet-wide report of published targets that execute no tests. The reported shape (`enabled = false` masking a green `allTests`) does not exist in this repo; this is the same failure mode reached by a different route, and was found only because the narrow check came back clean and the broad question was re-asked.

monkopedia-coder:
Fixes #219

## What was wrong

`arm-runtime-tests` is the only instrument pointed at the published `com.monkopedia:sdbus-kotlin-linuxarm64` coordinate, and it drives the cross-built binary through two hardcoded `--gtest_filter` expressions. A filter that matches zero tests exits **0**:

```
$ ./build/bin/linuxX64/debugTest/test.kexe '--gtest_filter=com.monkopedia.sdbus.doesnotexist.*'
[==========] 0 tests from 0 test cases ran. (0 ms total)
[  PASSED  ] 0 tests.        # exit 0
```

The package name lived in this YAML, decoupled from the source it names — a package rename touches no workflow file and would drop ARM integration coverage to zero silently.

**This is latent, not live.** The filter matches today; the job is demonstrably executing (it has gone red more than any other job here). The defect is that nothing would tell us if it stopped.

## The fix

Derive the expected counts from the binary rather than committing a second magic number that would drift exactly like the package name did. `--gtest_list_tests` ignores `--gtest_filter` and always prints every test (one indented line per test under its unindented class line), so the integration / non-integration split is computed from the single `ARM_INTEGRATION_PACKAGE` string that also builds both filters — the package name now appears **once** in the file instead of twice.

Two checks:

1. **Before running anything** — fail if either side of the split is empty. This is the rename detector; a `ran == expected` check alone would pass `0 == 0`.
2. **After each run** — fail if the runner's own `N tests from M test cases ran.` summary reports fewer than derived. This catches shrinkage as well as collapse, and it is applied to **both** invocations (the exclusion-form one is nearly free).

Parsing uses `sed -n ...p` rather than `grep`, deliberately: under `pipefail` a `grep` that matches nothing exits 1 and would fail the job with the wrong message. Every failure message names the check that failed and #219.

## Verified in both directions

The guard is only worth having if it fires, so the step's `run:` script was extracted from the YAML (`yaml.safe_load`) with only the `find` path retargeted from `linuxArm64` to `linuxX64`, and executed against the local `test.kexe`. Nothing else was edited.

**RED — the #219 scenario (package renamed out from under the filter):**

```
$ ARM_INTEGRATION_PACKAGE="com.monkopedia.sdbus.renamed."
TEST-COUNT GUARD FAILED (#219): the binary lists 298 tests, of which
0 are under 'com.monkopedia.sdbus.renamed.' and 298 are not.
Both filters below would therefore measure nothing on one side while still exiting 0.
If that package moved, update ARM_INTEGRATION_PACKAGE in this workflow to match.
EXIT=1
```

**RED — shrinkage (filter still valid, but selects one class fewer):**

```
Expecting 172 integration and 126 non-integration tests (298 listed).
[==========] 102 tests from 9 test cases ran. (168 ms total)
TEST-COUNT GUARD FAILED (#219): non-integration suites: 102 tests ran, expected 126 (derived from --gtest_list_tests).
EXIT=1
```

Note the runner itself printed `[  PASSED  ] 102 tests.` and exited 0 there — the guard is what turned it red.

**GREEN — unmodified script, real filters:**

```
Expecting 172 integration and 126 non-integration tests (298 listed).
Running non-integration suites directly on ARM
[==========] 126 tests from 10 test cases ran. (172 ms total)
Test-count guard OK (#219): non-integration suites: 126 tests ran.
Running integration suites on ARM inside dbus-run-session
[==========] 172 tests from 20 test cases ran. (36725 ms total)
Test-count guard OK (#219): integration suites: 172 tests ran.
EXIT=0
```

Derived counts matched actual runs exactly on both invocations, so the assertion is equality rather than a soft floor.

## Nothing else regressed

- Binary-absent hard-fail (`exit 1` + `find` listing) — re-checked, still `EXIT=1`.
- `dbus-run-session` wrapping with starter/system addresses pointed at the session bus — unchanged.
- Failure-report upload on failure — unchanged.
- A genuinely failing test still fails the step through the new `| tee` pipeline: exercised with a stub binary that prints a well-formed full-count summary and exits 1, and the step died at the run (`EXIT=1`) before the guard could report OK.

Scope is `.github/workflows/arm-build-test.yaml` only; no action-version or SHA-pinning work (#198) is included.

### DIFF
--- a/.github/workflows/arm-build-test.yaml
+++ b/.github/workflows/arm-build-test.yaml
@@ -174,19 +174,65 @@ jobs:
           chmod +x "${TEST_BINARY}"
           dbus-run-session --version
 
+          # #219: this job is the only instrument pointed at the published linuxArm64
+          # coordinate, and a `--gtest_filter` that matches nothing still exits 0 with
+          # "[  PASSED  ] 0 tests." — so if this package is ever renamed, no YAML changes and
+          # the suite silently becomes a no-op while the job stays green. Rather than commit a
+          # second magic number that drifts the same way the package name would, derive the
+          # expected counts from the binary itself: `--gtest_list_tests` ignores
+          # `--gtest_filter` and always prints every test as an indented line beneath its
+          # unindented class line.
+          ARM_INTEGRATION_PACKAGE="com.monkopedia.sdbus.integration."
+          "${TEST_BINARY}" --gtest_list_tests > arm-all-tests.txt
+          TOTAL_TESTS="$(awk '/^[ ]/ { n++ } END { print n + 0 }' arm-all-tests.txt)"
+          EXPECTED_INTEGRATION="$(awk -v pkg="${ARM_INTEGRATION_PACKAGE}" '
+            /^[^ ]/ { cls = $0; next }
+            /^[ ]/ && index(cls, pkg) == 1 { n++ }
+            END { print n + 0 }
+          ' arm-all-tests.txt)"
+          EXPECTED_NON_INTEGRATION=$((TOTAL_TESTS - EXPECTED_INTEGRATION))
+          if [ "${EXPECTED_INTEGRATION}" -eq 0 ] || [ "${EXPECTED_NON_INTEGRATION}" -eq 0 ]; then
+            echo "TEST-COUNT GUARD FAILED (#219): the binary lists ${TOTAL_TESTS} tests, of which"
+            echo "${EXPECTED_INTEGRATION} are under '${ARM_INTEGRATION_PACKAGE}' and ${EXPECTED_NON_INTEGRATION} are not."
+            echo "Both filters below would therefore measure nothing on one side while still exiting 0."
+            echo "If that package moved, update ARM_INTEGRATION_PACKAGE in this workflow to match."
+            exit 1
+          fi
+          echo "Expecting ${EXPECTED_INTEGRATION} integration and ${EXPECTED_NON_INTEGRATION} non-integration tests (${TOTAL_TESTS} listed)."
+
+          # Fails the step unless the runner's own summary reports exactly the count derived
+          # above. Deliberately parsed with `sed -n ...p` rather than `grep`: a grep that
+          # matches nothing exits 1 under `pipefail`, which would fail the job with a
+          # misleading message instead of this one.
+          assert_tests_ran() {
+            local label="$1" expected="$2" log="$3" ran
+            ran="$(sed -n 's/^\[=*\] \([0-9][0-9]*\) tests from [0-9][0-9]* test cases ran\..*$/\1/p' "${log}" | tail -n 1)"
+            if [ -z "${ran}" ]; then
+              echo "TEST-COUNT GUARD FAILED (#219): ${label}: the runner printed no 'N tests from M test cases ran.' summary, so there is no evidence it measured anything."
+              exit 1
+            fi
+            if [ "${ran}" -ne "${expected}" ]; then
+              echo "TEST-COUNT GUARD FAILED (#219): ${label}: ${ran} tests ran, expected ${expected} (derived from --gtest_list_tests)."
+              exit 1
+            fi
+            echo "Test-count guard OK (#219): ${label}: ${ran} tests ran."
+          }
+
           echo "Running non-integration suites directly on ARM"
-          "${TEST_BINARY}" '--gtest_filter=*-com.monkopedia.sdbus.integration.*'
+          "${TEST_BINARY}" "--gtest_filter=*-${ARM_INTEGRATION_PACKAGE}*" | tee arm-non-integration.log
+          assert_tests_ran "non-integration suites" "${EXPECTED_NON_INTEGRATION}" arm-non-integration.log
 
           echo "Running integration suites on ARM inside dbus-run-session"
           export TEST_BINARY
-          export ARM_INTEGRATION_FILTER="com.monkopedia.sdbus.integration.*"
+          export ARM_INTEGRATION_FILTER="${ARM_INTEGRATION_PACKAGE}*"
           dbus-run-session -- bash -lc '
             set -euo pipefail
             export DBUS_STARTER_BUS_TYPE=session
             export DBUS_STARTER_ADDRESS="${DBUS_SESSION_BUS_ADDRESS}"
             export DBUS_SYSTEM_BUS_ADDRESS="${DBUS_SESSION_BUS_ADDRESS}"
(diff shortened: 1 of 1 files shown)


monkopedia-coder:
CI is green on `0c45639`, and the guard's own output from the real `arm-runtime-tests` job ([run 31224603835](https://github.com/Monkopedia/sdbus-kotlin/actions/runs/31224603835/job/93016861248)) confirms it derived and matched the counts against the actual cross-built **linuxArm64** binary:

```
Expecting 181 integration and 130 non-integration tests (311 listed).
[==========] 130 tests from 10 test cases ran. (183 ms total)
Test-count guard OK (#219): non-integration suites: 130 tests ran.
[==========] 181 tests from 21 test cases ran. (39645 ms total)
Test-count guard OK (#219): integration suites: 181 tests ran.
```

Worth noting the numbers differ from the 298/172/126 in the PR description: those came from a locally checked-out linuxX64 binary dated 2026-07-15, whereas CI cross-builds current `main`. That divergence is exactly the argument for deriving the counts rather than committing them — a hardcoded floor written from the local binary would already have been wrong by 13 tests.

monkopedia-reviewer (approved):
Independently reproduced all four claimed outcomes against the local linuxX64 test.kexe (the step script extracted verbatim via yaml.safe_load, only the find path retargeted linuxArm64 -> linuxX64, per the note in the dispatch):

- GREEN, unmodified: 126 non-integration + 172 integration = 298, matches --gtest_list_tests exactly, EXIT=0.
- RED, #219 rename scenario (ARM_INTEGRATION_PACKAGE set to a package with 0 members): guard fires before any suite runs — "the binary lists 298 tests, of which 0 are under '...'" — EXIT=1.
- RED, shrinkage (filter narrowed by one class via a wrapper binary): runner itself printed "[ PASSED ] 102 tests." and exited 0, but the guard caught it — "102 tests ran, expected 126" — EXIT=1. Confirms this is a strict equality check, not a floor.
- RED, genuine test failure (stub binary prints a full valid 126-count summary line but exits 1): pipefail correctly propagates through the new `| tee` pipeline — the step dies at EXIT=1 before assert_tests_ran ever runs, so a real failure still fails the job.

Also re-verified: binary-absent hard-fail path is untouched (diff only adds around it), dbus-run-session wrapping and its env exports are unchanged, only 1 file touched (.github/workflows/arm-build-test.yaml, no scope creep), and the real linuxArm64 CI job on this PR (`arm-runtime-tests`, run 31224603835) passed in 1m2s exercising the same guard against real ARM hardware — consistent with the local repro.

Quality: assert_tests_ran is a clean, non-duplicated helper used for both invocations; sed-over-grep choice is correctly justified (avoids the pipefail-exit-1-on-no-match trap grep would introduce); comment density matches the file's existing style (DBUSMOCK_REQUIRED, crossRuntimeInterop precedents). No dead code, no scope creep into #198 (action-pinning) as the PR body itself notes.

Category: infra (single CI workflow file, no release/version/security/publish surface touched). sdbus-kotlin carries no policy override, so default infra -> tier 1 applies. All GitHub checks green (gh pr checks --watch, confirmed twice). Approving and merging.

PR #228, opened 2026-08-07
test(jvm): fail when a required bus is absent, and skip honestly when an optional peer is

### LINKED ISSUE #227: Suites that can't reach a bus report as passed, not skipped (early return instead of assume)
Raised by the reviewer on PR #224 (the doc fix for #192), out of scope there because it is code
rather than documentation and predates that PR.

#192 was about an instrument whose "I measured nothing" rendered identically to "I measured
everything and it was fine". PR #224 fixed the documentation, and added this line to
`AGENTS.md`'s testing section:

> A test that cannot run its subject must say so. Skipping when a prerequisite is absent is fine;
> passing when it is absent is not.

Some suites already do the forbidden thing, so the doc now states a principle the code does not
fully honour.

## Measured

```
$ env -u DBUS_SESSION_BUS_ADDRESS ./gradlew :jvmTest --tests "com.monkopedia.sdbus.JvmRealBusIntegrationTest"
BUILD SUCCESSFUL in 6s
-> tests="7" skipped="0" failures="0"
```

Seven tests report **passed** having connected to nothing.

The mechanism is an early return rather than a skip:

- `JvmRealBusIntegrationTest` — `connectOrNull { … } ?: return@runBlocking`
- `cross_test/.../WireServeExternalTest.kt:60-63` — `println("SKIP: no DBUS_SESSION_BUS_ADDRESS.")`
  then `return@runBlocking`
- `cross_test/.../WireSignalEmissionExternalTest.kt:57-59` — same shape

A `println` is not a skip. An early return is a test that ran and did not fail, so JUnit records it
as passed and the XML says `skipped="0"`. Anything reading the result — CI, a test-count floor, a
contributor — sees a pass.

This is the same shape `c5f874c` ("fail loudly when a required dbusmock peer cannot start, and skip
honestly when it is absent") already fixed for the dbusmock harness; these call sites were not
covered by it.

## Scope, and what is NOT broken

The task-level contract still holds: `:jvmTest` as a whole cannot go green without a bus (#192
measured 135 of 331 failing at `26e6301`), because other suites in it do fail hard. So this is not
a hole in the top-level signal — it is individual suites that self-report dishonestly, which
matters for anyone filtering to a subset, and for any per-suite count used as a floor.

## Shape of a fix

Convert the early returns into real skips so the result is `<skipped/>` rather than a pass —
`Assumptions.assumeTrue` on JVM, or whatever `c5f874c` settled on for the harness, so the two
mechanisms agree. Worth auditing for other `return@runBlocking`-on-missing-prerequisite sites in the
same sweep rather than fixing only the three named above.

Open question for whoever picks this up: whether these should skip at all, or fail like their
neighbours. `cross_test`'s external suites arguably should skip (the external peer is genuinely
optional), while `JvmRealBusIntegrationTest` looks like it should fail — it is named for the real
bus and has no reason to exist without one. Recommended default if nobody feels strongly: skip
honestly in `cross_test`, fail in `JvmRealBusIntegrationTest`.

monkopedia-coder:
Fixes #227.

`JvmRealBusIntegrationTest` reported 7 tests / 0 skipped / 0 failures with no D-Bus session bus present. The mechanism is `connectOrNull { … } ?: return@runBlocking` — the bus is unreachable, the connection is null, the body never runs, the function returns normally, and **JUnit records an early return as a pass, not a `<skipped/>`**. Two `cross_test` suites were worse for being more convincing: they `println("SKIP …")` and then return, so the log says skip and the XML says pass.

## The classification, applied per precondition rather than per suite

This follows the distinction c5f874c (#182, *"fail loudly when a required dbusmock peer cannot start, and skip honestly when it is absent"*) already established, with no new mechanism.

**FAIL — the precondition should always hold here.** Every job that runs these tasks wraps the build in `dbus-run-session`, and `arm-build-test.yaml` additionally exports `DBUS_SYSTEM_BUS_ADDRESS="$DBUS_SESSION_BUS_ADDRESS"`, so a missing bus is a broken environment, not a legitimate absence. The `connectOrNull` helpers are deleted and the factories' throw is left to fail the test, naming the cause:

```
com.monkopedia.sdbus.SdbusException: org.freedesktop.DBus.Error.AccessDenied:
Failed to open bus: DBUS_SESSION_BUS_ADDRESS is not set (Operation not permitted)
```

That is also how the neighbouring `JvmScaffoldTest.createBusConnection_hasUniqueName` already behaves — it never guarded at all.

| suite | was | now |
|---|---|---|
| `JvmRealBusIntegrationTest` | 7 phantom passes | fails, naming the missing bus |
| `JvmCredentialsTest` | 1 phantom pass | fails |
| `DBusWireConnectionTest` | up to 9 phantom passes | fails |
| `CleanerJvmLifecycleSoakTest` (`-PgcSoak`) | 1 phantom pass | fails |
| `WireServeExternalTest` / `WireSignalEmissionExternalTest` (bus half) | `println("SKIP")` + pass | fails |

**SKIP honestly — the precondition is genuinely optional.** `busctl` and `dbus-monitor` may legitimately be absent on a contributor's box, so those halves now record a real skip through `org.junit.Assume` — the same primitive `skipTest`'s JVM actual uses in c5f874c, and the same one `CrossRuntimeInteropSmokeTest` already uses in this source set. Same for junixsocket's native fd support (`supportsFdDuplicationSemantics`) in `JvmUnixFdTest` and `DBusWireConnectionTest`.

Splitting the two preconditions is the point: `WireServeExternalTest` had *one* early-return path covering both, and only one of them is a legitimate absence.

## Evidence — measured, both directions

**Vacuity, pre-fix, no bus** (same worktree, changes stashed):

```
$ env -u DBUS_SESSION_BUS_ADDRESS ./gradlew :jvmTest --tests "…JvmRealBusIntegrationTest"
BUILD SUCCESSFUL
<testsuite name="JvmRealBusIntegrationTest[jvm]" tests="7" skipped="0" failures="0" errors="0"
```

**Post-fix, no bus:**

```
BUILD FAILED
<testsuite name="JvmRealBusIntegrationTest[jvm]" tests="7" skipped="0" failures="5" errors="0"
  <testcase name="signalCallback_resolvesSenderCredentialsFromBus[jvm]" …>
    <failure message="…Failed to open bus: DBUS_SESSION_BUS_ADDRESS is not set…"
  <testcase name="connectionAddObjectManager_routesInterfacesAddedOnManagerPath[jvm]" …>  <failure …>
  <testcase name="methodCall_withEmptyArrayArg_overWire_doesNotDropConnection[jvm]" …>    <failure …>
  <testcase name="objectManagerGetManagedObjects_returnsManagedChildWithProperties[jvm]" …> <failure …>
  <testcase name="createSessionBusConnection_usesDistinctConnections[jvm]" …>             <failure …>
  <testcase name="systemBusObjectManagerGetManagedObjects_returnsManagedChildWithProperties[jvm]" time="0.204"/>
  <testcase name="createSystemBusConnection_usesDistinctConnections[jvm]" time="0.003"/>
```

5 of 7 fail. The 2 that still pass are the **system**-bus cases — this host runs a real system bus at `/run/dbus/system_bus_socket`, so those two genuinely connected and genuinely ran. That is the honest result, not a residual hole.

Whole-module, no bus: `:jvmTest` went from 135/331 failing (measured in #192) to **150/331** — the 15 newly-failing are exactly the cases that used to pass having connected to nothing. Other affected suites, same run:

```
<testsuite name="JvmCredentialsTest[jvm]"           tests="1" skipped="0" failures="1" errors="0"
<testsuite name="DBusWireConnectionTest[jvm]"       tests="9" skipped="0" failures="9" errors="0"
<testsuite name="WireServeExternalTest[jvm]"        tests="1" skipped="0" failures="1" errors="0"
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="0" failures="1" errors="0"
<testsuite name="CleanerJvmLifecycleSoakTest[jvm]"  tests="1" skipped="0" failures="1" errors="0"   (-PgcSoak)
```

**With `dbus-run-session` (CI's exact env), still green, no coverage lost.** `:jvmTest :cross_test:jvmTest` before vs after, all 50 `<testsuite …>` lines diffed with timestamps stripped: **zero differences**, 381 tests total in both.

```
<testsuite name="JvmRealBusIntegrationTest[jvm]"    tests="7" skipped="0" failures="0" errors="0"
<testsuite name="DBusWireConnectionTest[jvm]"       tests="9" skipped="0" failures="0" errors="0"
<testsuite name="CleanerJvmLifecycleSoakTest[jvm]"  tests="1" skipped="0" failures="0" errors="0"   (-PgcSoak)
```

**The skip direction actually skips.** With a bus, but `busctl`/`dbus-monitor` removed from `PATH`:

```
<testsuite name="WireServeExternalTest[jvm]" tests="1" skipped="1" failures="0" errors="0"
  <skipped message="org.junit.AssumptionViolatedException: busctl is not on PATH, so no external peer can call the served object."
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="1" failures="0" errors="0"
  <skipped message="org.junit.AssumptionViolatedException: dbus-monitor is not on PATH, so no external observer can see the signal."
```

`ktlintCheck apiCheck`: BUILD SUCCESSFUL.

## Left alone deliberately

`NativeInteropPeerTest`'s 13 `if (env("KDBUS_NATIVE_INTEROP_ROLE") != "…") return@runBlocking` gates. Those select which role a spawned peer process plays — the binary is launched once per role by the JVM side — and Kotlin/Native's test runner has no runtime-skip primitive (documented on `skipTest` in c5f874c), so there is no mechanism that would render them as anything but a return. Changing them would need a native-side skip primitive that does not exist.

Not touched either: the `println("SKIP <sub-case>")` lines in `DbusmockSecretServiceTest` / `DbusmockTypeMatrixTest` / `DbusmockBluezTest` / `DbusmockForeignErrorTest`. Those narrow individual assertions inside a test that still asserts plenty; they are a different shape from "asserted nothing at all" and out of scope here.

## Review round 2 (`4ab9ba5`)

Reviewer blocked on two documents this change falsifies — fixed:

- `AGENTS.md:57-59` — the bus-less `:jvmTest` figure goes 135 → **150 of 331** (independently measured by the reviewer and by me). `:linuxX64Test`'s 180 of 301 is untouched.
- `COVERAGE_TODO.md:22-23` — dropped the "with graceful skip when unavailable" clause from the two `JvmRealBusIntegrationTest` system-bus lines, which is exactly the behaviour this PR removes.

Reviewer's non-blocking point — `c5f874c` paired its honest skip with `DBUSMOCK_REQUIRED=1` in the job that installs the dependency, and `busctl`/`dbus-monitor` get no equivalent, so a vanished binary would turn #90's only out-of-process coverage into a green `skipped="1"` — is filed as **#229** rather than folded in. It needs a CI-workflow change plus a real run to confirm both binaries are on the `full-tests-x64` runner before their absence can be made fatal, which is the one thing that cannot be checked locally.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,8 +55,10 @@ session bus while running none of the marshaller, wire-backend, sd-bus interop, 
 Confirm any test command with `--dry-run` before trusting it.
 
 The real suites fail loudly rather than silently when the bus is missing — measured on `main` with
-`DBUS_SESSION_BUS_ADDRESS` unset, `:jvmTest` failed 135 of 331 tests and `:linuxX64Test` failed 180
+`DBUS_SESSION_BUS_ADDRESS` unset, `:jvmTest` failed 150 of 331 tests and `:linuxX64Test` failed 180
 of 301. A green run of those tasks is therefore real evidence; a green bare `./gradlew test` is not.
+(`:jvmTest` was 135 of 331 before #227 removed the suites that returned early — and so reported a
+pass — instead of failing when they could not reach a bus.)
 
 The `samples/bluez-scan` and blue-falcon-sdbus integration suites need a real BlueZ adapter and a
 BLE peripheral, so they cannot run in CI.
--- a/COVERAGE_TODO.md
+++ b/COVERAGE_TODO.md
@@ -19,8 +19,8 @@ This backlog is intentionally ordered by priority and should be worked top-to-bo
   - [x] Added JVM session-bus coverage for distinct real connections and connection-level ObjectManager signal routing (`JvmRealBusIntegrationTest`).
   - [x] Added JVM session-bus coverage for `GetManagedObjects` returning child interface/property data (`JvmRealBusIntegrationTest`).
   - [x] Added JVM session-bus coverage for signal callback sender credential resolution (`JvmRealBusIntegrationTest`).
-  - [x] Added JVM system-bus coverage for distinct real connections (`createSystemBusConnection`) with graceful skip when unavailable (`JvmRealBusIntegrationTest`).
-  - [x] Added JVM system-bus ObjectManager `GetManagedObjects` coverage using unique-name routing with graceful skip when unavailable (`JvmRealBusIntegrationTest`).
+  - [x] Added JVM system-bus coverage for distinct real connections (`createSystemBusConnection`), failing when the bus is unavailable (#227) (`JvmRealBusIntegrationTest`).
+  - [x] Added JVM system-bus ObjectManager `GetManagedObjects` coverage using unique-name routing, failing when the bus is unavailable (#227) (`JvmRealBusIntegrationTest`).
 - Done when:
   - [x] `:jvmTest` is green,
   - [x] `:cross_test:jvmTest` is green,
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireServeExternalTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireServeExternalTest.kt
@@ -37,6 +37,7 @@ import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 import kotlinx.coroutines.runBlocking
+import org.junit.Assume.assumeTrue
 
 /**
  * Proves an object exported by a JVM sdbus-kotlin service on the OWNED wire backend is reachable by
@@ -51,20 +52,25 @@ import kotlinx.coroutines.runBlocking
  *    busctl gets correct results.
  *
  * The owned wire connection is now the only JVM backend (epic #93 phase 6 retired dbus-java), so
- * this always runs. Skips cleanly when no session bus or no `busctl` is available.
+ * this always runs. `busctl` is genuinely optional, so its absence is a real skip; a missing
+ * session bus is not — every job that runs this wraps the build in `dbus-run-session` — so that
+ * FAILS. Neither is an early return, which JUnit records as a pass (issue #227).
  */
 class WireServeExternalTest {
 
     @Test
     fun externalBusctl_reachesServedObject() = runBlocking {
-        val sessionBus = System.getenv("DBUS_SESSION_BUS_ADDRESS")
-        if (sessionBus == null) {
-            println("[WireServeExternalTest] SKIP: no DBUS_SESSION_BUS_ADDRESS.")
-            return@runBlocking
+        val sessionBus = checkNotNull(System.getenv("DBUS_SESSION_BUS_ADDRESS")) {
+            "DBUS_SESSION_BUS_ADDRESS is unset: this test serves an object on the session bus " +
+                "and hands its address to busctl. Run the suite under " +
+                "`dbus-run-session -- ./gradlew …`."
         }
         val busctl = findExecutable("busctl")
         if (busctl == null) {
-            println("[WireServeExternalTest] SKIP: busctl not on PATH.")
+            assumeTrue(
+                "busctl is not on PATH, so no external peer can call the served object.",
+                false
+            )
             return@runBlocking
         }
 
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireSignalEmissionExternalTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireSignalEmissionExternalTest.kt
@@ -35,6 +35,7 @@ import kotlin.random.Random
 import kotlin.test.Test
 import kotlin.test.assertTrue
 import kotlinx.coroutines.runBlocking
+import org.junit.Assume.assumeTrue
 
 /**
  * Proves that a signal our object emits actually traverses the BUS and reaches an INDEPENDENT
@@ -47,21 +48,26 @@ import kotlinx.coroutines.runBlocking
  *  - dbus-java backend: always emitted over the real wire, so it captures it too.
  *
  * JVM-only (cross_test `jvmTest`) because the owned-connection backend is JVM-only and toggle-gated;
- * the native backend's over-the-wire emission is already covered elsewhere. Skips cleanly when no
- * session bus or no `dbus-monitor` is available.
+ * the native backend's over-the-wire emission is already covered elsewhere. `dbus-monitor` is
+ * genuinely optional, so its absence is a real skip; a missing session bus is not — every job that
+ * runs this wraps the build in `dbus-run-session` — so that FAILS. Neither is an early return,
+ * which JUnit records as a pass (issue #227).
  */
 class WireSignalEmissionExternalTest {
 
     @Test
     fun emittedSignal_reachesExternalDbusMonitor() = runBlocking {
-        val sessionBus = System.getenv("DBUS_SESSION_BUS_ADDRESS")
-        if (sessionBus == null) {
-            println("[WireSignalEmissionExternalTest] SKIP: no DBUS_SESSION_BUS_ADDRESS.")
-            return@runBlocking
+        checkNotNull(System.getenv("DBUS_SESSION_BUS_ADDRESS")) {
+            "DBUS_SESSION_BUS_ADDRESS is unset: this test emits a signal onto the session bus " +
+                "and watches for it with dbus-monitor --session. Run the suite under " +
+                "`dbus-run-session -- ./gradlew …`."
         }
         val dbusMonitor = findExecutable("dbus-monitor")
         if (dbusMonitor == null) {
-            println("[WireSignalEmissionExternalTest] SKIP: dbus-monitor not on PATH.")
+            assumeTrue(
+                "dbus-monitor is not on PATH, so no external observer can see the signal.",
+                false
+            )
             return@runBlocking
         }
 
--- a/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmCredentialsTest.kt
+++ b/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmCredentialsTest.kt
@@ -16,14 +16,12 @@ import kotlinx.coroutines.withTimeout
  * is host-dependent — a label where SELinux is enforcing, or a thrown [SdbusException] otherwise.
  * The native backend covers the same surface against the live POSIX identity in
  * CredentialsIntegrationTest.
+ *
+ * Needs a real session bus, and every job that runs it wraps the build in `dbus-run-session`, so an
+ * unreachable bus is a broken environment: the factory's throw is left to FAIL the test rather than
+ * swallowed into an early return that JUnit would record as a pass (issue #227).
  */
 class JvmCredentialsTest {
-
-    private fun connectOrNull(connect: () -> Connection): Connection? = try {
-        connect()
-    } catch (e: Exception) {
-        null
-    }
 
     @Test
     fun receivedSignalResolvesSupportedCredentialsAndRejectsUnsupported() = runBlocking {
@@ -37,12 +35,8 @@ class JvmCredentialsTest {
         val expectedUid = unix.uid.toUInt()
         val expectedGid = unix.gid.toUInt()
 
-        val serverConnection = connectOrNull { createSessionBusConnection(service) }
-            ?: return@runBlocking
-        val proxyConnection = connectOrNull { createSessionBusConnection() } ?: run {
-            serverConnection.release()
-            return@runBlocking
-        }
+        val serverConnection = createSessionBusConnection(service)
+        val proxyConnection = createSessionBusConnection()
 
         val pidSeen = CompletableDeferred<Int>()
         val uidSeen = CompletableDeferred<UInt>()
--- a/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmRealBusIntegrationTest.kt
+++ b/src/jvmTest/kotlin/com/monkopedia/sdbus/JvmRealBusIntegrationTest.kt
@@ -7,17 +7,15 @@ import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withTimeout
 
+/**
+ * These tests need a *real* bus, and every job that runs them wraps the build in
+ * `dbus-run-session` (exporting `DBUS_SYSTEM_BUS_ADDRESS` alongside the session one), so an
+ * unreachable bus here is a broken environment rather than a legitimate absence. The connection
+ * factories throw an [SdbusException] ("Failed to open bus", issue #81) and that is left to FAIL
+ * the test: the early return they used to be swallowed into was recorded by JUnit as a PASS for a
+ * case that connected to nothing (issue #227).
+ */
 class JvmRealBusIntegrationTest {
-
-    // These tests need a *real* bus. The connection factories throw when the bus is
-    // unreachable (issue #81) instead of silently returning the in-process stub backend, so
-    // "no usable bus in this environment" now surfaces as a thrown SdbusException -- treat it as a
-    // skip, the way the old ":jvm-stub" unique-name gate did.
-    private fun connectOrNull(connect: () -> Connection): Connection? = try {
-        connect()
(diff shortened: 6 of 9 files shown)


monkopedia-reviewer (changes requested):
Reviewed at `dc70837`. I re-ran the measurement independently rather than reading the diff, in two
fresh worktrees (`44682aa` = merge-base, `dc70837` = head), `--no-daemon` throughout so the test
workers actually inherit the manipulated environment rather than a stale daemon's copy.

**Every claim in the body reproduces.** The classification is right, the split of "bus = fail /
external tool = skip" is better than the issue's suggested default, and it does not conflict with
`c5f874c`. Two documentation statements that this change makes false are the only thing holding it.

---

## What I measured (`<testsuite …>` attributes, mine, not the body's)

`:jvmTest :cross_test:jvmTest`, 50 suites, 381 tests in every direction.

**No bus (`env -u DBUS_SESSION_BUS_ADDRESS …`) — before → after:**

| suite | `44682aa` (before) | `dc70837` (after) |
|---|---|---|
| `JvmRealBusIntegrationTest` | `tests="7" skipped="0" failures="0"` | `tests="7" skipped="0" failures="5"` |
| `JvmCredentialsTest` | `tests="1" skipped="0" failures="0"` | `tests="1" skipped="0" failures="1"` |
| `DBusWireConnectionTest` | `tests="9" skipped="0" failures="0" time="0.003"` | `tests="9" skipped="0" failures="9"` |
| `WireServeExternalTest` | `tests="1" skipped="0" failures="0"` | `tests="1" skipped="0" failures="1"` |
| `WireSignalEmissionExternalTest` | `tests="1" skipped="0" failures="0"` | `tests="1" skipped="0" failures="1"` |
| **whole run** | `failures=139` | `failures=156` |

The 9-real-bus-tests-in-3ms line is the speed tell doing its usual work. The two
`JvmRealBusIntegrationTest` cases that still pass after the fix are the **system**-bus pair — this
host has a live system bus at `/run/dbus/system_bus_socket`, so they genuinely connected. That is
the honest result, as the body says.

Subtracting the four `cross_test` failures that are already there at the base gives **135 → 150 for
`:jvmTest` alone**, which corroborates the body's figure exactly.

**With a bus (CI's env: `dbus-run-session` + `DBUS_STARTER_BUS_TYPE`/`DBUS_STARTER_ADDRESS`/
`DBUS_SYSTEM_BUS_ADDRESS`) — before vs after:** all 50 `<testsuite …>` lines, timestamps and times
stripped, **diff empty**. `failures=0` both sides, `skipped=44` both sides (the dbusmock suites,
unaffected by this change — `DBUSMOCK_REQUIRED` is unset locally). No coverage lost.

```
<testsuite name="JvmRealBusIntegrationTest[jvm]" tests="7" skipped="0" failures="0" errors="0" time="0.023">
<testsuite name="DBusWireConnectionTest[jvm]"    tests="9" skipped="0" failures="0" errors="0" time="2.091">
```

`DBusWireConnectionTest` at **2.091 s with a bus vs 0.008 s without** is the counterfactual that
makes the green mean something.

**Skip direction (bus present, `busctl` + `dbus-monitor` removed from `PATH` via a symlink farm;
positive control: the farm still resolves `dbus-run-session`, negative control: no `busctl`):**

```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="1" failures="0" errors="0">
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="1" failures="0" errors="0">
```

`ktlintCheck apiCheck` on the head: BUILD SUCCESSFUL. Remote CI: all 15 checks green.

## Classification — I pushed on this and it holds

- **`assumeTrue(reason, false)` is not a third mechanism.** It is literally the body of `skipTest`'s
  JVM actual (`DbusmockHarness.jvm.kt:30`) from `c5f874c`, and `CrossRuntimeInteropSmokeTest:44`
  already imports `org.junit.Assume.assumeTrue` directly in the same source set. The trailing
  `return@runBlocking` after it is unreachable at runtime but required for the smart cast — the same
  shape `DbusmockSmokeTest:63-64` and `DbusmockBluezTest:548-549` already use. Consistent.
- **Fail-on-missing-bus in `:cross_test:jvmTest` does not contradict `c5f874c`.** That commit's
  skip-by-default is keyed on *"a dbusmock peer cannot start"*, which bundles python/module/bus into
  one optional-peer precondition, and `full-tests-x64` sets `DBUSMOCK_REQUIRED=1` so it fails there
  anyway. Failing on a missing bus is already the module norm — `CrossModuleSignalIntegrationTest`
  and `EventLoopStartIdempotencyTest` fail hard without one at the base commit (I measured them).
- **The workflow claim checks out.** Every job that runs these tasks wraps them:
  `arm-build-test.yaml:74` (`allTests test`, exporting `DBUS_SYSTEM_BUS_ADDRESS`) and
  `stress-matrix.yaml:256` (`-PgcSoak`). The only unwrapped Gradle invocations are
  `:stress_test:jvmInteropStressTest`, `assemble`, `dokka` and `ktlintCheck apiCheck` — none of
  which run a changed suite.
- **The scope sweep is complete.** I re-grepped with `/usr/bin/grep -ra` (the shell's ugrep skips
  binary-flagged files here) plus a positive control, for `?: return`, `return@runBlocking`,
  `println("SKIP")`-then-return and swallowing `catch`. Everything left is either a decoder helper
  returning `null`, a `skipTest(...)`-then-return in the dbusmock suites, or the sub-case `SKIP`
  printlns the body scopes out — those sit inside tests that still assert plenty, correctly left
  alone.
- **`NativeInteropPeerTest` is correctly excluded.** Its 14 gates select which role a spawned peer
  process plays, one per invocation; they are dispatch, not a prerequisite check, and
  `DbusmockHarness.linuxX64.kt:48` confirms K/N has no runtime-skip primitive to convert them to.
- **The `DBusWireConnectionTest` fd restructure is sound.** `createPipePair` moving above the
  connections means an `assertNotNull` failure now happens with nothing open; the new
  `assertTrue(server.unixFdNegotiated && client.unixFdNegotiated, …)` sits inside the existing `try`
  whose `finally` closes both fds and both connections, so it leaks nothing, and it is a
  strengthening rather than over-strictness — a regression in our own `NEGOTIATE_UNIX_FD` handshake
  used to silently blank the only SCM_RIGHTS test. Verified green with a bus, 9/9.

---

## Blocking: this change falsifies two documents and does not update them

Small, but this is the exact failure #227 was filed for, in mirror image — #227 exists because
`AGENTS.md` stated a principle the code did not honour, and merging this leaves `AGENTS.md` stating
a measurement the code no longer produces.

**1. `AGENTS.md:57-59`** — the sentence whose whole job is to be checkable:

> The real suites fail loudly rather than silently when the bus is missing — measured on `main` with
> `DBUS_SESSION_BUS_ADDRESS` unset, `:jvmTest` failed **135 of 331** tests and `:linuxX64Test` failed
> 180 of 301.

On merge, `main` produces **150 of 331** — a number this PR's own body already contains. Please
update the figure (`:linuxX64Test`'s 180/301 is untouched by this change and stays). The claim it
supports gets *stronger*, which is the argument for keeping the digits right rather than for
shrugging at them.

**2. `COVERAGE_TODO.md:22-23`** — both lines assert behaviour this PR deliberately removes:

```
- [x] Added JVM system-bus coverage for distinct real connections (`createSystemBusConnection`)
      with graceful skip when unavailable (`JvmRealBusIntegrationTest`).
- [x] Added JVM system-bus ObjectManager `GetManagedObjects` coverage using unique-name routing
      with graceful skip when unavailable (`JvmRealBusIntegrationTest`).
```

There is no graceful skip any more, by design. Drop the clause, or replace it with "fails when the
bus is unavailable (#227)". (These are also the two cases whose test methods this PR renames off
`…WhenRealBusIsAvailable`, so the rename and the doc line drifted apart in the same commit.)

## Non-blocking, but please file it before this merges

**The precedent has two halves and this adopts one.** `c5f874c` paired the honest skip with
`DBUSMOCK_REQUIRED=1` in the job that installs the dependency, precisely so the skip could not
silently delete coverage *there*. `busctl` / `dbus-monitor` now get half (a) and none of half (b):
if either disappears from the `full-tests-x64` runner image, `WireServeExternalTest` — the repo's
only proof that a JVM-served object is reachable by a genuinely external process (#90) — reports
`skipped="1"` and the job stays green. That is strictly better than the phantom pass it replaces, so
I am not blocking on it, but it is the same hole one notch out, and nothing currently asserts those
two binaries exist on the runner. A `REQUIRED`-style flag honoured by the `findExecutable` gates and
set in `full-tests-x64` would close it in about six lines. Either add it here or open an issue and
say so in the body.

---

Everything else — the classification, the deletion of the four `connectOrNull` helpers, the KDoc
rewrites, the test renames, the fd-test restructure, the scoping decisions — I reviewed to merge
standard and it is sound. Fix the two doc lines and I will approve on the next push.

monkopedia-reviewer (approved):
Re-reviewed at `4ab9ba5`. Both blocking items are fixed. Approving.

## What I checked this round

**The delta is docs-only, and I confirmed that rather than taking it on report.** `git diff dc70837..4ab9ba5` is exactly `AGENTS.md` (+3/-1) and `COVERAGE_TODO.md` (+2/-2), 2 files, +5/-3. No test source and no production source moved.

**So I am explicitly relying on the measurement I took at `dc70837` rather than re-running it** — that is sound here only because the diff above proves the test sources are byte-identical to the head I measured. Recording that as a stated dependency, not an assumption.

**`AGENTS.md:57-60` — the figure matches my own derivation, not the PR body's.** I re-split my saved `<testsuite …>` data by module (13 cross_test suites — the 9 dbusmock + `CrossModuleSignalIntegrationTest` + `EventLoopStartIdempotencyTest` + the two `Wire*ExternalTest` — and 37 in `:jvmTest`):

```
pre-fix,  no bus:  :jvmTest suites=37 tests=331 failures=135
post-fix, no bus:  :jvmTest suites=37 tests=331 failures=150
```

Numerator **and** denominator both land on the sentence as written. `:linuxX64Test`'s 180 of 301 is untouched by this change and correctly left alone.

**`COVERAGE_TODO.md:22-23`** — both "with graceful skip when unavailable" clauses are now "failing when the bus is unavailable (#227)", which is what the code does. The lines and the test names they describe are back in agreement.

**Issue #229 is an adequate record of the gap** — it carries the `c5f874c` two-halves argument, the #90 / phase-3b stakes, the `EXTERNAL_TOOLS_REQUIRED` shape mirroring the existing convention rather than inventing a second one, and the caveat that whoever takes it must confirm both binaries are on the `full-tests-x64` runner before making their absence fatal. It also adds something I had not written down: `full-tests-x64` only uploads test-results on failure, so from the job log a real external-peer run and a skipped one are currently indistinguishable. That is the sharper statement of the gap.

**Merge state:** 15 checks SUCCESS, 2 SKIPPED (`cleaner-soak` / `nightly-stress`, expected on PRs), rollup head = `4ab9ba5`, `mergeStateStatus: CLEAN`, `behind_by: 0` against `main`, and zero runs on this branch with `run_attempt > 1` — every green is a first-attempt green.

## One non-blocking nit, deliberately not held for

The new `AGENTS.md` parenthetical reads:

> (`:jvmTest` was 135 of 331 before #227 removed the suites that returned early …)

Parsed strictly, the object of "removed" is "the suites", which would mean suites were deleted — what was removed is the early *returns*. The sentence disproves the misreading on its own (the denominator is 331 on both sides of the change, so nothing was deleted), so this is a wording nit and not worth another round on a green test-only PR. Worth folding into the next `AGENTS.md` touch: "before #227 made the suites that returned early fail instead".

## Standing verdict from round 1, unchanged

The classification (bus → fail, external tool → skip) is right and better than the issue's suggested default; `assumeTrue(reason, false)` is the same primitive `skipTest`'s JVM actual already uses rather than a third mechanism; failing on a missing bus in `cross_test` does not contradict `c5f874c`; the `NativeInteropPeerTest` exclusion holds; the scope sweep is complete; and the `DBusWireConnectionTest` fd restructure leaks nothing and strengthens rather than over-constrains. Measured both directions myself at `dc70837`, including a `dbus-run-session` run that is line-identical to the pre-fix baseline at 381 tests / 0 failures.

Approving and merging.

Issue #230, opened 2026-08-08
Capability flags carry no provenance: one encodes an undecided divergence, another is dead and its branches unreachable


monkopedia-coder:
This repo uses per-backend capability flags to encode cross-backend divergence in tests — the mechanism #141 mandates and #213 built on. It works. But the flags carry no convention for **recording why they exist**, and two of them have drifted in opposite directions as a result.

Found by a sweep for "a conditional wrapping an assertion with no comment explaining it", run across 94 test files in `src/commonTest`, `src/jvmTest`, `src/nativeTest`, `cross_test/src`, `stress_test/src`. 19 candidates, 17 benign. These are the two that are not.

## 1. `preservesWireSignatureForEmptyCollections` cites no issue, and is formatted exactly like two flags that mean the opposite

`src/commonTest/…/integration/EmptyCollectionWireTest.kt:397`:

```kotlin
assertEmpty(withTimeout(2_000) { decoded.await() })
if (EmptyCollectionWireSupport.preservesWireSignatureForEmptyCollections) {
    assertEquals(expectedSignature, withTimeout(2_000) { observedSignature.await() },
        "wire signature of empty signal body")
}
```

The flag is `true` on native and **`false` on JVM** (`EmptyCollectionWireSupport.jvm.kt:6`), so the wire-signature assertion **never executes on the JVM backend**. The rationale exists one hop away in the `expect` declaration's KDoc — the JVM backend reconstructs bodies as plain Kotlin values with no retained per-element type — so this is a *documented* divergence, not a hidden one.

**The problem is that it is indistinguishable from an encoded bug.** The repo's two other capability flags of this exact kind each name the issue they encode:

- `peerStructMarshallingSupported` → cites #71
- `peerErrorNameMappingSupported` → cites #72, with a `KNOWN JVM BACKEND BUG` comment

`preservesWireSignatureForEmptyCollections` names nothing. So it reads as "accepted limitation" while being formatted identically to two flags that mean "unfixed bug we chose to encode". Whether the JVM backend's inability to report a received empty collection's element signature is a permanent limitation of the wire representation or a gap someone should close is a real question, and **nothing in the tree records that anyone decided it**.

Minimum fix: give the flag the same provenance its siblings have — either an issue number, or an explicit "this is a permanent consequence of X, not a bug" note. If it turns out nobody ever made that call, that is worth knowing before 1.1.

[...]

PR #234, opened 2026-08-08
test: pin sd-bus's PropertiesChanged flag contract and record the JVM divergence (Refs #193)

### LINKED ISSUE #193: #159's annotation→vtable-flag fix is native-only: the JVM backend reads no `Flags` at all, so served introspection still disagrees with the XML

monkopedia-coder:
**Test-only. `Refs #193` — this records reality and commits to nothing.** No file under
`src/jvmMain/**` or `src/nativeMain/**` moves; #193 stays open; #231 stays a draft.

This is **option 4** from #231 §5e — the one that spike surfaced and recommended *over* merging its
own production change. #231 priced making the JVM `PropertiesChanged` emitter honour property flags
and concluded **do not merge**: the change is breaking twice over (it silently drops broadcasts for
`const`/`false` properties, and it makes `Object.notifying()` throw `SdbusException` from an ordinary
Kotlin assignment via the list overload's `-EDOM`). That decision is the owner's and is untouched
here.

What stands independently of it is #231's test: **native already implements sd-bus's real contract
and nothing in this repo verified that** — a gap #193's own closing section complains about. So the
test lands; the fix does not.

---

## 1. What sd-bus actually does — re-derived, not taken on #231's word

`emit_properties_changed_on_interface`, `src/libsystemd/sd-bus/bus-objects.c`, checked against
**systemd v257** (line numbers below are that file; #231 quoted v261, and the two agree):

**Path A — `names == NULL`** (the no-argument overload). *Skip:*

```c
if (v->flags & SD_BUS_VTABLE_PROPERTY_EMITS_INVALIDATION) { has_invalidating = true; continue; }
if (!(v->flags & SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE))    continue;   /* :2179 */
```

…and if that leaves nothing to say, **no signal is sent at all**:

```c
if (!has_invalidating && !has_changing)   /* :2193 */
        return 0;
```

**Path B — an explicit name list.** ***Reject, not skip:***

```c
assert_return(v->vtable->flags & SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE ||
              v->vtable->flags & SD_BUS_VTABLE_PROPERTY_EMITS_INVALIDATION, -EDOM);   /* :2141 */
```

`assert_return` *returns*; `-EDOM` propagates out of `sd_bus_emit_properties_changed_strv`, and
`ConnectionImpl.kt:631`'s `sdbusRequire(r < 0, …, -r)` turns it into a thrown `SdbusException`.
**Nothing is sent** — not even the named properties that would have been fine.

Our two overloads map onto those two paths: `emitPropertiesChangedSignal(iface)` → `emptyList()` →
`names = null` → Path A; `emitPropertiesChangedSignal(iface, listOf(…))` → Path B.

## 2. The test

One new file, `src/commonTest/…/integration/PropertiesChangedFlagParityTest.kt`, run on both
targets. It serves two interfaces on one object — a *mixed* one carrying a default, an
`invalidates`, a `const` and a `false` property, and a *quiet* one carrying only `const` + `false` —
and pins five cases:

| Test | Native (flag `true`) | JVM (flag `false`) |
| --- | --- | --- |
| `noArgEmissionSkipsSilentPropertiesOnlyWhereTheBackendHonoursTheFlags` | `changed = {Changing}`, `invalidated = [Invalidating]` | `changed = {Changing, Invalidating, Const, Silent}` with values, `invalidated = []` |
| `noArgEmissionSendsNothingAtAllWhereTheBackendHonoursTheFlags` | **no signal at all** for the quiet interface | a signal carrying `{Const, Silent}` |
| `namedEmissionOfAnInvalidatingPropertyIsNameOnlyWhereTheBackendHonoursTheFlags` | `changed = {}`, `invalidated = [Invalidating]` | `changed = {Invalidating}` with its value |
| `namedEmissionOfAConstPropertyIsRejectedOnlyWhereTheBackendHonoursTheFlags` | throws `SdbusException` | succeeds, emits `{Const}` |
| `namedEmissionOfASilentPropertyIsRejectedOnlyWhereTheBackendHonoursTheFlags` | throws `SdbusException` | succeeds, emits `{Silent}` |

Two notes on how it is written:

- **The "no signal at all" case needs a negative, so it uses an ordering sentinel** rather than a
  sleep: emit on the quiet interface, then on the mixed one, wait for the *mixed* signal, and assert
  the quiet deferred is still incomplete. The daemon preserves per-sender ordering, so the sentinel's
  arrival proves the quiet emission is never coming rather than merely late.
- **Both branches assert; neither is weakened.** The `false` branch is not a skip — it pins what the
  JVM backend does *today*, so the divergence cannot change unobserved in either direction.

## 3. The capability flag, and its provenance

The JVM divergence is recorded with the repo's existing mechanism (#141/#213), as one new
`expect val` in `TestBackendCapabilities`. Per #230 — flags that cite nothing are indistinguishable
from encoded bugs — it names the issue **and** says explicitly that the divergence is undecided:

```kotlin
/**
 * Whether the active backend applies a property's
 * [com.monkopedia.sdbus.Flags.PropertyUpdateBehaviorFlags] when it builds the `PropertiesChanged`
 * payload -- skipping properties that announce no change, putting `invalidates` properties in the
 * name-only array, and rejecting an explicitly named property that announces neither.
 *
 * **This flag records an OPEN DECISION on #193, not an accepted limitation.** Unlike the
 * `backendServes*` flags above -- introspection metadata, where each backend's answer is simply
 * what it is -- this is the signal payload a client consumes, and the two backends give different
 * answers to the same question. Which one is right has not been decided: matching native would
 * silently stop broadcasts for `const`/`false` properties and make an ordinary assignment through
 * the `Object.notifying` delegate throw, on a released artifact, so PR #231 priced that change and
 * did not merge it. #193 stays open until the owner rules.
 *
 * So do not read `false` as "fine on JVM" or `true` as "settled in native's favour". Whoever
 * closes #193 flips an `actual` or deletes the flag; until then [PropertiesChangedFlagParityTest]
 * pins both sides so neither can move unobserved.
 */
internal expect val backendHonoursEmitsChangedSignalOnEmit: Boolean
```

The two `actual`s carry the same provenance — native cites `bus-objects.c` and marks #193 OPEN; JVM
opens with `KNOWN JVM/NATIVE DIVERGENCE, #193 -- OPEN, not an accepted limitation`, matching the
house style of `peerErrorNameMappingSupported`'s `KNOWN JVM BACKEND BUG` (#72).

No second mechanism was invented, and no `backendServes*` flag was touched.

## 4. Evidence

All runs under `dbus-run-session`, `JAVA_HOME=/usr/lib/jvm/java-21-openjdk`, each preceded by
`find . -type d -name test-results -prune -exec rm -rf {} +`. `BUILD SUCCESSFUL` is not the evidence;
counts come from the fresh JUnit XML.

### GREEN — full suites, both targets, `9ca8c3a` + this commit

| Run | Result |
| --- | --- |
| `:jvmTest` | **38 XML files — 336 tests, 0 failures, 0 errors, 0 skipped** |
| `:linuxX64Test` | **30 XML files — 306 tests, 0 failures, 0 errors, 0 skipped** |
| `ktlintCheck` + `apiCheck` | exit 0; `git status api/` and `git diff --stat api/` both empty |

The new suite is 5 tests on each target:

```
<testsuite name="PropertiesChangedFlagParityTest[jvm]" tests="5" skipped="0" failures="0" errors="0"
<testsuite name="linuxX64Test.…PropertiesChangedFlagParityTest" tests="5" skipped="0" failures="0" errors="0"
```

Reproduced independently on a second host (adolin, fresh clone of this branch): `BUILD SUCCESSFUL`,
`jvmTest 38 XML / 336 tests / 0 failures`, `linuxX64Test 30 XML / 306 tests / 0 failures`.

### The native pin can fail — negative control (#216's precedent)

A pin that cannot go red is the defect #219/#227 spent this week removing, so it was made to fail:
**invert the native `actual` to `false`** (i.e. claim native does not honour the flags) and the whole
suite goes red — **5 tests, 5 failures, 0 errors, 0 skipped** — with each failure quoting sd-bus's
real behaviour:

| Test | `<failure message=…>` |
| --- | --- |
| `noArgEmissionSkipsSilentPropertiesOnlyWhereTheBackendHonoursTheFlags` | `kotlin.AssertionError: changed_properties. Expected <[Changing, Invalidating, Const, Silent]>, actual <[Changing]>.` |
| `noArgEmissionSendsNothingAtAllWhereTheBackendHonoursTheFlags` | `kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 5000 ms` — the quiet interface's signal never arrives, because sd-bus never sent one |
| `namedEmissionOfAnInvalidatingPropertyIsNameOnlyWhereTheBackendHonoursTheFlags` | `kotlin.AssertionError: changed_properties. Expected <[Invalidating]>, actual <[]>.` |
| `namedEmissionOfAConstPropertyIsRejectedOnlyWhereTheBackendHonoursTheFlags` | `com.monkopedia.sdbus.SdbusException: System.Error.EDOM: Failed to emit PropertiesChanged signal (Numerical argument out of domain)` |
| `namedEmissionOfASilentPropertyIsRejectedOnlyWhereTheBackendHonoursTheFlags` | `com.monkopedia.sdbus.SdbusException: System.Error.EDOM: Failed to emit PropertiesChanged signal (Numerical argument out of domain)` |

Those last two are the `-EDOM` of §1 arriving at a Kotlin caller, verbatim, from a live sd-bus. The
flag was restored afterwards and `git status --porcelain` came back empty.

### No production change

```
 .../integration/PropertiesChangedFlagParityTest.kt | 255 +++++++++++++++++++++
 .../sdbus/integration/TestBackendCapabilities.kt   |  20 ++
 .../integration/TestBackendCapabilities.jvm.kt     |   5 +
 .../integration/TestBackendCapabilities.native.kt  |   6 +
 4 files changed, 286 insertions(+)
```

`git diff --name-only 9ca8c3a | grep Main/` → empty. Nothing under `src/jvmMain/**` or
`src/nativeMain/**`, no deletions anywhere, and `api/*.api` unchanged.

## 5. What this deliberately does NOT do

- **No production change**, on either backend.
- **No `docs/BACKENDS.md` edit.** Line 35 currently marks `PropertiesChanged` emission ✅/✅, which
  is now demonstrably incomplete — but amending it is #193's *option 1* ("document and freeze"), and
  choosing option 1 is the owner's call, not a side effect of landing a test. Left for whoever
  settles #193.
- **No claim about which backend is right.** The test pins both; the flag says so in as many words.

Refs #193

### DIFF
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/integration/PropertiesChangedFlagParityTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/integration/PropertiesChangedFlagParityTest.kt
@@ -1,0 +1,255 @@
+package com.monkopedia.sdbus.integration
+
+import com.monkopedia.sdbus.Flags.PropertyUpdateBehaviorFlags.CONST_PROPERTY_VALUE
+import com.monkopedia.sdbus.Flags.PropertyUpdateBehaviorFlags.EMITS_INVALIDATION_SIGNAL
+import com.monkopedia.sdbus.Flags.PropertyUpdateBehaviorFlags.EMITS_NO_SIGNAL
+import com.monkopedia.sdbus.InterfaceName
+import com.monkopedia.sdbus.Object
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.PropertiesProxy
+import com.monkopedia.sdbus.PropertyName
+import com.monkopedia.sdbus.SdbusException
+import com.monkopedia.sdbus.ServiceName
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.Variant
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.createBusConnection
+import com.monkopedia.sdbus.createObject
+import com.monkopedia.sdbus.createProxy
+import com.monkopedia.sdbus.onSignal
+import com.monkopedia.sdbus.prop
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+
+/**
+ * Pins what each backend puts in a `PropertiesChanged` signal for a property carrying a non-default
+ * [com.monkopedia.sdbus.Flags.PropertyUpdateBehaviorFlags].
+ *
+ * `EmitsChangedSignal` is not introspection decoration -- it is an instruction to the emitter, and
+ * sd-bus acts on it in `emit_properties_changed_on_interface` (`bus-objects.c`) with two distinct
+ * rules, one per overload of [Object.emitPropertiesChangedSignal]:
+ *
+ * - **No explicit names** (`names == NULL`): a property flagged `EMITS_INVALIDATION` goes to the
+ *   name-only array, a property with neither `EMITS_CHANGE` nor `EMITS_INVALIDATION` is skipped,
+ *   and if that leaves nothing to say **no signal is sent at all**.
+ * - **An explicit name list**: a named property carrying neither flag is not skipped, it is
+ *   *rejected* -- `assert_return(..., -EDOM)` aborts the whole emission, so nothing is sent and
+ *   `ConnectionImpl` turns the errno into a thrown [SdbusException].
+ *
+ * Nothing in this repo verified either rule before, which is a gap #193 itself raises: its
+ * divergence table was derived from reading C rather than from a run. The JVM wire backend reads no
+ * flags at all, so it disagrees on every case here -- that divergence is recorded in
+ * [backendHonoursEmitsChangedSignalOnEmit] and pinned in both directions rather than asserted away.
+ *
+ * This test records reality and commits to nothing: closing the divergence is a breaking change
+ * that PR #231 priced and deliberately did not merge, and #193 remains open. Refs #193.
+ */
+class PropertiesChangedFlagParityTest {
+
+    private class Emission(
+        val changed: Map<PropertyName, Variant>,
+        val invalidated: List<PropertyName>
+    )
+
+    private class Fixture(
+        val obj: Object,
+        /** Carries [CHANGING], [INVALIDATING], [CONST] and [SILENT]. */
+        val mixedIface: InterfaceName,
+        /** Carries [CONST] and [SILENT] only, so nothing on it announces a change. */
+        val quietIface: InterfaceName,
+        val mixedEmission: CompletableDeferred<Emission>,
+        val quietEmission: CompletableDeferred<Emission>
+    )
+
+    /**
+     * Serves both interfaces on one object, with a proxy already subscribed to `PropertiesChanged`.
+     */
+    private suspend fun withFixture(body: suspend (Fixture) -> Unit) {
+        val id = Random.nextInt(100_000, 999_999)
+        val service = ServiceName("com.monkopedia.sdbus.pcflags$id")
+        val path = ObjectPath("/com/monkopedia/sdbus/pcflags$id")
+        val mixedIface = InterfaceName("com.monkopedia.sdbus.pcflags$id.Mixed")
+        val quietIface = InterfaceName("com.monkopedia.sdbus.pcflags$id.Quiet")
+
+        val server = createBusConnection(service)
+        val client = createBusConnection()
+        val obj = createObject(server, path)
+        val mixedReg = obj.addVTable(mixedIface) {
+            prop(CHANGING) { withGetter { 1 } }
+            prop(INVALIDATING) {
+                +EMITS_INVALIDATION_SIGNAL
+                withGetter { 2 }
+            }
+            prop(CONST) {
+                +CONST_PROPERTY_VALUE
+                withGetter { 3 }
+            }
+            prop(SILENT) {
+                +EMITS_NO_SIGNAL
+                withGetter { 4 }
+            }
+        }
+        val quietReg = obj.addVTable(quietIface) {
+            prop(CONST) {
+                +CONST_PROPERTY_VALUE
+                withGetter { 5 }
+            }
+            prop(SILENT) {
+                +EMITS_NO_SIGNAL
+                withGetter { 6 }
+            }
+        }
+        server.startEventLoop()
+        client.startEventLoop()
+        val proxy = createProxy(client, service, path)
+
+        val mixed = CompletableDeferred<Emission>()
+        val quiet = CompletableDeferred<Emission>()
+        val sigReg = proxy.onSignal(
+            PropertiesProxy.INTERFACE_NAME,
+            SignalName("PropertiesChanged")
+        ) {
+            call {
+                    changedInterface: InterfaceName,
+                    changed: Map<PropertyName, Variant>,
+                    invalidated: List<PropertyName>
+                ->
+                val emission = Emission(changed, invalidated)
+                if (changedInterface == mixedIface) mixed.complete(emission)
+                if (changedInterface == quietIface) quiet.complete(emission)
+            }
+        }
+
+        try {
+            body(Fixture(obj, mixedIface, quietIface, mixed, quiet))
+        } finally {
+            sigReg.release()
+            quietReg.release()
+            mixedReg.release()
+            proxy.release()
+            obj.release()
+            client.stopEventLoop()
+            server.stopEventLoop()
+            client.release()
+            server.release()
+        }
+    }
+
+    @Test
+    fun noArgEmissionSkipsSilentPropertiesOnlyWhereTheBackendHonoursTheFlags() = runBlocking {
+        withFixture { fixture ->
+            fixture.obj.emitPropertiesChangedSignal(fixture.mixedIface)
+
+            val emission = withTimeout(5_000) { fixture.mixedEmission.await() }
+            if (backendHonoursEmitsChangedSignalOnEmit) {
+                assertEquals(setOf(CHANGING), emission.changed.keys, "changed_properties")
+                assertEquals(1, emission.changed.getValue(CHANGING).get<Int>())
+                assertEquals(listOf(INVALIDATING), emission.invalidated, "invalidated_properties")
+            } else {
+                assertEquals(
+                    setOf(CHANGING, INVALIDATING, CONST, SILENT),
+                    emission.changed.keys,
+                    "changed_properties"
+                )
+                assertEquals(2, emission.changed.getValue(INVALIDATING).get<Int>())
+                assertEquals(emptyList(), emission.invalidated, "invalidated_properties")
+            }
+        }
+    }
+
+    @Test
+    fun noArgEmissionSendsNothingAtAllWhereTheBackendHonoursTheFlags() = runBlocking {
+        withFixture { fixture ->
+            // Nothing on quietIface announces a change, so sd-bus's names==NULL path leaves the
+            // message empty and returns without sending it. Emitting on mixedIface afterwards is
+            // the sentinel: both backends send that one, and the daemon preserves per-sender order,
+            // so its arrival proves the quiet emission is never coming rather than merely late.
+            fixture.obj.emitPropertiesChangedSignal(fixture.quietIface)
+            fixture.obj.emitPropertiesChangedSignal(fixture.mixedIface)
+
+            withTimeout(5_000) { fixture.mixedEmission.await() }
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):
Reviewed at head `b3d97c8e5458063fd42e2bbc89527add1739899c`. Checked all six standards named in the dispatch, independently rather than on the PR body's say-so:

**1. Zero production files moved.** Confirmed from the diff itself: 4 files, all under `src/commonTest/**`, `src/jvmTest/**`, `src/nativeTest/**`. Nothing under `src/jvmMain/**` or `src/nativeMain/**`.

**2. Capability flag provenance (#230's bar).** `backendHonoursEmitsChangedSignalOnEmit`'s KDoc and both `actual` comments cite #193 by number and state in as many words that this is an OPEN decision, not an accepted limitation — clearing #230's minimum ("cite the issue, or say it's a permanent limitation") by actually saying which of the two it is (neither, yet).

**3. The pin is load-bearing.** Reproduced the full cycle myself in a fresh clone at this SHA, not on the PR's word: baseline `linuxX64Test --tests "*PropertiesChangedFlagParityTest*"` → fresh XML `tests="5" skipped="0" failures="0"`. Flipped `backendHonoursEmitsChangedSignalOnEmit` to `false` in `TestBackendCapabilities.native.kt` → fresh XML `tests="5" skipped="0" failures="5"`, matching the PR body's claimed failure shapes (AssertionError on the two changed-set tests, TimeoutCancellationException on the "sends nothing" test, SdbusException on both named-rejection tests). Reverted → `git status --porcelain` empty, re-ran → `failures="0"` again. Also ran the JVM side of the same test class directly (`:jvmTest --tests "*PropertiesChangedFlagParityTest*"`) → `tests="5" skipped="0" failures="0"`, confirming the `false` branch isn't a skip.

**4. sd-bus claims.** Fetched `bus-objects.c` fresh from `github.com/systemd/systemd` at tag `v257` (not trusting any locally cached copy) and grepped `emit_properties_changed_on_interface` directly. Both rules check out at the exact cited lines: `:2179` skips a property lacking `EMITS_CHANGE` in the `names==NULL` path, `:2193`'s `if (!has_invalidating && !has_changing) return 0` sends nothing when that empties the message, and `:2141-2142`'s `assert_return(... EMITS_CHANGE || ... EMITS_INVALIDATION, -EDOM)` rejects rather than skips on the named-list path.

**5. Both branches assert something real.** Read all 5 tests — the `false` (JVM) branch always asserts specific keys/values/absence of exception, never a bare skip. Confirmed live via the `:jvmTest` run above.

**6. Scope.** No `docs/BACKENDS.md` edit, no other file touched. The PR's own "what this deliberately does NOT do" section is accurate.

**Quality:** the new test's naming (`XxxOnlyWhereTheBackendHonoursTheFlags`) and the `withFixture`/`Random.nextInt(100_000, 999_999)` id-uniqueness pattern match the established conventions in sibling parity tests (`VtableFlagIntrospectionTest`'s `...OnlyWhereTheBackendCarriesIt`, `AddMatchSenderParityTest`/`SignalPayloadParityTest`'s fixture id scheme) almost exactly. Reads like it was always there.

**CI:** all GitHub checks green at this head (`gh pr checks --watch`), including `full-tests-x64`; `mergeStateStatus` CLEAN, `mergeable` MERGEABLE.

Classification: `test_only` (no production/doc files touched) → default tier 1. No reviewer-judgment bump — internal-only flag, no public API surface, no scope creep.

Approving and merging.

PR #236, opened 2026-08-08
docs: document that the creds* accessors throw, and that a direct connection's sender is not authoritative (Refs #199)

### LINKED ISSUE #199: Direct connections stamp OUR credentials as the peer-supplied sender's identity
On a brokerless direct connection, the library will stamp **this process's real pid/uid/gid/SELinux context** onto an incoming message as the *sender's* credentials, when the peer-supplied `sender` string happens to resolve locally. A consumer reading `credsUid` for an authorization decision is therefore told the sender is us.

## Mechanism

`WireDbusBackend.kt:815-836` (`withLocalSenderCredentials`): if the wire `sender` resolves via `LocalJvmServiceRegistry.resolveLocalUniqueName`, our own credentials are attached to the message as the sender's. `resolveLocalUniqueName` (`WireDbusObject.kt:692-702`) matches unique names **and well-known aliases**.

## Why brokered buses are fine and direct connections are not

On a brokered bus this is safe by construction: `dbus-daemon` stamps `sender` authoritatively, so a peer cannot claim our name. The assumption holds.

In **direct mode** it does not. `addMatch` is a no-op and `getNameOwner` returns null (`DBusWireConnection.kt:500,519`), nothing stamps `sender`, and the single peer chooses it freely. The identity is derived from an attribute its own subject controls.

## Severity, stated with its precondition

**Low practical severity today.** A direct connection has exactly one peer by construction, so "an attacker could impersonate us" reduces to "the one party you deliberately connected to can lie about its name" — you already trusted it enough to connect. There is no privilege boundary being crossed in the common case.

**It is filed anyway, and filed publicly, because it is a contract defect consumers cannot see.** Anyone writing an authorization check against `credsUid` on a direct connection is relying on a guarantee this library does not provide there, and nothing in the KDoc says so. Disclosure serves the consumer more than it serves anyone else — the code is already public, and the mitigation ("do not authorize on credentials from a direct connection") is only actionable if it is written down.

## Fleet context

This is the fifth instance in this fleet of **identity derived from a non-unique or corruptible attribute**, after a shared `app_id`, a narrow digest, a shared filename, and a truncated agent id that was itself a valid agent id. This repo fixed a sixth in #174 (`":jvm-wire"` shared across all direct connections). The pattern is worth naming as a class rather than fixing case by case.

## Disposition

Two defensible resolutions, and this is the same shape as the three parity items settled on #141:

1. **Do not stamp local credentials when the connection has no authoritative sender** — i.e. treat direct mode as "sender unverified" and leave credentials absent rather than wrong. Absent is honest; wrong is not.
2. **Document it** on the credential accessors and on `createDirectBusConnection`, as a known limitation of brokerless mode.

Option 1 is a behaviour change on a released artifact, which is why this is not simply dispatched. Related to #141 (cross-backend parity) but distinct — that epic tracks JVM-vs-native divergence, and this is a JVM-vs-reality gap that native does not share.

monkopedia-coder:
**Documentation only. Refs #199 — this does not decide it, and #235 stays a draft.**

Same pattern as the #193 test-only extraction: take the part of a spike that is correct regardless of how the owner rules, and leave the ruling untouched.

## Why this is separable from the decision

#235 priced the behaviour change proposed in #199 and concluded *do not merge yet* — on sequencing, not severity: junixsocket 2.10.1 (already a direct dependency) exposes `AFUNIXSocket.getPeerCredentials()`, probed live as returning the peer's real pid/uid/gid. So the JVM could eventually report **correct** credentials on a direct connection rather than merely absent ones, which makes #199 a three-option decision, not two. Spending an un-deprecatable behaviour break now and then changing again later is worse for consumers than documenting now and changing once.

But two things #235 established are true under **every** option, and neither was written down:

1. **`credsPid` and the other `creds*` accessors throw when credentials are unavailable.** `requireJvmCredential` on JVM and `sdbusRequire` on native, both raising `SdbusException`; the native path was confirmed live as `System.Error.ENODATA`. The published KDoc said only *"if credentials are available"* and never stated that it throws — so a consumer could not tell it was calling a throwing accessor, and would not guard the read.
2. **A direct connection has no authoritative sender.** `dbus-daemon` stamps `sender` on a brokered bus; in brokerless direct mode nothing does, and the single peer supplies it. Credentials reached through that path are peer-asserted, not verified, and must not carry an authorization decision.

Both are true whether #199 ends in "leave credentials absent", "report the peer's real credentials via `SO_PEERCRED`", or "document only".

## What this deliberately does NOT say

The KDoc describes the accessors' contract **without blessing current behaviour**. It does not claim the present direct-connection behaviour is intended or settled; it marks what the accessors report there as **backend-dependent and unspecified**, and links #199 as the open question. A doc that closed #199 rhetorically would pre-empt the owner just as surely as merging the code would.

It also states the *contract* ("credentials on a direct connection are not authoritative") rather than any procedure — no impersonation recipe. The mechanism is already public in #199; this goes no further.

Note that #235's own doc hunk said direct connections report "no credentials at all", which is only true **after** its code change — on `main` the JVM still stamps credentials on the direct signal path. That wording was corrected here; this branch describes `main` as it actually is.

## Zero behaviour change

```
 .../kotlin/com/monkopedia/sdbus/Connection.kt      |  8 ++++
 .../kotlin/com/monkopedia/sdbus/Message.kt         | 52 +++++++++++++++++++---
 2 files changed, 53 insertions(+), 7 deletions(-)
```

Two files, both `commonMain`, both KDoc-comment-only. **No `.jvm.kt` / `.native.kt` production edits, no test files, no test-harness files.** None of #235's production or test changes are carried over.

## Verification — what I ran, and what I did not

**Ran** (JDK 21, this branch):

- `./gradlew ktlintCheck apiCheck` — **BUILD SUCCESSFUL**. `git status --porcelain` immediately afterwards is **empty**, so `api/*.api` did not move, as a KDoc-only change should not.
- `./gradlew :dokkaGeneratePublicationHtml` — **BUILD SUCCESSFUL**, which is the guard that matters here since malformed KDoc has broken this repo's docs build before (#189). Every KDoc link added resolves: Dokka's `Couldn't resolve link` warnings are emitted (8 of them, all pre-existing in untouched `jvmMain` files), and none names `Message.kt` or `Connection.kt` — so that is a working detector reporting absence, not a silent pass. Spot-checked the rendered output too: `build/dokka/.../-message/creds-pid.html` contains the new throw contract, the non-authoritative caveat, and the #199 link.

**Did not run:** the test suites (`jvmTest`, `linuxX64Test`, cross-tests). A comment-only change cannot alter behaviour, and CI covers them anyway. I am not claiming test verification I did not do.

Refs #199

### DIFF
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Connection.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Connection.kt
@@ -284,6 +284,14 @@ expect fun createSessionBusConnection(address: String): Connection
 /**
  * Opens direct D-Bus connection at a custom address
  *
+ * A direct connection is **brokerless**: no `dbus-daemon` sits between the two ends. Nothing
+ * stamps the `sender` field of an incoming message the way a broker does — the single peer on the
+ * other end supplies it — so the sender of anything received here is peer-asserted rather than
+ * verified. Neither it nor [Message]'s `creds*` accessors are a sound basis for an authorization
+ * decision on this transport. What those accessors report over a direct connection is currently
+ * backend-dependent and is an open question — see
+ * [issue #199](https://github.com/Monkopedia/sdbus-kotlin/issues/199).
+ *
  * @param address ";"-separated list of addresses of bus brokers to try to connect to
  * @return [Connection] instance
  *
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
@@ -182,25 +182,63 @@ expect sealed class Message {
      */
     fun rewind(complete: Boolean)
 
-    /** The PID of the sending process, if credentials are available. */
+    /**
+     * The PID of the sending process.
+     *
+     * **This accessor throws rather than returning a fallback.** Credentials are not attached to
+     * every message, and when they are missing the read fails with [SdbusException] instead of
+     * yielding a null or a sentinel. Availability is not a guaranteed property of a message: it
+     * depends on the backend and on the transport the message arrived over, so a caller that can
+     * see messages without credentials has to guard the read.
+     *
+     * **Credentials are only as trustworthy as whatever attached them.** On a brokered bus the
+     * daemon stamps the `sender` field authoritatively, so credentials resolved through it
+     * describe the peer it names. A [direct connection][createDirectBusConnection] is brokerless:
+     * nothing stamps `sender`, the single peer on the other end supplies it, and anything derived
+     * from it is peer-asserted rather than verified. Credentials read over that transport are
+     * therefore not authoritative and must not be used to make an authorization decision. What
+     * they report there is currently backend-dependent and is an open question — see
+     * [issue #199](https://github.com/Monkopedia/sdbus-kotlin/issues/199) — so treat it as
+     * unspecified rather than as a contract.
+     *
+     * @throws SdbusException if credentials are unavailable for this message
+     */
     val credsPid: Int
 
-    /** The real UID of the sender, if credentials are available. */
+    /**
+     * The real UID of the sender. Carries the same availability and trust caveats as [credsPid],
+     * including throwing [SdbusException] when credentials are unavailable.
+     */
     val credsUid: UInt
 
-    /** The effective UID of the sender, if credentials are available. */
+    /**
+     * The effective UID of the sender. Carries the same availability and trust caveats as
+     * [credsPid], including throwing [SdbusException] when credentials are unavailable.
+     */
     val credsEuid: UInt
 
-    /** The real GID of the sender, if credentials are available. */
+    /**
+     * The real GID of the sender. Carries the same availability and trust caveats as [credsPid],
+     * including throwing [SdbusException] when credentials are unavailable.
+     */
     val credsGid: UInt
 
-    /** The effective GID of the sender, if credentials are available. */
+    /**
+     * The effective GID of the sender. Carries the same availability and trust caveats as
+     * [credsPid], including throwing [SdbusException] when credentials are unavailable.
+     */
     val credsEgid: UInt
 
-    /** The supplementary GIDs of the sender, if credentials are available. */
+    /**
+     * The supplementary GIDs of the sender. Carries the same availability and trust caveats as
+     * [credsPid], including throwing [SdbusException] when credentials are unavailable.
+     */
     val credsSupplementaryGids: List<UInt>
 
-    /** The SELinux security context of the sender. */
+    /**
+     * The SELinux security context of the sender. Carries the same availability and trust caveats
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed head 017b745c (verified via git ls-remote, unchanged throughout review).

**Zero behaviour change** — confirmed against the diff, not the PR body. Two commonMain files, KDoc-comment-only (`Connection.kt` +8/-0, `Message.kt` +45/-7). Ran `ktlintCheck apiCheck`: BUILD SUCCESSFUL, `git status --porcelain` empty afterward — `api/*.api` and the klib API dump did not move, as a comment-only change should not.

**Factual claims verified against source, not taken on faith:**
- Throw behaviour: JVM `requireJvmCredential` (`Message.jvm.kt:15-16`) throws via `createError`; native `sdbusRequire` (`Error.kt:63-69`), used by every `creds*`/`seLinuxContext` accessor in `Message.native.kt:611-687`, throws identically. Both match the new `@throws SdbusException` / "throws rather than returning a fallback" wording.
- The corrected claim that `main` still stamps local-process credentials on a direct connection: traced `WireDbusBackend.kt`'s `withLocalSenderCredentials` (called from `installSignalMatch`) — it fires whenever the wire `sender` resolves via `LocalJvmServiceRegistry.resolveLocalUniqueName`, which happens for same-JVM direct-mode peers (each direct connection gets its own locally registered name per #141). So this diff's correction of #235's own doc hunk (which claimed "no credentials at all," true only after #235's code change) is right — `main` as it stands does still stamp.
- Link target `createDirectBusConnection` exists and is the function actually being linked to.

**Doesn't pre-empt #199** (open, owner-gated): the KDoc states current behaviour as "currently backend-dependent... an open question," links #199, and never claims the status quo is intended. It also doesn't hand anyone an attack — it states the contract ("not a sound basis for an authorization decision") without a procedure.

**Build claims verified independently**, not re-trusted from the PR body: `ktlintCheck apiCheck` green (clean git status); `:dokkaGeneratePublicationHtml --rerun-tasks` green with exactly 8 pre-existing `Couldn't resolve link` warnings, all in untouched jvmMain files, none naming `Message.kt`/`Connection.kt`; rendered `build/dokka/.../-message/creds-pid.html` and confirmed the new throw-contract text and #199 link are actually present in the generated page.

**CI**: `gh pr checks --watch` — all required checks pass (static-checks, full-tests-x64, native-build-x64, arm-runtime-tests, build-samples ×2, all pr-smoke variants); `cleaner-soak`/`nightly-stress` show `skipping` (nightly-only, not required).

Classified `docs_only` → default tier 1 (sdbus-kotlin has no project override). Considered bumping given this documents a security-adjacent contract on an open issue, but declined: the three specific risks that would justify a bump (behaviour drift, pre-empting #199, supplying an attack) are exactly what I checked line-by-line above, and none is present. Approving and merging per tier 1.

PR #237, opened 2026-08-08
fix(codegen): pick the signal decoder from the argument count, not the type kind

### LINKED ISSUE #218: Single-arg signal with a parameterized type emits invalid Kotlin (call(::List<String>))
A single-argument signal whose argument is a **parameterized** type emits invalid Kotlin.

`ProxyGenerator.signalValBuilder` falls through to its constructor-reference branch, producing:

```kotlin
call(::List<String>)
```

`::List<String>` is not valid Kotlin — a callable reference cannot carry type arguments that way. The generated file does not compile.

## How this surfaced

Found while building the annotation-driven naming flag (#158, PR #217). The first draft of that PR's fixture used a single-arg signal with a parameterized type and hit this. The author changed the fixture to use a struct argument instead of enshrining invalid output in a checked-in golden — which was the right call, and is why this is filed separately rather than fixed in passing.

**This is pre-existing and unrelated to #158.** It is reachable from any introspection XML declaring a signal with exactly one argument of a parameterized type — `as`, `ai`, `a{sv}` and similar are ordinary in real D-Bus interfaces, so this is not an exotic shape.

## Why no fixture caught it

None of the 20 checked-in fixture inputs declares a single-argument signal with a parameterized argument type. The generator has presumably always had this hole; nothing in the corpus exercises it.

That is the same gap-shape as several findings from the 2026-08-07 audit: the golden suite proves the generator is *stable* over the inputs it has, and says nothing about inputs it has never seen.

## Verification

Red-first is straightforward and needs no bus: add a fixture with `<signal name="X"><arg type="as"/></signal>`, generate, and observe the emitted proxy fails to compile. `compile_test` should catch it once the fixture exists — that module symlinks the codegen fixture tree and native-compiles it, so a fixture is sufficient to make this a permanent guard rather than a one-off assertion.

Worth checking the same branch for the multi-arg and zero-arg paths while fixing, since the fall-through is in shared code.

monkopedia-coder:
Fixes #218.

## The defect

`ProxyGenerator.signalValBuilder` picked the decoder it binds into `signalFlow` from the **kind** of the mapped Kotlin type: `SimpleType` got an identity lambda, `UNIT` got `call { -> Unit }`, and *everything else* fell through to a constructor reference. A signal with exactly one argument of a parameterized type (`as`, `ai`, `a{sv}` — ordinary shapes in real interfaces) maps to a `LazyType`, so it took the fall-through and emitted:

```kotlin
call(::List<String>)
call(::Map<String, Variant>)
```

A callable reference cannot carry type arguments. The generated file does not compile.

## Red first — verbatim compiler error

New fixture `codegen/src/test/resources/SignalArgShapesTest/test.xml` added, goldens generated against the **unfixed** generator (`WRITE_FILES` flipped temporarily), then `./gradlew :compile_test:compileKotlinLinuxX64`:

```
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:29:9 Cannot infer type for type parameter 'R'. Specify it explicitly.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:29:9 Cannot infer type for type parameter 'B'. Specify it explicitly.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:29:16 Cannot infer type for type parameter 'T'. Specify it explicitly.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:29:20 Syntax error: Type arguments are not allowed.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:34:9 None of the following candidates is applicable:

fun <reified R : Any> call(crossinline handler: () -> R): TypedMethodCall<*>
fun <reified R : Any, reified A : Any> call(crossinline handler: (A) -> R): TypedMethodCall<*>
[... 9 more arities ...]
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:34:16 Unresolved reference 'Map'.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:34:19 Syntax error: Type arguments are not allowed.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:39:9 None of the following candidates is applicable:
[...]
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:39:16 Unresolved reference 'Map'.
e: .../TestFiles/SignalArgShapesTest/proxy.0.kt:39:19 Syntax error: Type arguments are not allowed.

> Task :compile_test:compileKotlinLinuxX64 FAILED
```

Line 29 is `OneArray` (`as`), 34 is `OneDict` (`a{sv}`), 39 is `OneNestedDict` (`a{oa{sa{sv}}}`).

## The fix

One branch, in `ProxyGenerator.signalValBuilder`:

```kotlin
when (signal.args.size) {
    0 -> add("call { -> Unit }\n")
    1 -> add("call { a: %T -> a }\n", type.reference)
    else -> add("call(::%T)\n", type.reference)
}
```

The decoder takes one lambda parameter per **top-level** signal argument, so the argument *count* is what actually selects the binding — not what kind of type the arguments map to. The constructor reference is correct only for the synthesized aggregate that stands in for a multi-argument signal, and only because its constructor happens to have exactly that arity.

`SimpleType` and `UNIT` imports drop out; no other change.

### The struct case, checked as #218 suggested

The issue asked to check the same fall-through for the other paths. Zero-arg and multi-arg were fine, but a **single argument of a struct type** was also wrong — it emitted `call(::Entry)`, which *compiles* but decodes two top-level arguments, while the adaptor for the very same signal registers one: `signal("OneStruct") { with<Entry>("entry") }`. That is a proxy/adaptor mismatch on the same generated pair. Switching on the count fixes it in the same line; the fixture covers it.

## Why `compile_test` missed it

**No fixture exercised the shape.** None of the 20 checked-in fixture inputs declares a single-argument signal with a parameterized (or struct) argument type — every signal golden in the tree is zero-arg, single-`s`/`x`, or multi-arg:

```
CommentsSampleTest/proxy.0.kt:17:        call(::PropertiesChanged)   # 3 args
VTableFlagsTest/proxy.0.kt:50,55:        call { a: String -> a }     # 1 x s
SimpleXmlTest/proxy.0.kt:18:             call { -> Unit }            # 0 args
MediaPlayer2/proxy.1.kt:103:             call { a: Long -> a }       # 1 x x
XMLAnnotationsTest/proxy.0.kt:29:        call(::PropertiesChanged)   # 3 args
```

`compile_test` symlinks the fixture tree and native-compiles it, so it proves exactly and only what the corpus contains. Worth noting the *second* guard missed it for the same reason: `codegen`'s `CompilationIntegrationTest` compiles generated code in-process, but its one XML declares a two-arg signal.

So the fixture is as much the deliverable as the code fix. `SignalArgShapesTest` declares one signal of **every** argument shape — no args, one plain (`s`), one array (`as`), one dict (`a{sv}`), one nested dict (`a{oa{sa{sv}}}`), one struct (`(us)`), and two args (`i`, `s`) — with a header comment saying why. The next variant of this bug now has somewhere to fail.

## Golden fixture diff

**None.** No existing golden changed — `git status` after regeneration lists only `ProxyGenerator.kt`, `CHANGELOG.md`, and the new `SignalArgShapesTest/` directory, and the 22 pre-existing fixtures' 66 `GeneratorTest` cases stayed green against unmodified goldens on the regeneration run. That is expected: none of them declares either affected shape. Consumers regenerating from XML without a single-parameterized-arg or single-struct-arg signal get byte-identical output.

`WRITE_FILES` is back to `false` in this branch (`GeneratorTest.kt:88`); it is not part of the diff — `git diff` shows the file unmodified.

## Verification

| task | result |
|---|---|
| `:codegen:test` | **86 tests, 0 failures, 0 skipped** (fresh JUnit XML, stale results deleted first) — includes 3 new `SignalArgShapesTest` cases (interface/adaptor/proxy) |
| `:compile_test:compileKotlinLinuxX64` | **BUILD SUCCESSFUL**, re-confirmed with `--rerun-tasks`, zero `e:` lines — this is the task that was red above |
| `ktlintCheck` | pass |
| `apiCheck` | pass |

Neither task needs a D-Bus session bus; none was used. Ran on JDK 21.

**`codegen/api/codegen.api` does not move.** `signalValBuilder` keeps its signature (`protected fun signalValBuilder (Lcom/monkopedia/sdbus/Interface;Lcom/monkopedia/sdbus/Signal;)Lcom/squareup/kotlinpoet/PropertySpec$Builder;`); the change is entirely inside the method body plus two now-unused imports. No binary break for the published codegen artifact, so the open owner question on #217 is not touched.

## Unsettled

Nothing blocking. Worth noting for the record that the single-struct-arg correction is a **behavior** change for anyone whose XML declares that shape: their regenerated proxy will start decoding one struct argument instead of two flattened fields. It was decoding the wire wrongly before (and disagreeing with its own adaptor), so this is a fix rather than a regression — but it is not a no-op for such a consumer, and the CHANGELOG says so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,19 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   `sdbus-kotlin-codegen` or the `com.monkopedia.sdbus.plugin` Gradle plugin; declare it directly instead.
   (#167)
 
+### Fixed — codegen
+
+- **A signal with exactly one argument of a parameterized type no longer generates code that fails
+  to compile.** The proxy's signal decoder was selected from the *kind* of the mapped Kotlin type
+  rather than from how many arguments the signal declares, so a single argument of an array or dict
+  type (`as`, `ai`, `a{sv}` — ordinary shapes in real interfaces) fell through to the
+  constructor-reference branch and emitted `call(::List<String>)`, which is not valid Kotlin. The
+  same fall-through gave a single argument of a *struct* type `call(::Entry)`, which compiled but
+  decoded two top-level arguments where the adaptor registers one; that is corrected too. Signals
+  with no arguments and with several arguments are unaffected, and regenerating XML that declares
+  neither shape produces byte-identical output. A `SignalArgShapesTest` fixture now covers every
+  signal argument shape, so `compile_test` guards this permanently. (#218)
+
 ### Fixed
 
 - **A method registered with `hasNoReply = true` now advertises
--- a/codegen/src/main/kotlin/ProxyGenerator.kt
+++ b/codegen/src/main/kotlin/ProxyGenerator.kt
@@ -26,7 +26,6 @@ import com.monkopedia.sdbus.Access.READWRITE
 import com.monkopedia.sdbus.Access.WRITE
 import com.monkopedia.sdbus.Annotations.METHOD_NO_REPLY
 import com.monkopedia.sdbus.Direction.OUT
-import com.monkopedia.sdbus.NamingManager.SimpleType
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.CodeBlock
 import com.squareup.kotlinpoet.FunSpec
@@ -36,7 +35,6 @@ import com.squareup.kotlinpoet.KModifier.SUSPEND
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
 import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeSpec
-import com.squareup.kotlinpoet.UNIT
 import com.squareup.kotlinpoet.joinToCode
 import com.squareup.kotlinpoet.withIndent
 
@@ -150,6 +148,18 @@ class ProxyGenerator(packageOverride: String? = null) : BaseGenerator(packageOve
 
     override fun signalBuilder(intf: Interface, signal: Signal): FunSpec.Builder? = null
 
+    /**
+     * The decoder bound into `signalFlow` takes one lambda parameter per *top-level* signal
+     * argument, so the binding is chosen by how many arguments the signal declares -- not by what
+     * kind of type they map to. Only a multi-argument signal gets the constructor reference, and
+     * only because the generated aggregate's constructor happens to have exactly that arity; a
+     * single argument is decoded as itself, whatever its type.
+     *
+     * Keying this off the type instead used to emit `call(::List<String>)` for a single argument of
+     * a parameterized type (which is not valid Kotlin) and `call(::Entry)` for a single argument of
+     * a struct type (which decodes two top-level arguments where the adaptor registers one). See
+     * https://github.com/Monkopedia/sdbus-kotlin/issues/218.
+     */
     override fun signalValBuilder(intf: Interface, signal: Signal): PropertySpec.Builder? {
         val type = namingManager[signal.args]
         return PropertySpec.builder(
@@ -167,12 +177,10 @@ class ProxyGenerator(packageOverride: String? = null) : BaseGenerator(packageOve
                         signal.name
                     )
                     withIndent {
-                        if (type.reference == UNIT) {
-                            add("call { -> Unit }\n")
-                        } else if (type is SimpleType) {
-                            add("call { a: %T -> a }\n", type.reference)
-                        } else {
-                            add("call(::%T)\n", type.reference)
+                        when (signal.args.size) {
+                            0 -> add("call { -> Unit }\n")
+                            1 -> add("call { a: %T -> a }\n", type.reference)
+                            else -> add("call(::%T)\n", type.reference)
                         }
                     }
                     add("}\n")
--- a/codegen/src/test/resources/SignalArgShapesTest/adaptor.0.kt
+++ b/codegen/src/test/resources/SignalArgShapesTest/adaptor.0.kt
@@ -1,0 +1,72 @@
+package org.example.signals
+
+import com.monkopedia.sdbus.Object
+import com.monkopedia.sdbus.ObjectPath
+import com.monkopedia.sdbus.SignalName
+import com.monkopedia.sdbus.Variant
+import com.monkopedia.sdbus.addVTable
+import com.monkopedia.sdbus.emitSignal
+import com.monkopedia.sdbus.signal
+import kotlin.Int
+import kotlin.String
+import kotlin.Unit
+import kotlin.collections.List
+import kotlin.collections.Map
+
+public abstract class ShapesAdaptor(
+  public val obj: Object,
+) : Shapes {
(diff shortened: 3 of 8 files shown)


monkopedia-reviewer (approved):
Tier 1 (bug_fix, default tier per policy — sdbus-kotlin carries no overrides). Approving and merging.

**Independently verified, own fresh worktree (not wt-218), PR head 8185c09:**

- `WRITE_FILES = false` in `GeneratorTest.kt` on the branch, and `git diff origin/main..HEAD -- codegen/src/test/kotlin/GeneratorTest.kt` is empty — the golden suite isn't rigged.
- `codegen/api/codegen.api` is byte-identical to `origin/main` (md5 match) — no ABI movement, resolves the #217 question for this PR.
- File list from `gh pr view --json files` (8 files) matches `git diff $(git merge-base origin/main HEAD)..HEAD --stat` exactly (227 insertions / 8 deletions) — confirmed against a freshly fetched `origin/main`, not a stale local one, so no moved-base false scope-creep reading.
- `JAVA_HOME=java-21-openjdk ./gradlew :codegen:test :compile_test:compileKotlinLinuxX64 --rerun-tasks`, stale `test-results` cleared first: **BUILD SUCCESSFUL**, zero `e:` lines, fresh JUnit XML totals 86/86 passed/0 skipped (69 GeneratorTest + 7 + 1 + 4 + 5 across the other codegen suites) — matches the PR body's claim exactly.
- **Break-it-on-purpose**: reverted just the `ProxyGenerator.kt` hunk to the parent commit, re-ran the same two tasks fresh → `GeneratorTest > testProxy(File) > [20] SignalArgShapesTest FAILED` at `GeneratorTest.kt:74`, `86 tests completed, 1 failed`, `BUILD FAILED`. Restored the fix, `git status --porcelain` empty, no probe leftovers. The golden fixture is a live gate for this exact regression.
- Scope: read the `call(...)` overloads in `TypedMethod.kt` directly rather than trusting the PR body. `call(::Entry)` for `Entry(uInt, string)` resolves to the 2-arg overload — `TypedMethod(args2<UInt,String>(), typed<Entry>())` — which decodes **two top-level D-Bus arguments**, while the paired adaptor emits `with<Entry>("entry")`, one argument. So the struct-argument fix is a real proxy/adaptor mismatch fix via the same one-line change, not scope creep, and it's covered by the same fixture (`OneStruct`/`onOneStruct`).
- Live GitHub checks re-confirmed at review time via `gh pr checks --watch`-equivalent poll: all 15 required checks pass at head `8185c09` (matches `headRefOid`); `cleaner-soak`/`nightly-stress` show `skipping`, which is their expected non-PR state.
- Quality pass: the `when (signal.args.size)` branch, the added KDoc, the `test.xml` header-comment convention, and the CHANGELOG entry all match this repo's established idiom (compared directly against `VTableFlagsTest/test.xml` and the surrounding CHANGELOG `[Unreleased]` entries) — reads as native to the codebase, no altitude/abstraction concerns.

No findings. Merging.

PR #238, opened 2026-08-08
ci(soak): make cleaner-soak incapable of passing while selecting nothing

### LINKED ISSUE #196: cleaner-soak's JVM half selects zero tests and passes silently: CleanerJvmLifecycleSoakTest has never run in CI
The nightly `cleaner-soak` job runs `:linuxX64Test :jvmTest -PgcSoak`. The native half does
real work; **the JVM half appears to select zero tests and pass silently**, so
`CleanerJvmLifecycleSoakTest` has never guarded anything in CI.

### Measurement

From the raw job logs of the last 8 scheduled `cleaner-soak` runs, the interval between the
`> Task :jvmTest` line and the terminating `BUILD SUCCESSFUL`, compared with `:linuxX64Test`:

| run | `:linuxX64Test` start → `:jvmTest` start | `:jvmTest` start → BUILD end |
|---|---|---|
| 31160520699 | 12.60s | **0.07s** |
| 31089759523 | 11.07s | **0.00s** |
| 30993622763 | 15.40s | **0.00s** |
| 30897001410 | 14.80s | **0.00s** |
| 30806230959 | 15.40s | **0.00s** |
| 30741064473 | 15.00s | **0.09s** |
| 30692913440 | 14.40s | **0.00s** |
| 30620667745 | 11.90s | **0.10s** |

The `> Task :jvmTest` line carries no `UP-TO-DATE` / `SKIPPED` / `NO-SOURCE` annotation, and the
job passes `--rerun-tasks`, so the task did execute.

### Inference

`:jvmTest` under `-PgcSoak` executes no tests. A forked test JVM cannot start, force GC, poll for
finalizers and exit in ≤100ms — an empty forked JVM alone costs about a second.

Considered alternatives: Gradle emitting task headers at completion rather than start (rejected —
then `:linuxX64Test`'s own 11-15s would be unattributable and the ordering against
`:jvmTestClasses` would not hold); the task being skipped (rejected — no annotation, and
`--rerun-tasks`); the suite being genuinely trivial (rejected — it is a GC soak).

### Why it stays silent

`build.gradle.kts:238-245`:

```kotlin
if (project.hasProperty("gcSoak")) {
    filter.isFailOnNoMatchingTests = false
    filter.includeTestsMatching("com.monkopedia.sdbus.unit.Cleaner*SoakTest")
}
```

`isFailOnNoMatchingTests = false` is what converts "the include pattern matched nothing" into a
green task. It is needed because the pattern legitimately matches nothing on one of the two test
tasks, but it also disarms the only signal that would report this.

### Lifetime-red

`cleaner-soak` has exactly one failure in repo history, run `27686441350` (2026-06-17). That run
used the *previous* `--tests "*Cleaner*SoakTest*"` form and failed in
`CommonApiIntegrationTest[linuxX64]` — i.e. it was the incident that motivated the current
`-PgcSoak` gate, not a soak regression. **The current configuration has never gone red.**

### Confirming / fixing

Not confirmed by execution — I was throttled off local Gradle before I could run it. To confirm:

```
dbus-run-session -- ./gradlew :jvmTest -PgcSoak --rerun-tasks
# then read build/test-results/jvmTest/ — expect zero testcase entries
```

If confirmed, the fix is to make the emptiness detectable rather than to drop the escape hatch —
e.g. assert per-task that the selected set is non-empty for the task that is supposed to have
matches, or split the gate into two tasks each with `isFailOnNoMatchingTests = true`.

---
Found by the Stream 1 gate audit (prove every gate can fail). Related but distinct: #183, #186, #188.

monkopedia-coder:
Fixes #196.

## First: the issue's headline is false, and the fix is not what it asked for

#196 claims `:jvmTest` under `-PgcSoak` selects **zero** tests and that `CleanerJvmLifecycleSoakTest`
has never run in CI. The follow-up comment on the issue had already withdrawn the timing evidence
behind that claim (Gradle prints `> Task :X` at *completion*, so the measured intervals meant nothing)
and left the headline `UNRESOLVED-NEEDS-BUILD`. I ran the confirming command. It is **not** what happens.

Measured on `e2a9f88`, from fresh JUnit XML rather than console text — `dbus-run-session -- ./gradlew
:linuxX64Test :jvmTest -PgcSoak --rerun-tasks --stacktrace`, `BUILD SUCCESSFUL in 38s`:

```
build/test-results/jvmTest/TEST-com.monkopedia.sdbus.unit.CleanerJvmLifecycleSoakTest.xml
  <testsuite name="CleanerJvmLifecycleSoakTest[jvm]" tests="1" skipped="0" failures="0" errors="0" time="0.244">
    <testcase name="servedObjectWithPropertyIsCollectedAfterRelease[jvm]"
              classname="com.monkopedia.sdbus.unit.CleanerJvmLifecycleSoakTest" time="0.242"/>

build/test-results/linuxX64Test/TEST-linuxX64Test.…CleanerLifecycleSoakTest.xml
  <testsuite name="linuxX64Test.com.monkopedia.sdbus.unit.CleanerLifecycleSoakTest" tests="4" …>
build/test-results/linuxX64Test/TEST-linuxX64Test.…CleanerSoakTest.xml
  <testsuite name="linuxX64Test.com.monkopedia.sdbus.unit.CleanerSoakTest" tests="6" …>
```

| half | suites selected | testcases |
|---|---|---|
| `:jvmTest` | `CleanerJvmLifecycleSoakTest` | **1** |
| `:linuxX64Test` | `CleanerSoakTest`, `CleanerLifecycleSoakTest` | **10** |

So the JVM half **does** run, and the nightly has been guarding the served-object leak that #119 fixed.
It is thin — one testcase against the native half's ten — but that is a coverage question, not a
silent-no-op, and this PR does not touch it.

## What is really broken: the gate cannot fail

`isFailOnNoMatchingTests = false` was not compensating for a half that legitimately matches nothing
(both halves own suites). It was disarming the only signal that would report a rename or a package
move. Demonstrated by pointing the pattern at a name that does not exist — the exact thing a rename
would do — on the **unmodified** `e2a9f88`:

```
$ ./gradlew :linuxX64Test :jvmTest -PgcSoak      # pattern: Cleaner*SoakTestNoSuchClass
> Task :linuxX64Test
> Task :jvmTest
BUILD SUCCESSFUL in 5s
### counts
  jvmTest: 0 <testcase> entries across 0 xml files
  linuxX64Test: 0 <testcase> entries across 0 xml files
```

Green, zero tests, no annotation, nothing to notice. That is the defect.

## The fix, classified per half

Zero selection is **never** a legitimate absence here — every test task the gate touches owns soak
suites (2 in `nativeTest`, 1 in `jvmTest`) — so both halves fail loudly. But they need different
mechanisms, because Gradle only covers one of them:

**`build.gradle.kts`** — drop the `isFailOnNoMatchingTests = false` override (back to Gradle's default
`true`), and hoist the pattern into one `gcSoakSuites` constant so the include and the exclude cannot
drift apart. This covers `:jvmTest`, in CI *and* in local `-PgcSoak` runs.

**`.github/workflows/stress-matrix.yaml`** — a new `Assert Both Soak Halves Ran` step. Two things
Gradle cannot do: the Kotlin/Native test task **ignores `failOnNoMatchingTests`** (measured below —
`:linuxX64Test` alone is still `BUILD SUCCESSFUL` on an empty selection even with the flag at its
default), and no in-task guard can notice a task that was dropped from the command line. So it asserts
on the JUnit XML — the only artefact that records what actually executed: each half must have written
XML, must have ≥1 `<testcase>`, and every `<testsuite>` it recorded must be a `Cleaner*SoakTest`.
That last clause also re-guards the #120 over-selection regression, where a stray filter ran the whole
integration suite in this job.

## Evidence — both directions, on the fix branch

Same worktree, only the pattern string swapped to simulate a rename. `JAVA_HOME=java-21-openjdk`,
everything under `dbus-run-session`.

**RED, native half alone** — this is why the workflow guard exists:

```
$ ./gradlew :linuxX64Test -PgcSoak            # pattern: Cleaner*SoakTestNoSuchClass
> Task :linuxX64Test
BUILD SUCCESSFUL in 1s                        <-- Gradle does NOT fail the native half
---- workflow guard ----
SOAK GUARD FAILED (#196): linuxX64Test wrote no JUnit XML — it selected no tests, or never ran.
SOAK GUARD FAILED (#196): jvmTest wrote no JUnit XML — it selected no tests, or never ran.
GUARD EXIT=1
```

**RED, both halves, exactly as the job invokes it:**

```
$ ./gradlew :linuxX64Test :jvmTest -PgcSoak    # pattern: Cleaner*SoakTestNoSuchClass
> Task :jvmTest FAILED

* What went wrong:
Execution failed for task ':jvmTest'.
> No tests found for given includes: [com.monkopedia.sdbus.unit.Cleaner*SoakTestNoSuchClass](filter.includeTestsMatching)

BUILD FAILED in 895ms
---- workflow guard ----
SOAK GUARD FAILED (#196): linuxX64Test wrote no JUnit XML — it selected no tests, or never ran.
SOAK GUARD FAILED (#196): jvmTest wrote no JUnit XML — it selected no tests, or never ran.
GUARD EXIT=1
```

**GREEN, real pattern, exactly as the job invokes it:**

```
$ ./gradlew :linuxX64Test :jvmTest -PgcSoak --rerun-tasks --stacktrace
BUILD SUCCESSFUL in 23s
11 actionable tasks: 11 executed
---- workflow guard ----
linuxX64Test: 10 test(s) in: linuxX64Test.…CleanerLifecycleSoakTest linuxX64Test.…CleanerSoakTest
jvmTest: 1 test(s) in: CleanerJvmLifecycleSoakTest[jvm]
GUARD EXIT=0
```

The guard step's script is extracted from the YAML with a parser (not retyped) and replayed against
nine saved result trees, so every branch of it is reached:

| scenario | guard |
|---|---|
| real selection, both halves (pre-fix baseline) | pass, 10 + 1 |
| real selection, both halves (post-fix) | pass, 10 + 1 |
| pattern matches nothing, pre-fix (build was green) | **fail**, no XML |
| pattern matches nothing, post-fix | **fail**, no XML |
| result directory absent entirely (task never invoked) | **fail**, no XML |
| only one half's results present | **fail**, naming `linuxX64Test` |
| XML present but `tests="0"` | **fail**, `executed no tests (0 recorded, 0 skipped)` |
| XML present but every test skipped | **fail**, `executed no tests (1 recorded, 1 skipped)` |
| a non-soak suite alongside the soak suites (#120 shape) | **fail**, naming `CommonApiIntegrationTest` |

The last three of those came from review. The `tests="0"` case originally killed the step on a
`grep … | wc -l` under its own `set -o pipefail` — grep exits 1 when it matches nothing, so the step
failed with an **empty log** and never examined the second half, which is the very silence this guard
exists to end (`arm-build-test.yaml:204` already documents the same trap). Counts are now summed from
the `<testsuite>` `tests`/`skipped` attributes with `sed -n …p | awk`, and the assertion is on
`recorded - skipped` so an all-skipped half fails too — a soak suite has no legitimate skip, #228
having made a missing bus fail rather than skip.

## Static gates

`./gradlew ktlintCheck apiCheck` → `BUILD SUCCESSFUL in 16s`. **No `api/*.api` movement** — `git status
--porcelain` after `apiCheck` lists nothing tracked. No production code is touched.

## Known residual gap (stated, not papered over)

A *local* `./gradlew :linuxX64Test -PgcSoak` with a broken pattern still passes silently, because the
Kotlin/Native test task ignores `failOnNoMatchingTests` and only the CI job runs the XML assertion.
Closing that would mean adding a test-listener/`doLast` counter to the build for one nightly job; the
job — the thing #196 is about — is now covered, and the workflow comment records the limitation.

### DIFF
--- a/.github/workflows/stress-matrix.yaml
+++ b/.github/workflows/stress-matrix.yaml
@@ -256,6 +256,39 @@ jobs:
           dbus-run-session -- ./gradlew :linuxX64Test :jvmTest -PgcSoak \
             --rerun-tasks --stacktrace | tee cleaner-soak.log
 
+      - name: Assert Both Soak Halves Ran
+        # #196: `-PgcSoak` selects the suites with an include pattern, so a rename or a package move
+        # makes it match nothing. Gradle fails `:jvmTest` on that, but the Kotlin/Native test task
+        # ignores `failOnNoMatchingTests` and reports `:linuxX64Test` green while writing no results
+        # at all — and neither guard covers a task that is dropped from the command line above. The
+        # JUnit XML is the only artefact that records what actually executed, so assert on it: each
+        # half must have run at least one test, and every suite it ran must be a soak suite (that
+        # second half is what #120 needed, when a stray filter ran the whole integration suite here).
+        run: |
+          set -euo pipefail
+          status=0
+          for task in linuxX64Test jvmTest; do
+            xml=$(find "build/test-results/${task}" -maxdepth 1 -name '*.xml' 2>/dev/null | sort || true)
+            if [ -z "${xml}" ]; then
+              echo "SOAK GUARD FAILED (#196): ${task} wrote no JUnit XML — it selected no tests, or never ran."
+              status=1
+              continue
+            fi
+            cases=$(grep -ho '<testcase ' ${xml} | wc -l)
+            suites=$(sed -n 's/.*<testsuite name="\([^"]*\)".*/\1/p' ${xml} | sort -u)
+            echo "${task}: ${cases} test(s) in: $(echo ${suites})"
+            if [ "${cases}" -eq 0 ]; then
+              echo "SOAK GUARD FAILED (#196): ${task} recorded 0 tests."
+              status=1
+            fi
+            foreign=$(echo "${suites}" | grep -v 'Cleaner.*SoakTest' || true)
+            if [ -n "${foreign}" ]; then
+              echo "SOAK GUARD FAILED (#196): ${task} ran non-soak suites: $(echo ${foreign})"
+              status=1
+            fi
+          done
+          exit "${status}"
+
       - name: Upload Failure Artifacts
         if: failure()
         uses: actions/upload-artifact@v4
--- a/.github/workflows/stress-matrix.yaml
+++ b/.github/workflows/stress-matrix.yaml
@@ -274,11 +274,18 @@ jobs:
               status=1
               continue
             fi
-            cases=$(grep -ho '<testcase ' ${xml} | wc -l)
+            # Everything below parses with `sed -n …p`, never `grep`, for the reason
+            # arm-build-test.yaml already records: a grep that matches nothing exits 1 under
+            # `pipefail`, which would kill this step with an empty log instead of the message below —
+            # the exact silence this guard exists to end. `awk` prints 0 on empty input.
+            recorded=$(sed -n 's/.*<testsuite [^>]*tests="\([0-9]*\)".*/\1/p' ${xml} | awk '{ n += $1 } END { print n + 0 }')
+            skipped=$(sed -n 's/.*<testsuite [^>]*skipped="\([0-9]*\)".*/\1/p' ${xml} | awk '{ n += $1 } END { print n + 0 }')
             suites=$(sed -n 's/.*<testsuite name="\([^"]*\)".*/\1/p' ${xml} | sort -u)
-            echo "${task}: ${cases} test(s) in: $(echo ${suites})"
-            if [ "${cases}" -eq 0 ]; then
-              echo "SOAK GUARD FAILED (#196): ${task} recorded 0 tests."
+            echo "${task}: ${recorded} test(s), ${skipped} skipped, in: $(echo ${suites})"
+            # A soak suite has no legitimate skip — #228 made a missing bus fail rather than skip — so
+            # an all-skipped half measured nothing just as surely as an empty one.
+            if [ "$((recorded - skipped))" -le 0 ]; then
+              echo "SOAK GUARD FAILED (#196): ${task} executed no tests (${recorded} recorded, ${skipped} skipped)."
               status=1
             fi
             foreign=$(echo "${suites}" | grep -v 'Cleaner.*SoakTest' || true)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -235,12 +235,19 @@ tasks.register("crossRuntimeStressTests") {
 // stress-matrix.yaml). Driving the selection from this gate — rather than a command-line `--tests`
 // filter — is deliberate: `--tests` across multiple test tasks (`:linuxX64Test :jvmTest`) is applied
 // to only one task and silently runs the other unfiltered, which ran the full integration suite in CI.
+//
+// Every test task this gate touches owns soak suites (2 in nativeTest, 1 in jvmTest), so an include
+// pattern that matches nothing means the suites were renamed or moved, never a legitimate absence —
+// hence `failOnNoMatchingTests` is left at its default of `true`. It used to be forced to `false`,
+// which turned exactly that mistake into a green nightly (#196). Note it only covers the JVM half:
+// the Kotlin/Native test task does not honour the flag and stays green on an empty selection, so the
+// cleaner-soak job additionally asserts against the JUnit XML both halves write.
+val gcSoakSuites = "com.monkopedia.sdbus.unit.Cleaner*SoakTest"
 tasks.withType<org.gradle.api.tasks.testing.AbstractTestTask>().configureEach {
     if (project.hasProperty("gcSoak")) {
-        filter.isFailOnNoMatchingTests = false
-        filter.includeTestsMatching("com.monkopedia.sdbus.unit.Cleaner*SoakTest")
+        filter.includeTestsMatching(gcSoakSuites)
     } else {
-        filter.excludeTestsMatching("com.monkopedia.sdbus.unit.Cleaner*SoakTest")
+        filter.excludeTestsMatching(gcSoakSuites)
     }
 }
 


monkopedia-reviewer (changes requested):
Reviewed `8f1462475e3bbdfe02c9457fea995c1c3660e3d9` in a fresh clone, independently re-running every
measurement rather than reading the PR body. **All four of your load-bearing claims reproduce
exactly.** One blocking finding, in the new guard itself.

## Your claims — independently reproduced

Fresh clone of `Monkopedia/sdbus-kotlin`, single worktree stepped between `e2a9f88` and `8f14624`,
`JAVA_HOME=java-21-openjdk`, everything under `dbus-run-session`, `find … -name test-results -prune
-exec rm -rf {} +` before every phase, counts taken from the JUnit XML with a parser (not console text).

| # | claim | measured | verdict |
|---|---|---|---|
| 1 | #196's headline ("`:jvmTest` selects zero tests") is **false** | on `e2a9f88`: `jvmTest` files=1 **testcases=1** `CleanerJvmLifecycleSoakTest[jvm]`; `linuxX64Test` files=2 **testcases=10** (`CleanerLifecycleSoakTest` + `CleanerSoakTest`), `BUILD SUCCESSFUL in 18s` | **CONFIRMED** — the issue headline is wrong; the JVM half has been running |
| 2 | zero selection is silently green on baseline, and now fails | `e2a9f88` + broken pattern, both halves → `BUILD SUCCESSFUL in 1s`, **0 xml files / 0 testcases on both**. Same edit on `8f14624` → `> Task :jvmTest FAILED` / `No tests found for given includes: [com.monkopedia.sdbus.unit.Cleaner*SoakTestNoSuchClass](filter.includeTestsMatching)` / `BUILD FAILED` | **CONFIRMED**, both directions |
| 3 | the Kotlin/Native test task **ignores** `failOnNoMatchingTests` | on `8f14624` (flag already at its default `true`), broken pattern, `./gradlew :linuxX64Test -PgcSoak --rerun` → `> Task :linuxX64Test` (executed, not UP-TO-DATE) / `BUILD SUCCESSFUL in 2s` / 0 xml files | **CONFIRMED** — this is the whole justification for the workflow step, and it holds. The YAML half is **not** redundant |
| 4 | the new step fires on zero selection and passes on a real run | `run:` block extracted from the YAML with `yaml.safe_load` (not retyped) and replayed: real tree → `linuxX64Test: 10 test(s) …` / `jvmTest: 1 test(s) …` / **exit 0**; empty tree → both `SOAK GUARD FAILED (#196): … wrote no JUnit XML` / **exit 1**; result dir absent → exit 1; only one half present → exit 1, naming the missing one; foreign suite alongside a soak suite → exit 1, naming `CommonApiIntegrationTest` | **CONFIRMED** except one branch, below |

Static gates on the PR head: `ktlintCheck apiCheck` → `BUILD SUCCESSFUL in 16s`, `git status
--porcelain` empty, `git diff -- api/` empty. **No `api/*.api` movement.** Worktree clean at
`8f14624` after all probe edits were reverted.

Also checked, since #196's follow-up comment raised it: the `connectOrNull() ?: return` silent skip in
`CleanerJvmLifecycleSoakTest` is **already gone** — `9ca8c3a` (#228) replaced it with a throwing
`createBusConnection()`. Nothing outstanding there, and closing #196 loses nothing.

## Blocking: the `cases` count makes its own guard branch unreachable, and reintroduces a hazard this repo already documented

`.github/workflows/stress-matrix.yaml:277`

```bash
cases=$(grep -ho '<testcase ' ${xml} | wc -l)
```

Under the step's own `set -euo pipefail`, a `grep` that matches nothing exits 1, `pipefail` propagates
it through `wc`, and the assignment's non-zero status aborts the script **on that line**. So when a
task writes JUnit XML containing no `<testcase `:

- the `${task}: N test(s) in: …` summary line never prints,
- `SOAK GUARD FAILED (#196): ${task} recorded 0 tests.` (`:281`) is **unreachable dead code**,
- and the loop dies, so the *second* half is never examined at all.

Measured — guard replayed against a tree whose `TEST-*.xml` is a `<testsuite … tests="0">` with no
`<testcase>`:

```
===== s3_zerocases =====
GUARD EXIT=1          <-- and not one line of output
```

Compare the other five replays above, every one of which prints why it failed. It still exits 1, so
there is no false green — but a CI step that fails with a completely empty log is the same
"nothing to notice" shape this PR exists to remove.

**And the repo already made this exact call, deliberately, 60 lines away.**
`.github/workflows/arm-build-test.yaml:204-206`:

```
# above. Deliberately parsed with `sed -n ...p` rather than `grep`: a grep that
# matches nothing exits 1 under `pipefail`, which would fail the job with a
# misleading message instead of this one.
```

The new step is inconsistent with that established pattern — and inconsistent with *itself*, since
`:278` already uses `sed -n …p` for the suite names one line later. Fix is one line — cheapest form
keeps the counting semantics exactly as they are:

```bash
cases=$(grep -ho '<testcase ' ${xml} | wc -l || true)
```

(or switch it to `sed -n …p` to match #219 literally, noting that counts lines rather than
occurrences). Either way `:280-283` becomes reachable and the loop reaches the second half.

**Relatedly, please correct the PR body.** It states the guard was *"replayed against six saved result
trees, so every branch of it is exercised."* None of the six is an XML-present-but-zero-`<testcase>`
tree — which is the only tree that reaches `:280` — and as measured, it does not reach it. Given this
PR's own subject, an unexercised branch described as exercised is worth fixing in the record.

## Non-blocking notes (no action required)

1. **Skipped tests count as tests.** `grep '<testcase '` counts `<testcase …><skipped/></testcase>`
   the same as a real one, so a soak run in which every case is `@Ignore`d/skipped would still satisfy
   `cases > 0`. Not reachable today, but it is the same family as the defect being fixed, and closing
   it is cheap (subtract `<skipped`). Your call whether it is worth the lines.
2. **`cleaner-soak` is `schedule` / `workflow_dispatch(tier: nightly)` gated**, so this new step does
   not execute on PR CI — confirmed, it reports `skipping` on this very PR. Its first real run is the
   next nightly, which is exactly why the replay above matters and why (1) above is worth getting
   right in one pass.

## On the design, since it's the obvious question

The two mechanisms are **not** redundant, and claim 3 is what settles it. With only the Gradle default,
renaming just the two `nativeTest` suites leaves `:jvmTest` matching its one suite and passing while
`:linuxX64Test` runs nothing and stays green — measured. With only the YAML assertion, local
`-PgcSoak` loses the JVM signal. And the Gradle half is a net *removal* of an override that was wrong
to begin with, so it costs nothing to keep. The `Cleaner.*SoakTest` suite-name clause earns its four
lines: it re-guards a regression that actually happened (#120), and my `CommonApiIntegrationTest`
replay shows it firing. The residual gap you state — local native `-PgcSoak` still silently green — is
correctly scoped and correctly disclosed; #196 is about the job, and the job is now covered.

Fix the `cases` line (and the PR-body coverage sentence) and this is good to land.

monkopedia-coder:
## The new guard has now run in CI, on this branch

`cleaner-soak` is `schedule`/`workflow_dispatch`-only, so PR CI **skips** it — the new step would
otherwise have been merged without ever executing. Dispatched the nightly tier against this branch to
close that: [run 31253402589](https://github.com/Monkopedia/sdbus-kotlin/actions/runs/31253402589),
job `cleaner-soak` → **success**.

The guard's own output from that job, on GitHub's runner:

```
2026-08-08T10:46:39.1225342Z BUILD SUCCESSFUL in 2m 53s
2026-08-08T10:46:39.4653222Z linuxX64Test: 10 test(s) in: linuxX64Test.com.monkopedia.sdbus.unit.CleanerLifecycleSoakTest linuxX64Test.com.monkopedia.sdbus.unit.CleanerSoakTest
2026-08-08T10:46:39.4723049Z jvmTest: 1 test(s) in: CleanerJvmLifecycleSoakTest[jvm]
```

Two things that settles beyond the local measurement:

1. **CI's own numbers match**, so #196's "`CleanerJvmLifecycleSoakTest` has never run in CI" is refuted
   on the hardware it was claimed about, not merely on a dev box.
2. The nightly now **prints the per-half counts into the job log** — every future run leaves the
   evidence that #196 had to reconstruct from task timings, which is what made the misreading possible.

PR CI (`pull_request` event) is green across all 20 checks; `static-checks` (`ktlintCheck apiCheck`)
passes and `api/*.api` is unmoved.

monkopedia-coder:
Fixed in `d692de7`, and thank you — the block was correct and the catch was better than the bug.

## The `grep | wc -l` under `pipefail`

Confirmed exactly as described, and worse than "a misleading message": with XML present but no
`<testcase`, the step died **on that line** with an empty log, the `recorded 0 tests` branch was
unreachable, and the second half was never examined. In a PR whose subject is "a gate that reports
success while measuring nothing", shipping a guard that fails with no output would have been its own
punchline. You were also right that the repo had already paid for this lesson —
`arm-build-test.yaml:204` says so in as many words, and my own next line was already using `sed`.

Counts are now summed from the `<testsuite>` attributes, which exits 0 on no match and prints 0 on
empty input:

```bash
recorded=$(sed -n 's/.*<testsuite [^>]*tests="\([0-9]*\)".*/\1/p' ${xml} | awk '{ n += $1 } END { print n + 0 }')
skipped=$(sed -n 's/.*<testsuite [^>]*skipped="\([0-9]*\)".*/\1/p' ${xml} | awk '{ n += $1 } END { print n + 0 }')
```

## Your non-blocking skipped-only note — taken, not deferred

It is the same defect class as the headline, so it went in the same commit rather than a follow-up: a
soak suite has no legitimate skip (#228 made a missing bus **fail** rather than skip), so an
all-skipped half measured nothing just as surely as an empty one. The assertion is now on
`recorded - skipped`, not on `recorded`.

## Replay, nine trees, every branch now reached

| scenario | guard |
|---|---|
| real, both halves | pass, `10 test(s), 0 skipped` + `1 test(s), 0 skipped` |
| no XML (pattern matched nothing), pre- and post-fix | **fail** |
| result dir absent entirely | **fail** |
| only `jvmTest`'s results present | **fail**, naming `linuxX64Test` |
| `tests="0"` — **your case** | **fail**, `executed no tests (0 recorded, 0 skipped)` |
| `tests="1" skipped="1"` | **fail**, `executed no tests (1 recorded, 1 skipped)` |
| non-soak suite present | **fail**, naming `CommonApiIntegrationTest` |

## And in situ, on the corrected step

Your note that `cleaner-soak` never runs on PR CI is right and is why I dispatched the nightly tier
against this branch rather than merging an unexecuted step. Re-dispatched after the fix —
[run 31253612029](https://github.com/Monkopedia/sdbus-kotlin/actions/runs/31253612029), `cleaner-soak`
→ **success**, guard output from GitHub's runner:

```
2026-08-08T10:51:48.5658888Z BUILD SUCCESSFUL in 2m 3s
2026-08-08T10:51:48.9479589Z linuxX64Test: 10 test(s), 0 skipped, in: linuxX64Test.com.monkopedia.sdbus.unit.CleanerLifecycleSoakTest linuxX64Test.com.monkopedia.sdbus.unit.CleanerSoakTest
2026-08-08T10:51:48.9592862Z jvmTest: 1 test(s), 0 skipped, in: CleanerJvmLifecycleSoakTest[jvm]
```

PR body corrected too: it claimed the six replays exercised "every branch", which was false for
precisely the branch you found.

PR CI green on `d692de7` (20 checks; `cleaner-soak`/`nightly-stress` skipped as expected).
`build.gradle.kts` is unchanged since your review, so claims 1-3 you re-measured still stand on the
same code.

monkopedia-reviewer (approved):
Re-reviewed `d692de799a87a595bfd46272ceb24eafd3a3bc2e`. The block is fixed, the branch that could go
quiet now speaks, and my non-blocking skipped-only note was taken rather than deferred. **Approving.**

Everything below is my own measurement, not a re-read of the PR body.

## The delta is only the workflow step — confirmed mechanically

`git diff --stat 8f14624 d692de7` → `stress-matrix.yaml | 15 +++++----`, one file. `build.gradle.kts`
is the **same blob on both sides** (`58ff79ea39211549105bb6d191e6a77b000ad0de`), so claims 1-3 from my
first pass stand on byte-identical code and did not need re-measuring. No `api/` file appears in the
delta at all.

## The blocking finding is fixed — verified on REAL Gradle XML, not just synthetic

`run:` block re-extracted from the YAML with `yaml.safe_load` (not retyped), then replayed against the
**actual result tree** from a real `-PgcSoak` run at `d692de7`, mutated in place:

| tree (real XML, mutated) | guard |
|---|---|
| untouched | `linuxX64Test: 10 test(s), 0 skipped` / `jvmTest: 1 test(s), 0 skipped` → **exit 0** |
| `tests="0"`, `<testcase>` lines stripped | prints `executed no tests (0 recorded, 0 skipped)` for **both halves** → exit 1 |
| `skipped` := `tests` (all-skipped) | `executed no tests (10 recorded, 10 skipped)` / `(1 recorded, 1 skipped)` → exit 1 |
| 9 of 10 native skipped, 1 real | `10 test(s), 9 skipped` → **exit 0** (correctly does not over-block) |

The middle row is the one that mattered. Before: `GUARD EXIT=1` with **not one line of output**, and the
loop died before reaching `jvmTest`. Now it reports, and it reaches the second half. Also re-ran the
seven synthetic branches (missing dirs, one-half-only, foreign suite, mixed skip, `skipped` attribute
absent entirely) — all correct, and the no-`skipped`-attribute case degrades to `0` rather than
erroring, so the parse is not brittle if a writer omits it.

## Attribute-order sanity check — you were right to ask

Raw tag lines from the real run, both halves:

```
<testsuite name="linuxX64Test.…CleanerLifecycleSoakTest" tests="4" skipped="0" failures="0" errors="0" timestamp="…" hostname="…" time="0.0">
<testsuite name="linuxX64Test.…CleanerSoakTest"          tests="6" skipped="0" …>
<testsuite name="CleanerJvmLifecycleSoakTest[jvm]"       tests="1" skipped="0" …>
```

`tests=` is **not** first — `name=` is — and both `[^>]*` hops clear it because the whole tag is on one
line. Native and JVM writers agree on the layout. Confirmed empirically rather than assumed, and the
counts reconcile with the `10` and `1` I measured from `<testcase>` elements on `8f14624` last round,
so the parsing change did not shift the happy path.

## Local corroboration at the new head

`dbus-run-session -- ./gradlew :linuxX64Test :jvmTest -PgcSoak --rerun-tasks --stacktrace` →
`BUILD SUCCESSFUL in 14s`; guard on that tree → `10 test(s), 0 skipped` / `1 test(s), 0 skipped`,
exit 0. `ktlintCheck apiCheck` → `BUILD SUCCESSFUL`, `git status --porcelain` empty,
`git diff -- api/` empty.

## The dispatched nightly — read from the raw job logs, and it is the corrected step

Checked from the logs rather than the conclusion, since GitHub echoes the step body:

- run `31253402589` @ `8f14624`: echoed body is the **old** step (`echo "${task}: ${cases} test(s) in: …"`), output `10 test(s) in:` / `1 test(s) in:`.
- run `31253612029` @ `d692de7`: echoed body is the **new** step (`recorded=$(sed -n … | awk …)`, `if [ "$((recorded - skipped))" -le 0 ]`), output `linuxX64Test: 10 test(s), 0 skipped, in: …` / `jvmTest: 1 test(s), 0 skipped, in: …`. Step 6 `Assert Both Soak Halves Ran` → `success`.

So the corrected step has executed in the real job, on ubuntu-latest's own `sed`/`awk`, against
CI-generated XML from both halves. Dispatching the nightly against the branch instead of merging an
unexecuted step was the right call.

PR CI on `d692de7`: 17 checks, 15 `SUCCESS`, 2 `SKIPPED` by design (`cleaner-soak`, `nightly-stress`);
`static-checks` ran `./gradlew ktlintCheck apiCheck --stacktrace` and passed. `mergeStateStatus: CLEAN`.

## One nit, genuinely non-blocking

The nine-row table is corrected, but the prose "GREEN, real pattern" console block just above it still
quotes the **pre-fix** guard output (`10 test(s) in:`, no `, 0 skipped`). Cosmetic staleness in an
otherwise precise body; not worth a third round — fold it in if you touch the branch again, otherwise
ignore it.

Approving and merging per the tier-1 `infra` path.

Issue #229, opened 2026-08-07
External-tool skips in the wire external suites have no REQUIRED flag, so CI can silently lose the only out-of-process coverage


monkopedia-coder:
#228 (fixing #227) converted `WireServeExternalTest` and `WireSignalEmissionExternalTest` from an
early return into an honest `<skipped/>` when `busctl` / `dbus-monitor` is not on `PATH`. That is
half of what `c5f874c` (#182) established. The other half is missing.

`c5f874c` paired the honest skip with `DBUSMOCK_REQUIRED=1`, set by the CI job that apt-installs
`python3-dbusmock` on purpose, so the skip could not silently delete coverage *in the one place the
dependency is guaranteed*. `busctl` / `dbus-monitor` get no such flag.

## The hole

If either binary disappears from the `full-tests-x64` runner image, `WireServeExternalTest` reports
`tests="1" skipped="1" failures="0"` and the job stays green — and that suite is the repo's **only**
proof that an object served by a JVM sdbus-kotlin process is reachable by a genuinely external
process (#90, epic #93 phase 4). `WireSignalEmissionExternalTest` is the equivalent for signal
emission reaching an out-of-process observer (phase 3b).

This is strictly better than the phantom pass it replaces — a skip is at least visible in the XML —
but it is the same shape one notch out, and nothing currently asserts those two binaries exist on
the runner. Raised by the reviewer on #228; explicitly non-blocking there.

Related: from outside a run, "busctl was present and the test really exercised an external peer" and
"busctl was absent and the test skipped" are currently indistinguishable in the job log, because
`full-tests-x64` only uploads test-results on failure.

## Recommended fix (default if nobody feels strongly)

Mirror `c5f874c` exactly rather than inventing a second convention:

- An `EXTERNAL_TOOLS_REQUIRED`-style env var read at the `findExecutable` gates in both suites: unset
  → skip as today; set → fail naming the missing binary.
- Set it in the `full-tests-x64` job in `.github/workflows/arm-build-test.yaml`, next to the existing
  `DBUSMOCK_REQUIRED: "1"`, with a comment saying why (the job installs `dbus`, which provides
  `dbus-monitor`; `busctl` ships with systemd on `ubuntu-24.04`).

Roughly six lines. Whoever picks it up should confirm on a real CI run that both binaries actually
are on the runner before turning their absence into a hard failure — that is the one thing that
cannot be checked locally.

PR #240, opened 2026-08-08
test(cross): make the native side of a cross-runtime case able to fail, and stop reporting native cases that asserted nothing as passes

### LINKED ISSUE #183: Cross-runtime interop cases cannot fail on the native side: --ktest_no_exit_code makes the exit-code assertion a tautology

### LINKED ISSUE #186: linuxX64Test cannot report a runtime skip: 44 native dbusmock cases still read as PASSED when the harness is absent
## `linuxX64Test` cannot report a runtime skip, so a skipped native case still reads as PASSED

Follow-up from #177 / PR #182, filed so the residual half of that fix is tracked rather than forgotten.

**What #182 delivered.** The nine dbusmock suites now (a) fail loudly with the underlying reason when `DBUSMOCK_REQUIRED` is set — which `full-tests-x64` now sets, since it apt-installs `python3-dbusmock` on purpose — and (b) report a real `skipped` on the JVM when the harness is unavailable: `jvmTest` went from `tests=64 skipped=0` to `tests=64 skipped=44`.

**What it could not deliver.** `linuxX64Test` still reports `tests=62 skipped=0` for the same situation, i.e. 44 native cases that asserted nothing still read as passes.

**Why — measured, not assumed:**

- `TestCase.ignored` in `kotlin.native.internal.test` is a static `val` populated from `@Ignore`; `TestListener.ignore` is only ever called for those. Nothing lets a running native test change its own outcome to ignored.
- Emitting the TeamCity message the native runner itself uses is rejected by the Gradle-side parser. Emitting `##teamcity[testIgnored name='…']` from inside a test body produced:

  ```
  Error while processing test process output message "##teamcity[testIgnored name='callsMethodAndReadsPropertyOnDbusmockPeer' …]"
  java.lang.IllegalStateException: previous test `linuxX64Test/…/callsMethodAndReadsPropertyOnDbusmockPeer` not finished
      at …TCServiceMessagesClient.requireLeafGroup(TCServiceMessagesClient.kt:557)
      at …TCServiceMessagesClient.beginTest(TCServiceMessagesClient.kt:124)
  ```

  `beginTest` requires the leaf node to be a *group*, so `testIgnored` is only accepted between tests — somewhere no test code can reach. The result XML was not written at all.

**Scope of the remaining exposure is small but real.** In CI the lie is gone: `full-tests-x64` sets `DBUSMOCK_REQUIRED`, so a missing harness is a hard failure on both backends. What remains is native runs on machines without `python3-dbusmock` — contributor laptops, and the `arm-runtime-tests` job, which does not install dbusmock and runs the `.kexe` **directly**, with no Gradle in the loop.

**Options, none obviously right — hence filing rather than guessing:**

1. **Accept it**, and rely on `DBUSMOCK_REQUIRED` in CI (status quo after #182). Zero cost, and the only place the report is read as an artifact is already honest.
2. **Exclude the dbusmock suites from `linuxX64Test`** at Gradle configuration time when the harness cannot run, so the report shows nothing instead of phantom passes. Costs auto-detection for contributors who *do* have dbusmock (they would have to set an env var), needs a configuration-time `python3 -c "import dbusmock"` probe that duplicates the harness's own detection, and does nothing for the ARM job.
3. **Upstream**: ask JetBrains for a runtime-skip primitive in the Kotlin/Native test runner. Right fix, wrong timescale.

Recommendation: **option 1** unless the ARM job's phantom passes start mattering, in which case option 2 for `linuxX64Test` plus an explicit `--ktest_gradle_filter` on the ARM invocation.

_Filed by the agent that worked #177. Labelled `agent-workable` with option 1 (do nothing further) as the stated default — it is only worth reopening if you disagree with that default._

monkopedia-coder:
Fixes #183
Fixes #186

Surfaced and filed on the way: #241.

One PR, three commits: #183 and #186 are the same disease on the same surface (native test invocation, exit codes, result reporting) and #186's fix depends on #183's guarantee, so splitting them would have collided. The third commit quarantines the failure the first one surfaced (#241). No workflow file is touched — see "Nothing shared moved" below.

Everything below is measured on **two independent hosts** (both under `dbus-run-session`, JDK 21). Where a number differs from the issue, the issue is corrected.

---

## `--ktest_no_exit_code` — why it was there, and what happened to it

**It was not load-bearing.** It was added in the module's very first commit, `86eaa62` (`test(cross): add cross-runtime interop and stress modules`), with no rationale in the message or in any comment; `git log -S` finds no other commit that reasoned about it; it appears in no doc. The *one* place its behaviour was ever recorded is `fb44864`, which describes it **masking** a real failure:

> the fd interop test … the matching native peer expectation (`invalidRejected`) was unsatisfiable by any client … on main it merely timed out 20s and **the failure was masked by `--ktest_no_exit_code`**.

The plausible "wrapper must collect XML before the process dies" motive does not apply either: nothing wraps these spawns — the JVM test *is* the wrapper, it reads the peer's stdout itself, and it reads it before checking the exit code. So the flag is **dropped**, and the failure is *additionally* propagated from the collected output (the issue's option 2), because the exit code on its own is not enough:

```
$ … test.kexe --ktest_no_exit_code --ktest_logger=TEAMCITY '--ktest_gradle_filter=*nativeClientCanRoundTripUnixFd*'
WITH --ktest_no_exit_code: exit=0     # ##teamcity[testFailed …] in the output
WITHOUT the flag:          exit=1
```

```
$ … test.kexe --ktest_logger=TEAMCITY '--ktest_gradle_filter=*NativeInteropPeerTest.thisTestDoesNotExist*'
zero-match, NO flag: exit=0
0 183-zeromatch.txt        # zero bytes of output, exit 0
```

That second measurement is #219/#223's trap one module over: the filter strings live in these JVM files, decoupled from the native source they name, and a rename empties the case out while leaving it green. So `assertNativePeerPassed` requires three independent things — no `##teamcity[testFailed`, a `testFinished` for the case actually selected, and exit 0 — and the filter is built from the same test-name `val` the assertion reads, so filter and assertion cannot drift apart. Parsing is a plain `contains` / `Regex`, no shell.

Two more things this commit does:

- **Drains the captured output before asserting on it.** The forward cases snapshotted `nativeOutput` *before* `process.waitFor(...)`; now the peer is waited for, the pump thread joined, then the snapshot taken. Without this the new log assertions would be racily flaky.
- **Gives the three cases #183 lists as having no assertion but the exit code a JVM-side observation**, like their siblings have: the signal and `PropertiesChanged` cases assert the JVM peer was actually *asked* to emit; the fd case asserts the fd both arrived and was handed back. This closes the residual hole that a native peer whose role env var does not match returns immediately and reports a pass.

`CrossRuntimeInteropStressTest` gets the same treatment at all four spawn sites. The drop-in-flight case deliberately `SIGKILL`s its peer mid-reply, so it gets no `assertNativePeerPassed` and a comment saying why.

### #183, both directions — from JUnit XML, same host, same command

The mutation is a **purely native-side** assertion — `assertEquals(expectedArg, observedSignalValue.value)` → `expectedArg + 999` in `NativeInteropPeerTest.nativeClientCanObserveSignalFromJvmPeerOverDirectBus`. That is exactly the assertion #183 says is discarded, on one of the three cases it names.

**BEFORE (pristine `main` @ `e2a9f88`, broken native assertion):**

```
gradle exit=0
XML: com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest tests="1" skipped="0" failures="0" errors="0"
```

**AFTER (this branch, the same broken native assertion):**

```
gradle exit=1
XML: com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest tests="1" skipped="0" failures="1" errors="0"
  <failure message="java.lang.AssertionError: Native peer reported
    'nativeClientCanObserveSignalFromJvmPeerOverDirectBus' as failed:
    ##teamcity[testFailed name='nativeClientCanObserveSignalFromJvmPeerOverDirectBus'
      message='Expected &lt;1066&gt;, actual &lt;67&gt;.' …
```

The native assertion's own message reaches the JVM report — the reason the issue preferred scanning the output over relying on the exit code alone.

**AFTER, filter renamed so it matches nothing (the #219 trap):**

```
gradle exit=1
XML: … tests="1" skipped="0" failures="1" errors="0"
  <failure message="java.lang.AssertionError: Native peer never ran
    'nativeClientCanObserveSignalFromJvmPeerOverDirectBusRENAMED', so this case measured nothing
    — the filter --ktest_gradle_filter=*NativeInteropPeerTest.…RENAMED* matched no test.
    Was it renamed in NativeInteropPeerTest?"
```

**AFTER, unmutated, stress suite (`repeat=2`, forward + reverse):**

```
stress exit=0
XML: com.monkopedia.sdbus.stress.CrossRuntimeInteropStressTest tests="10" skipped="0" failures="0" errors="0"
```

---

## ⚠️ Enabling the failure path revealed a real one: the reverse unix-fd case has never worked

**`nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus` fails on this branch**, with the native client's own message:

```
XML: com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest tests="14" skipped="0" failures="1" errors="0"
  <failure message="java.lang.AssertionError: Native peer reported
    'nativeClientCanRoundTripUnixFdWithJvmPeerOverDirectBus' as failed:
    ##teamcity[testFailed name='…' message='Timed out waiting for call to succeed'
      details='kotlin.AssertionError: Timed out waiting for call to succeed| at … NativeInteropPeerTest.$retryCallCOROUTINE$14 …
```

**This is not a regression from this PR, and the receipt is in #185's own evidence table.** #185 recorded that case as "**pass** | **15.207s**" while its six siblings were 0.044s–0.144s. 15.2s is exactly `retryCall(timeoutMillis = 15_000)` exhausting. My run reports `time="15.207"` — the identical number, now as a failure. The 15-second retry loop was always timing out; `--ktest_no_exit_code` swallowed it and the timing was the only tell.

That is #183 working as intended, and it is the outcome the owner asked for on the parent issue — [#176 comment](https://github.com/Monkopedia/sdbus-kotlin/issues/176#issuecomment-5171520219): *"if a newly-live case fails, that is a successful outcome for this issue, not a blocker for it"*, with *"failures filed separately rather than fixed in the same PR"*.

**CI answered: it reproduces there too, so it is not environment-specific.** `full-tests-x64` on `ubuntu-24.04` reported exactly one failure across the whole job — [run 31253765479](https://github.com/Monkopedia/sdbus-kotlin/actions/runs/31253765479/job/93093975345):

```
> Task :cross_test:jvmInteropTest
CrossRuntimeInteropSmokeTest > nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus FAILED
    java.lang.AssertionError at CrossRuntimeInteropSmokeTest.kt:1359
14 tests completed, 1 failed
```

Three independent environments, same case, only that case. So it is **filed as #241 and marked `@Ignore` here, not fixed** — per the owner's ruling on #176 that a newly-live case failing is a successful outcome for the enabling issue and gets filed separately.

`@Ignore` rather than deletion, and rather than an `Assume`: the report then says `skipped`, which is *true*, and #241 says why — an assumption would invent a precondition this case does not actually have. The annotation and a KDoc block above the case both name #241 and say "do not re-enable without fixing it".

Everything else in that CI job passed, which is itself two pieces of #186 evidence: `:cross_test:linuxX64Test` ran green under `DBUSMOCK_REQUIRED=1` (so the 44 dbusmock cases genuinely executed rather than being excluded), and all nine `pr-smoke` stress jobs passed with `--ktest_no_exit_code` removed.

---

## #186 — native cannot record a skip, so the suites are not run

The **44** figure is **verified**: the linked `cross_test` binary's `--gtest_list_tests` lists 62 tests, of which 44 are under `Dbusmock*`. Baseline JUnit XML, no dbusmock on the box:

```
cross_test/build/test-results/linuxX64Test/ : tests=62 skipped=0 failures=0 errors=0
cross_test/build/test-results/jvmTest/      : tests=50 skipped=44 failures=0 errors=0
```

(The issue quoted `jvmTest` as `tests=64 skipped=44`; 64 predates #185's exclusion of `CrossRuntimeInteropSmokeTest` from `jvmTest`. The **44** is unchanged.)

**Two corrections to the issue.**

1. **`arm-runtime-tests` is not affected.** It runs the **root** binary, whose `--gtest_list_tests` lists **316** tests and **0** `Dbusmock*` cases — the dbusmock suites live only in `cross_test`. So there is no ARM exposure and no workflow change is needed.
2. **It is not 44 phantom passes, it is 58.** The same 62 also contain **14** `NativeInteropPeerTest` cases, every one of which returns immediately unless `KDBUS_NATIVE_INTEROP_ROLE` names its role — which only the JVM side sets when it *spawns* the binary. `linuxX64Test` never sets it, so all 14 always report a pass having asserted nothing. (#228 noted these and left them because no native skip primitive exists; the same lever fixes them.)

**Re-measured: there is no way to render these as `<skipped/>`.** On top of the issue's own finding that a mid-test TeamCity `testIgnored` is rejected by the Gradle parser, I checked the remaining candidate — whether the runner emits `testIgnored` for tests a `--ktest_gradle_filter` excludes. It does not (`testIgnored` count **0**, `testFinished` count 3, for a filter admitting one class of three). So the JVM's `skipped=44` has no native counterpart, and the only honest lever is not running them: **absent beats a phantom pass.** This is the issue's option 2, in c5f874c's idiom rather than a second mechanism.

`DBUSMOCK_REQUIRED` is the switch:

| condition | `:cross_test:linuxX64Test` | build |
| --- | --- | --- |
| before, no dbusmock | `tests="62" skipped="0" failures="0" errors="0"` | pass (the lie) |
| after, no dbusmock, `DBUSMOCK_REQUIRED` unset | `tests="4" skipped="0" failures="0" errors="0"` | pass — 58 phantom passes gone from the XML |
| after, no dbusmock, **`DBUSMOCK_REQUIRED=1`** | `tests="48" skipped="0" **failures="44"** errors="0"` | **FAILS** |

The third row is the one that matters: **CI cannot lose this coverage to the exclusion**, because `full-tests-x64` apt-installs `python3-dbusmock` and sets `DBUSMOCK_REQUIRED=1`, so absence there is a hard failure naming the reason — exactly c5f874c's contract, preserved. `48 = 62 − 14`, i.e. on CI the only change is the 14 role-gated peer cases no longer being reported by a task that cannot drive them.

Row two is a big collapse, and it is the truth: on a box without dbusmock, `:cross_test:linuxX64Test` only ever exercised those 4 cases. The remaining two suites are `CrossModuleSignalIntegrationTest` (3) and `EventLoopStartIdempotencyTest` (1). The exclusion announces itself so a log distinguishes "ran" from "was not run" — the visibility complaint raised in #229:

```
cross_test: excluding the Dbusmock* suites from linuxX64Test — 'python3 -m dbusmock' cannot run
here. Kotlin/Native cannot report these as skipped (#186), so running them would report passes
that asserted nothing. Set DBUSMOCK_REQUIRED=1 to run them anyway and fail loudly instead.
```

The probe mirrors the native `launchDbusmock` exactly — a session bus plus an importable `dbusmock` under `DBUSMOCK_PYTHON`/`python3` — so a contributor who *has* dbusmock keeps running the suites with no configuration.

**Known limit, stated rather than papered over:** if the `Dbusmock*` / `NativeInteropPeerTest` names drift, the exclusion silently stops matching. That fails *open* — back to today's behaviour, not worse — and Gradle's filter API offers no "this exclusion matched nothing" signal to guard it with. #229's `EXTERNAL_TOOLS_REQUIRED` work is untouched and uncontradicted.

---

## Gates and shared files

- `./gradlew ktlintCheck apiCheck` → **BUILD SUCCESSFUL**; `git status --porcelain api/` **empty** — `api/*.api` does not move (test-only + build-script changes).
- **No workflow file is touched.** #223 changed `arm-build-test.yaml` recently and the #196 agent may want it; this PR needs nothing from it, because `DBUSMOCK_REQUIRED=1` is already set in `full-tests-x64` and the root binary the ARM job drives has no dbusmock cases. Nothing to sequence.

## Tasks executed (all under `dbus-run-session`)

`:cross_test:linkDebugTestLinuxX64`, `:linkDebugTestLinuxX64`, `:cross_test:linuxX64Test`, `:cross_test:jvmTest`, `:cross_test:jvmInteropTest` (with and without `-Dkdbus.crossRuntimeInterop.reverse.enabled=true`, and with `--tests` on a single case), `:stress_test:jvmInteropStressTest`, `ktlintCheck`, `apiCheck`.

### DIFF
--- a/cross_test/build.gradle.kts
+++ b/cross_test/build.gradle.kts
@@ -1,3 +1,4 @@
+import org.gradle.api.tasks.testing.AbstractTestTask
 import org.gradle.api.tasks.testing.Test
 import org.jetbrains.kotlin.gradle.tasks.KotlinNativeLink
 
@@ -56,6 +57,55 @@ val reverseInteropEnabled = providers
 // actually configures it — mirroring the gcSoak gate in the root build.
 tasks.named<Test>("jvmTest") {
     filter.excludeTestsMatching("com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest")
+}
+
+// Kotlin/Native's test runner has no runtime-skip primitive, so `DbusmockHarness.skipTest`'s native
+// actual can only print its reason: a Dbusmock* case that asserted nothing because python-dbusmock
+// is missing is reported as a PASS, and `:cross_test:linuxX64Test` claimed 44 of them (#186). The
+// JVM half of that was fixed in c5f874c (#182) with `org.junit.Assume`; native has no equivalent, so
+// the only honest lever left is to not run the suites at all — absent beats a phantom pass.
+//
+// The decision follows c5f874c rather than adding a second convention:
+//  * `DBUSMOCK_REQUIRED` set — the `full-tests-x64` job, which apt-installs python3-dbusmock on
+//    purpose — the suites always run, and the harness itself fails loudly when it cannot start a
+//    peer. So CI can never lose this coverage to the exclusion below.
+//  * otherwise the harness is probed exactly as the native `launchDbusmock` probes it (a session bus
+//    plus an importable `dbusmock` module under `DBUSMOCK_PYTHON` / `python3`), and the suites are
+//    excluded only when it genuinely cannot run — a contributor who *has* dbusmock still runs them
+//    with no configuration.
+val dbusmockRequired = providers.environmentVariable("DBUSMOCK_REQUIRED").isPresent
+val dbusmockPython = providers.environmentVariable("DBUSMOCK_PYTHON").getOrElse("python3")
+val sessionBusPresent = providers.environmentVariable("DBUS_SESSION_BUS_ADDRESS").isPresent
+
+fun dbusmockCanRun(): Boolean {
+    if (dbusmockRequired) return true
+    if (!sessionBusPresent) return false
+    return runCatching {
+        ProcessBuilder(dbusmockPython, "-c", "import dbusmock")
+            .redirectErrorStream(true)
+            .redirectOutput(ProcessBuilder.Redirect.DISCARD)
+            .start()
+            .waitFor() == 0
+    }.getOrDefault(false)
+}
+
+tasks.named<AbstractTestTask>("linuxX64Test") {
+    // NativeInteropPeerTest's cases are the *peer half* of a two-process case: each one returns
+    // immediately unless KDBUS_NATIVE_INTEROP_ROLE names its role, which only the JVM side sets when
+    // it spawns this binary (see CrossRuntimeInteropSmokeTest). `linuxX64Test` never sets it, so
+    // every one of them runs to completion having asserted nothing — and, per #186, native reports
+    // that as a pass. They are covered where they are actually driven: `jvmInteropTest`, which now
+    // fails unless the spawned peer really ran its case and reported it passing (#183).
+    filter.excludeTestsMatching("com.monkopedia.sdbus.integration.NativeInteropPeerTest")
+    if (!dbusmockCanRun()) {
+        logger.lifecycle(
+            "cross_test: excluding the Dbusmock* suites from linuxX64Test — '$dbusmockPython -m " +
+                "dbusmock' cannot run here. Kotlin/Native cannot report these as skipped (#186), " +
+                "so running them would report passes that asserted nothing. Set " +
+                "DBUSMOCK_REQUIRED=1 to run them anyway and fail loudly instead."
+        )
+        filter.excludeTestsMatching("com.monkopedia.sdbus.integration.Dbusmock*")
+    }
 }
 
 tasks.register<Test>("jvmInteropTest") {
--- a/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockHarness.kt
+++ b/cross_test/src/commonTest/kotlin/com/monkopedia/sdbus/integration/DbusmockHarness.kt
@@ -90,8 +90,10 @@ internal fun dbusmockUnavailable(reason: String): DbusmockHandle? {
  * On the JVM this raises a JUnit assumption failure, which Gradle records as `skipped`. Kotlin/
  * Native's test runner has no runtime-skip primitive — an early return is indistinguishable from a
  * pass, and a TeamCity `testIgnored` emitted from inside a running test is rejected by the Gradle
- * parser — so the native actual can only write [reason] to the captured test output. CI does not
- * depend on that: [DBUSMOCK_REQUIRED_ENV] makes the harness fail rather than skip.
+ * parser — so the native actual can only write [reason] to the captured test output. Because a
+ * printed reason does not stop the report claiming a pass (#186), `:cross_test:linuxX64Test` does not
+ * run these suites at all when the harness cannot start: see the exclusion in
+ * `cross_test/build.gradle.kts`, which reads [DBUSMOCK_REQUIRED_ENV] as its override.
  *
  * Callers must still `return` afterwards; only the JVM actual aborts the test by throwing.
  */
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
@@ -32,6 +32,7 @@ import org.freedesktop.dbus.exceptions.DBusExecutionException
 import org.freedesktop.dbus.interfaces.DBusInterface
 import org.freedesktop.dbus.messages.DBusSignal
 import org.junit.Assume.assumeTrue
+import org.junit.Ignore
 
 class CrossRuntimeInteropSmokeTest {
     /**
@@ -929,6 +930,23 @@ class CrossRuntimeInteropSmokeTest {
         }
     }
 
+    /**
+     * KNOWN FAILING — tracked as #241, do not re-enable without fixing it.
+     *
+     * This case has never worked. Its native half times out for the full
+     * `retryCall(timeoutMillis = 15_000)` budget calling `StorePipeReadFd` on the dbus-java peer;
+     * before the exit-code assertion above became real (#183) the failure was discarded and the case
+     * reported PASSED. The only tell was the clock — PR #185 recorded it as "pass, 15.207s" next to
+     * siblings at 0.05s. It reproduces identically on two dev boxes and on `full-tests-x64`, so it is
+     * a real defect and not an environment artefact.
+     *
+     * `@Ignore` rather than deletion, and rather than an `Assume` that would invent a precondition
+     * this case does not actually have: the report says `skipped`, which is true, and #241 says why.
+     */
+    @Ignore(
+        "Known failing, tracked as #241: the native client's StorePipeReadFd call to the " +
+            "dbus-java peer never succeeds, so its 15s retry budget always exhausts."
+    )
     @Test
     fun nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus() {
         assumeCrossRuntimeInteropSelected()
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
@@ -71,11 +71,11 @@ class CrossRuntimeInteropSmokeTest {
         val objectPath = "/org/monkopedia/sdbus/phase3/Object"
         val expectedArg = 33
         val nativeOutput = ByteArrayOutputStream()
+        val peerTest = "jvmClientCanInvokeIncrementOverDirectBus"
         val process = ProcessBuilder(
             kexe.toString(),
-            "--ktest_no_exit_code",
             "--ktest_logger=TEAMCITY",
-            "--ktest_gradle_filter=*NativeInteropPeerTest.jvmClientCanInvokeIncrementOverDirectBus*"
+            nativePeerFilter(peerTest)
         ).also { builder ->
             builder.environment()["KDBUS_NATIVE_INTEROP_ROLE"] = "server"
             builder.environment()["KDBUS_INTEROP_SOCKET"] = socketPath.toString()
@@ -117,16 +117,14 @@ class CrossRuntimeInteropSmokeTest {
                 connection.release()
             }
 
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
             assertTrue(
                 process.waitFor(10, TimeUnit.SECONDS),
-                "Native peer did not exit in time. Output:\n$nativeLog"
+                "Native peer did not exit in time. Output:\n" +
+                    nativeOutput.toString(Charsets.UTF_8)
             )
-            assertEquals(
-                0,
-                process.exitValue(),
-                "Native peer failed. Output:\n$nativeLog"
-            )
+            outputPump.join(PUMP_DRAIN_MILLIS)
+            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
             outputPump.join(1_000)
@@ -151,11 +149,11 @@ class CrossRuntimeInteropSmokeTest {
         val seenValue = AtomicInteger(Int.MIN_VALUE)
         val signalLatch = CountDownLatch(1)
         val nativeOutput = ByteArrayOutputStream()
+        val peerTest = "jvmClientCanObserveSignalFromNativePeerOverDirectBus"
         val process = ProcessBuilder(
             kexe.toString(),
-            "--ktest_no_exit_code",
             "--ktest_logger=TEAMCITY",
-            "--ktest_gradle_filter=*NativeInteropPeerTest.jvmClientCanObserveSignalFromNativePeerOverDirectBus*"
+            nativePeerFilter(peerTest)
         ).also { builder ->
             builder.environment()["KDBUS_NATIVE_INTEROP_ROLE"] = "server-signal"
             builder.environment()["KDBUS_INTEROP_SOCKET"] = socketPath.toString()
@@ -199,16 +197,14 @@ class CrossRuntimeInteropSmokeTest {
                 runCatching { javaConnection.disconnect() }
             }
 
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
             assertTrue(
                 process.waitFor(10, TimeUnit.SECONDS),
-                "Native peer did not exit in time. Output:\n$nativeLog"
+                "Native peer did not exit in time. Output:\n" +
+                    nativeOutput.toString(Charsets.UTF_8)
             )
-            assertEquals(
-                0,
-                process.exitValue(),
-                "Native peer failed. Output:\n$nativeLog"
-            )
+            outputPump.join(PUMP_DRAIN_MILLIS)
+            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
             outputPump.join(1_000)
@@ -234,11 +230,11 @@ class CrossRuntimeInteropSmokeTest {
         val seenInterfaceName = AtomicReference<String?>(null)
         val signalLatch = CountDownLatch(1)
         val nativeOutput = ByteArrayOutputStream()
+        val peerTest = "jvmClientCanObservePropertiesChangedFromNativePeerOverDirectBus"
         val process = ProcessBuilder(
             kexe.toString(),
-            "--ktest_no_exit_code",
             "--ktest_logger=TEAMCITY",
-            "--ktest_gradle_filter=*NativeInteropPeerTest.jvmClientCanObservePropertiesChangedFromNativePeerOverDirectBus*"
+            nativePeerFilter(peerTest)
         ).also { builder ->
             builder.environment()["KDBUS_NATIVE_INTEROP_ROLE"] = "server-properties"
             builder.environment()["KDBUS_INTEROP_SOCKET"] = socketPath.toString()
@@ -284,16 +280,14 @@ class CrossRuntimeInteropSmokeTest {
                 runCatching { javaConnection.disconnect() }
             }
 
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
             assertTrue(
                 process.waitFor(10, TimeUnit.SECONDS),
-                "Native peer did not exit in time. Output:\n$nativeLog"
+                "Native peer did not exit in time. Output:\n" +
+                    nativeOutput.toString(Charsets.UTF_8)
             )
-            assertEquals(
-                0,
-                process.exitValue(),
-                "Native peer failed. Output:\n$nativeLog"
-            )
+            outputPump.join(PUMP_DRAIN_MILLIS)
+            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
             outputPump.join(1_000)
@@ -314,11 +308,11 @@ class CrossRuntimeInteropSmokeTest {
         val objectPath = "/org/monkopedia/sdbus/phase3/Object"
         val expectedByte = 0x4D
         val nativeOutput = ByteArrayOutputStream()
+        val peerTest = "jvmClientCanRoundTripUnixFdOverDirectBus"
         val process = ProcessBuilder(
             kexe.toString(),
-            "--ktest_no_exit_code",
             "--ktest_logger=TEAMCITY",
-            "--ktest_gradle_filter=*NativeInteropPeerTest.jvmClientCanRoundTripUnixFdOverDirectBus*"
+            nativePeerFilter(peerTest)
         ).also { builder ->
             builder.environment()["KDBUS_NATIVE_INTEROP_ROLE"] = "server-fd"
             builder.environment()["KDBUS_INTEROP_SOCKET"] = socketPath.toString()
@@ -397,16 +391,14 @@ class CrossRuntimeInteropSmokeTest {
                 connection.release()
             }
 
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
             assertTrue(
                 process.waitFor(10, TimeUnit.SECONDS),
-                "Native peer did not exit in time. Output:\n$nativeLog"
+                "Native peer did not exit in time. Output:\n" +
+                    nativeOutput.toString(Charsets.UTF_8)
             )
-            assertEquals(
-                0,
-                process.exitValue(),
-                "Native peer failed. Output:\n$nativeLog"
-            )
+            outputPump.join(PUMP_DRAIN_MILLIS)
+            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
             outputPump.join(1_000)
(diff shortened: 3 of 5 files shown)


monkopedia-reviewer:
**Reviewer, interim evidence** (independent fresh-context review of `c8a665e`; two more local runs still queued behind the fleet build slot — posting what is already measured so it is durable).

### Independent baseline, pristine `main` @ `e2a9f88`, second host (`adolin`, JDK 21, `dbus-run-session`, `python3 -c "import dbusmock"` → `ModuleNotFoundError`)

`:cross_test:linuxX64Test`, fresh JUnit XML, `BUILD SUCCESSFUL`, gradle exit **0**:

```
TOTAL tests=62 skipped=0 failures=0 errors=0   (12 XML files)
  CrossModuleSignalIntegrationTest        tests=3
  EventLoopStartIdempotencyTest           tests=1
  Dbusmock{AdaptorPropertiesChanged,Bluez,ForeignError,ObjectManager,
           PropertiesChanged,SecretService,Signal,Smoke,TypeMatrix}Test
                                          tests=3+7+5+2+5+6+3+1+12 = 44
  NativeInteropPeerTest                   tests=14
```

So every arithmetic claim in the PR body is confirmed against the binary rather than against the issue: **62 = 44 + 14 + 4**, and the residual 4 really are `CrossModuleSignalIntegrationTest` (3) + `EventLoopStartIdempotencyTest` (1). The issue's "44" is right and the PR's re-count to "58 phantom passes" is right.

### #183's load-bearing claim, reproduced

Same host, same command, pristine `main`, with the **purely native-side** mutation
(`NativeInteropPeerTest.nativeClientCanObserveSignalFromJvmPeerOverDirectBus`:
`assertEquals(expectedArg, …)` → `assertEquals(expectedArg + 999, …)`):

```
$ dbus-run-session -- ./gradlew :cross_test:jvmInteropTest \
    -Dkdbus.crossRuntimeInterop.reverse.enabled=true \
    --tests "…CrossRuntimeInteropSmokeTest.nativeClientObservesJvmSignalOverDirectBus" --rerun-tasks
BUILD SUCCESSFUL ; gradle exit=0
<testsuite name="…CrossRuntimeInteropSmokeTest" tests="1" skipped="0" failures="0" errors="0" … >
```

Confirmed: on `main` a deliberately-broken native assertion is discarded and the JVM case reports a pass. The branch-side half of this comparison is still running.

### This repo's CI is now the third environment, and it reproduces the fd failure

`full-tests-x64` on `c8a665e` is **red**, and it is exactly the case the PR body predicted:

```
CrossRuntimeInteropSmokeTest > nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus FAILED
    java.lang.AssertionError at CrossRuntimeInteropSmokeTest.kt:1359
14 tests completed, 1 failed
```

`CrossRuntimeInteropSmokeTest.kt:1359` is the `reportedFailure == null` assertion inside `assertNativePeerPassed` — i.e. the peer's own `##teamcity[testFailed …]`, not a false "never ran". Every other check on the PR is green, including all four `reverse` `pr-smoke` control scenarios.

That settles the open question the PR body left for CI: the failure is **not** Arch-specific. Per the author's own stated plan, that triggers the file-the-issue-and-mark-the-case follow-up commit. Verdict to follow once the branch-side measurements land.

monkopedia-reviewer (approved):
## Approve — reviewed at `4c82c99`, all three claims independently reproduced

Fresh-context review. I re-derived every load-bearing number from JUnit XML and from the test binary rather than from the PR body, and I broke the new gate on purpose three different ways to establish it is live. Both hosts are Arch, JDK 21, under `dbus-run-session`, with `python3 -c "import dbusmock"` → `ModuleNotFoundError` confirmed on each.

### 1. #183 — the gate is real (`pass → break → RED → revert → clean`)

Mutation: `NativeInteropPeerTest.nativeClientCanObserveSignalFromJvmPeerOverDirectBus`, `assertEquals(expectedArg, observedSignalValue.value)` → `expectedArg + 999`. Purely native-side; the JVM peer cannot observe it. **Same host, same command, both sides:**

| | gradle exit | `<testsuite …>` |
| --- | --- | --- |
| pristine `main` @ `e2a9f88` | **0**, `BUILD SUCCESSFUL` | `tests="1" skipped="0" failures="0" errors="0"` |
| this branch | **1**, `BUILD FAILED` | `tests="1" skipped="0" failures="1" errors="0"` |

The specific claim I was asked to check — that `main` reports `tests="1" skipped="0" failures="0"` and exits 0 — holds exactly. And the branch propagates the native assertion's own text:

```
Native peer reported 'nativeClientCanObserveSignalFromJvmPeerOverDirectBus' as failed:
##teamcity[testFailed name='…' message='Expected &lt;1066&gt;, actual &lt;67&gt;.' details='kotlin.AssertionError: …
```

I also broke the **other** two legs, because an exit-code-only fix would have satisfied the first:

- **Filter drift (the #219 trap).** Renaming only the JVM-side `peerTest` val → `failures="1"`, message `Native peer never ran '…RENAMED', so this case measured nothing — the filter … matched no test.` Independently confirmed why this leg is needed: `test.kexe --ktest_gradle_filter=*NativeInteropPeerTest.thisTestDoesNotExistAtAll*` exits **0** with **0 bytes** of output.
- **Format.** The runner really does emit `##teamcity[testFinished name='<bare test name>' duration='…']`, so the `contains` check matches the actual output rather than a guessed shape.

`git status --porcelain` empty after every cycle; no probe-shaped names left in the diff.

**On dropping `--ktest_no_exit_code`: I looked for a purpose and found none.** `86eaa62` is subject-line only (no body, no comment); `git log -S` surfaces only `08fbbf0`/`fb44864`, and `fb44864` is the one place its behaviour was recorded — as *masking* a failure. `git grep` over `main` finds it in no workflow and no doc. Nothing wraps these spawns. Not a blocking finding.

### 2. #186 — every number exact

Enumerated from the linked binary, not from the issue: `--gtest_list_tests` → **62** total = **44** `Dbusmock*` + **14** `NativeInteropPeerTest` + **4** (`CrossModuleSignalIntegrationTest` 3, `EventLoopStartIdempotencyTest` 1).

| condition | `:cross_test:linuxX64Test` | build |
| --- | --- | --- |
| `main`, no dbusmock | `tests="62" skipped="0" failures="0" errors="0"` | pass (exit 0) |
| branch, no dbusmock, `DBUSMOCK_REQUIRED` unset | `tests="4" skipped="0" failures="0" errors="0"` | pass |
| branch, no dbusmock, **`DBUSMOCK_REQUIRED=1`** | `tests="48" skipped="0" **failures="44"** errors="0"` | **FAILS**, exit 1 |

All three rows match the PR body exactly, and every one of the 44 failures carries `DBUSMOCK_REQUIRED is set but no dbusmock peer could be started: 'python3 -m dbusmock' exited immediately with status 1`. `48 = 62 − 14`, so CI loses only the 14 role-gated peer cases, which `full-tests-x64` drives via `crossRuntimeInteropTests` anyway.

**Re-measurement confirmed:** for a `--ktest_gradle_filter` admitting one class, `testIgnored` count is **0** while `testFinished` fires normally. There is no native lever that renders a filtered case as `<skipped/>`, so "not run" really is the only honest option, and it beats a phantom pass.

**Announcement verified from full output (not a `tail`):** the `logger.lifecycle` line prints, once, and is the first line of the run — so a log distinguishes "ran" from "was not run", which is #229's visibility point.

**Both corrections to the issue check out.**
- `arm-runtime-tests` is genuinely unaffected — structurally, not just by count: `native-build-x64` uploads only `build/bin/linux{X64,Arm64}/debugTest/test.kexe`, i.e. the **root** binary, and the `Dbusmock*` classes exist only under `cross_test/src/commonTest`. No ARM exposure, so no workflow change was needed.
- 58, not 44. And excluding the 14 is **in scope, not scope creep**: #228's own review already recorded *"`NativeInteropPeerTest` is correctly excluded. Its 14 gates…"* and left them only because no lever existed. This is the same defect, the same task, and the lever that finally fits.

`DBUSMOCK_REQUIRED` is reused rather than a second convention being invented — c5f874c's contract preserved, #229's `EXTERNAL_TOOLS_REQUIRED` work untouched and uncontradicted.

### 3. Gates

`./gradlew ktlintCheck apiCheck` → `BUILD SUCCESSFUL`; `git status --porcelain api/` **empty**; whole tree clean. **CI green at `4c82c99`** — 15/15 real checks pass (`full-tests-x64` 5m49s), 2 skipping by design.

### 4. #241 and the `@Ignore` — you asked me to push back rather than rubber-stamp, so here is the actual reasoning

The fd case reproduces on **four** environments now (two dev boxes, mine, and `full-tests-x64`), and I confirmed the #185 reading independently: my unmutated branch run reports `time="15.207"` for that case against 0.038s–0.235s siblings, and the native client's `retryCall(timeoutMillis = 15_000)` at `NativeInteropPeerTest.kt:791,805` is exactly that budget. #185's table recording it as `pass | 15.207s` was the tell, and the timing was the *only* tell. Good find.

I considered the alternative I'd normally prefer — **pin the broken behaviour as an expectation** so it self-retires the moment someone fixes it, which is this repo's own idiom in `df4db9c` ("pin sd-bus's PropertiesChanged flag contract and record the JVM divergence"). **I don't think it applies here, and `@Ignore` is the better instrument:** the only observable is "the 15 s retry budget exhausts", the cause is unknown (sdbus-kotlin's fd marshalling vs. dbus-java's `NEGOTIATE_UNIX_FD` is undetermined), and a pin on that observable would be satisfied by *any* broken implementation — a test that passes for the wrong reason. It would also spend 15 s of wall clock per run to assert nothing specific. Agreed too that an `Assume` would be worse: it would invent a precondition the case does not have.

Verified the marker is not itself a phantom: `cross_test` has no `useJUnitPlatform()` and `kotlin.test.Test` resolves to kotlin-test-junit here (the file already uses `org.junit.Assume`), so `org.junit.Ignore` is the matching annotation — and the XML confirms it, `tests="14" skipped="1" failures="0"` with the case **present** as a `<skipped/>` entry rather than vanishing from the report. The annotation reason and the KDoc both name #241 and say "do not re-enable without fixing it", which is what makes the quarantine retirable instead of permanent. Right call.

### 5. Output drain

Necessary and, in practice, sufficient: the snapshot now happens after `waitFor` **and** the pump join, and `Thread.join` supplies the memory barrier for the `ByteArrayOutputStream` read. Since the peer has exited and spawns no children, the pipe is at EOF and the pump terminates in microseconds. No flakiness signal across my runs, 13/13 non-ignored cases and 9/9 `pr-smoke` jobs. Not a blocking finding — see the one nit below.

---

## Non-blocking findings

1. **`cross_test/build.gradle.kts` — "probed exactly as the native `launchDbusmock` probes it" overstates it.** `launchDbusmock` forks `python3 -m dbusmock --session [-t <template>] …` and gives it a 500 ms window to stay alive; the configuration-time probe only runs `import dbusmock`. So on a box where dbusmock imports but a peer cannot actually start — an older dbusmock missing a template like `bluez5`, or a bus-name claim failure — the suites still run and native still reports phantom passes. Narrow, but it is the exact window the PR's premise is about. Suggest dropping "exactly" and listing this next to the name-drift limit you already state.

2. **`docs/TEST_COVERAGE.md:82`** still says *"dbusmock suites can false-pass locally without a configured dbusmock venv; trust CI / the venv for those rows, not a bare local green."* Worth refreshing: on `linuxX64Test` they are now excluded rather than false-passing, and on `jvmTest` they have reported `<skipped/>` since #182. **I deliberately did not treat this as the #228 blocking finding**, and the reason is finding 1: because the probe is not equivalent to the launch, the sentence is not actually *falsified* — the residual importable-but-unlaunchable window keeps it literally true, and its advice ("don't trust a bare local green for those rows") is if anything more important now that the rows are silently not run. Stale in mechanism, not false. Your call.

3. **`assertNativePeerPassed` / `nativePeerFilter` duplicated across the two modules — right call, and I checked rather than assumed.** `stress_test` depends on `project(":")` only, has no route to `cross_test`'s test source set, and the two files already duplicate `nativeTestBinaryPath`, `waitForSocket`, `installNullSafeSaslCollatorWorkaround` and `setStaticField`. Sharing would cost a new module or a test-fixtures publication for two helpers. That said the duplicated count is now **6** — if it grows again, a shared test-fixtures source set is the follow-up.

4. **`outputPump.join(PUMP_DRAIN_MILLIS)` discards whether the drain finished.** If it ever did time out, the truncated snapshot fails as *"Native peer never ran 'X', so this case measured nothing"* — the most misleading of the three messages, because it points at a rename that did not happen. One line fixes it: `assertFalse(outputPump.isAlive, …)` after the join, so an incomplete drain reports as an incomplete drain. Both files.

5. Two micro-nits, take or leave: `assertTrue(reportedFailure == null, …)` reads better as `assertNull(reportedFailure, …)`; and `TEAMCITY_TEST_FAILED` is not scoped to `peerTest`, so a `testFailed` for some other test would be reported under `peerTest`'s name — impossible while the filter admits exactly one test, but the message would lie if that ever changed.

6. PR body still opens with "One PR, two commits" — there are three now.

None of these blocks. Approving and merging: `test_only` → tier 1, CI confirmed green at `4c82c99`.

Issue #241, opened 2026-08-08
Reverse cross-runtime unix-fd case has never worked: nativeClientRoundTripsUnixFd... times out against the dbus-java peer


monkopedia-coder:
## `nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus` has never worked; the exit-code tautology hid it

**Category:** test-gap turned real failure (surfaced by #183, filed separately per the owner's ruling on #176)

**Where:** `cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt` — `nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus`, whose native half is `NativeInteropPeerTest.nativeClientCanRoundTripUnixFdWithJvmPeerOverDirectBus`.

### What happens

The native client connects to a dbus-java `DirectConnectionBuilder` peer (AUTH_ANONYMOUS, over a unix socket fd) and calls `StorePipeReadFd` with a real pipe read end. That call never succeeds. `retryCall(timeoutMillis = 15_000)` retries for the full 15 seconds and then throws:

```
##teamcity[testFailed name='nativeClientCanRoundTripUnixFdWithJvmPeerOverDirectBus'
  message='Timed out waiting for call to succeed'
  details='kotlin.AssertionError: Timed out waiting for call to succeed|n
    at … NativeInteropPeerTest.$retryCallCOROUTINE$14.invokeSuspend …
```

The JVM side of the case, after #183, reports:

```
<testsuite name="com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest"
           tests="14" skipped="0" failures="1" errors="0">
```

The other 13 cases in the suite pass, including the **forward** unix-fd case (`jvmClientRoundTripsUnixFdWithNativePeerOverDirectBus`), so this is specific to the native-client → dbus-java-peer direction.

### Why it was invisible

Before #183, the only assertion on this case's native half was `assertEquals(0, process.exitValue())`, and the peer was spawned with `--ktest_no_exit_code`, so a failing native test still exited 0. The case reported PASSED.

**The tell was in the timing, and it is on the record.** PR #185's own evidence table, which turned these seven reverse cases on, lists:

| reverse case | result | time |
| --- | --- | --- |
| `nativeClientInvokesJvmPeerOverDirectBus` | pass | 0.044s |
[...]

PR #242, opened 2026-08-08
test(cross): address the non-blocking review findings from #240

### LINKED ISSUE #183: Cross-runtime interop cases cannot fail on the native side: --ktest_no_exit_code makes the exit-code assertion a tautology

monkopedia-coder:
Follows up #240 (which fixed #183 and #186) with four of the six **non-blocking** findings from its review. No behaviour under test changes; the two gates #240 installed are re-verified in both directions afterwards.

## What changed

**1. An incomplete output drain now reports itself.** `outputPump.join(PUMP_DRAIN_MILLIS)` threw away whether the pump had actually finished. Since `assertNativePeerPassed` reads that capture, a timed-out drain would have failed as *"Native peer never ran 'X', so this case measured nothing"* — the most misleading of the three messages, because it accuses a rename that never happened. The join and the snapshot move into one helper:

```kotlin
private fun drainedPeerOutput(pump: Thread, sink: ByteArrayOutputStream): String {
    pump.join(PUMP_DRAIN_MILLIS)
    assertFalse(pump.isAlive, "Native peer output was still draining after …")
    return sink.toString(Charsets.UTF_8)
}
```

The reverse cases (seven in `cross_test`, one in `stress_test`) hoist the pump into a local `val` and assign that to the existing `var outputPump: Thread?`, so the helper takes a non-null `Thread` rather than a `!!` on a nullable.

**2. The `testFailed` lookup is scoped to the case actually selected.** It matched *any* `##teamcity[testFailed` line and then reported it under `peerTest`'s name, so the message would have lied if the filter ever admitted more than one test. It now looks for that test's own marker. This drops the `TEAMCITY_TEST_FAILED` regex constant from both companions — the check is a plain `contains` over lines — and `assertTrue(x == null)` becomes `assertNull`. Coverage is unchanged: a `testFailed` for some *other* test still fails the case through the exit-code leg.

**3. The build-script comment no longer overstates the probe.** It claimed the configuration-time check works "**exactly** as the native `launchDbusmock` probes it". It does not: `launchDbusmock` forks `python3 -m dbusmock --session [-t <template>] …` and gives it a liveness window, while the probe only checks that the module imports. So on a box where `dbusmock` imports but a peer cannot actually start — an older dbusmock missing a template like `bluez5`, a bus-name claim failure — the suites still run and native still reports passes that asserted nothing. That window is now documented next to the name-drift limit #240 already stated, with both labelled as failing **open** (back to the phantom passes, not to something worse).

**4. `docs/TEST_COVERAGE.md` refreshed.** "dbusmock suites can false-pass locally" is stale in mechanism — since #182 `jvmTest` reports a real `<skipped/>`, and since #186 `:cross_test:linuxX64Test` does not run them at all. Its *advice* still holds and is arguably more important now, so the sentence is restated rather than deleted: those rows are typically **unmeasured** locally, not covered.

## Deliberately not changed

The `assertNativePeerPassed` / `nativePeerFilter` duplication across `cross_test` and `stress_test`. The reviewer verified rather than assumed this: `stress_test` depends on `project(":")` only and has no route to `cross_test`'s test source set, and the two files already duplicated four helpers. Sharing would cost a new module or a test-fixtures publication for two functions. The duplicated count is now **7** — this PR adds `drainedPeerOutput` to the set, so the "if it grows again" trigger below is already tripped and is filed as a follow-up issue rather than being actioned here.

## Evidence — both gates still work, both directions, on two hosts

Re-run after the refactor, because a cleanup that quietly disarms the gate it is cleaning up would be exactly the defect #183 was about. Identical results on two hosts (Arch, JDK 21, under `dbus-run-session`, no python-dbusmock):

| check | result (`<testsuite …>` from fresh JUnit XML) |
| --- | --- |
| interop suite, reverse on — **green** | `tests="14" skipped="1" failures="0" errors="0"`, exit 0 |
| the `@Ignore`d #241 case | present as a `<skipped/>` entry, not vanished, not passed |
| **RED** — native-only assertion broken (`expectedArg` → `expectedArg + 999`) | `tests="1" skipped="0" failures="1" errors="0"`, exit **1**, message `Native peer reported 'nativeClientCanObserveSignalFromJvmPeeer…' as failed: ##teamcity[testFailed …` |
| #186 — no dbusmock, `DBUSMOCK_REQUIRED` unset | `linuxX64Test` = 4 tests (3 + 1), exit 0 — no phantom passes |
| #186 — no dbusmock, **`DBUSMOCK_REQUIRED=1`** | 48 tests, **44 failures**, exit **1** — still fails loudly |
| stress smoke, forward + reverse, `repeat=2` | `tests="10" skipped="0" failures="0" errors="0"`, exit 0 |

`./gradlew ktlintCheck apiCheck` → `BUILD SUCCESSFUL`; `git status --porcelain api/` **empty** — `api/*.api` does not move.

No workflow file is touched.

### DIFF
--- a/cross_test/build.gradle.kts
+++ b/cross_test/build.gradle.kts
@@ -69,10 +69,18 @@ tasks.named<Test>("jvmTest") {
 //  * `DBUSMOCK_REQUIRED` set — the `full-tests-x64` job, which apt-installs python3-dbusmock on
 //    purpose — the suites always run, and the harness itself fails loudly when it cannot start a
 //    peer. So CI can never lose this coverage to the exclusion below.
-//  * otherwise the harness is probed exactly as the native `launchDbusmock` probes it (a session bus
-//    plus an importable `dbusmock` module under `DBUSMOCK_PYTHON` / `python3`), and the suites are
-//    excluded only when it genuinely cannot run — a contributor who *has* dbusmock still runs them
-//    with no configuration.
+//  * otherwise the harness is probed — a session bus plus an importable `dbusmock` module under
+//    `DBUSMOCK_PYTHON` / `python3` — and the suites are excluded only when it cannot run, so a
+//    contributor who *has* dbusmock still runs them with no configuration.
+//
+// Two limits, stated because both fail *open* (back to the phantom passes, not to something worse):
+//  * The probe is weaker than the launch. `launchDbusmock` forks `python3 -m dbusmock --session
+//    [-t <template>] …` and gives it a liveness window; this only checks the module imports. On a
+//    box where `dbusmock` imports but a peer cannot actually start — an older dbusmock missing a
+//    template like `bluez5`, a bus-name claim failure — the suites still run and native still
+//    reports passes that asserted nothing.
+//  * If the class names below drift, the exclusions silently stop matching. Gradle's filter API
+//    offers no "this exclusion matched nothing" signal to guard that with.
 val dbusmockRequired = providers.environmentVariable("DBUSMOCK_REQUIRED").isPresent
 val dbusmockPython = providers.environmentVariable("DBUSMOCK_PYTHON").getOrElse("python3")
 val sessionBusPresent = providers.environmentVariable("DBUS_SESSION_BUS_ADDRESS").isPresent
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/CrossRuntimeInteropSmokeTest.kt
@@ -23,6 +23,8 @@ import java.util.concurrent.atomic.AtomicReference
 import kotlin.concurrent.thread
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertNull
 import kotlin.test.assertTrue
 import org.freedesktop.dbus.annotations.DBusInterfaceName
 import org.freedesktop.dbus.annotations.DBusMemberName
@@ -123,8 +125,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -203,8 +204,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -286,8 +286,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -397,8 +396,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -471,8 +469,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -548,8 +545,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -622,8 +618,7 @@ class CrossRuntimeInteropSmokeTest {
                 "Native peer did not exit in time. Output:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(outputPump, nativeOutput)
             assertNativePeerPassed(process, nativeLog, peerTest)
         } finally {
             process.destroyForcibly()
@@ -699,9 +694,10 @@ class CrossRuntimeInteropSmokeTest {
                 procBuilder.redirectErrorStream(true)
             }.start()
             process = launchedProcess
-            outputPump = thread(start = true, isDaemon = true, name = "cross-native-output") {
+            val pump = thread(start = true, isDaemon = true, name = "cross-native-output") {
                 launchedProcess.inputStream.use { input -> input.copyTo(nativeOutput) }
             }
+            outputPump = pump
 
             assertTrue(
                 launchedProcess.waitFor(30, TimeUnit.SECONDS),
@@ -709,8 +705,7 @@ class CrossRuntimeInteropSmokeTest {
                     "Listen failure:\n${listenFailureDetails()}\nOutput:\n" +
                     nativeOutput.toString(Charsets.UTF_8)
             )
-            outputPump?.join(PUMP_DRAIN_MILLIS)
-            val nativeLog = nativeOutput.toString(Charsets.UTF_8)
+            val nativeLog = drainedPeerOutput(pump, nativeOutput)
             assertNativePeerPassed(
                 launchedProcess,
                 nativeLog,
@@ -800,9 +795,10 @@ class CrossRuntimeInteropSmokeTest {
                 procBuilder.redirectErrorStream(true)
             }.start()
             process = launchedProcess
-            outputPump = thread(start = true, isDaemon = true, name = "cross-native-output") {
+            val pump = thread(start = true, isDaemon = true, name = "cross-native-output") {
                 launchedProcess.inputStream.use { input -> input.copyTo(nativeOutput) }
             }
+            outputPump = pump
 
             assertTrue(
                 launchedProcess.waitFor(30, TimeUnit.SECONDS),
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer:
### Interim: reproduced the PR's own table locally at `efff80b` (fresh worktree, JDK 21, Arch, under `dbus-run-session`, `python3 -c "import dbusmock"` → `ModuleNotFoundError`)

Posting the measurements before the deeper analysis so they are durable. All from **fresh** JUnit XML (`test-results` wiped inside the payload first), not from a `BUILD SUCCESSFUL` line.

| check | `<testsuite …>` / observation | exit |
| --- | --- | --- |
| `:cross_test:jvmInteropTest`, reverse on | `tests="14" skipped="1" failures="0" errors="0"` | 0 |
| the `@Ignore`d #241 case | `nativeClientRoundTripsUnixFdWithJvmPeerOverDirectBus` present, carrying `<skipped/>` — and `skipped="1"`, not 7, so the reverse cases really ran | — |
| `:cross_test:linuxX64Test`, `DBUSMOCK_REQUIRED` unset | `CrossModuleSignalIntegrationTest tests="3"` + `EventLoopStartIdempotencyTest tests="1"` = **4**; the `logger.lifecycle` exclusion notice printed exactly **once** | 0 |
| `:cross_test:linuxX64Test`, **`DBUSMOCK_REQUIRED=1`** | 11 suites summing to **48 tests / 44 failures / 0 errors**; every failure `kotlin.IllegalStateException: DBUSMOCK_REQUIRED is set but no dbusmock peer could be started …` | **1** |
| `:stress_test:jvmInteropStressTest` reverse on, `repeat=2` | `tests="10" skipped="0" failures="0" errors="0"` | 0 |
| `ktlintCheck apiCheck` | `BUILD SUCCESSFUL`; `git status --porcelain` empty (whole tree, `api/` included) | 0 |

So #186 still fires in both directions and the green leg is green. The RED / filter-drift legs are running now; verdict to follow.

monkopedia-reviewer (approved):
## Approve — reviewed at `efff80b`; the cleanup does not disarm either gate, and I broke both to prove it

Fresh-context review in my own worktree (Arch, JDK 21, under `dbus-run-session`, `python3 -c "import dbusmock"` → `ModuleNotFoundError`). The interim comment above carries the green-side numbers; this is the RED side plus the judgement calls. Nothing here rests on the PR body — every number below came out of JUnit XML I generated after wiping `test-results`.

### The risk this PR carries, tested directly

A cleanup that quietly disarms the gate it is cleaning is the exact shape of #183. So `pass → break → RED → revert → clean`, on all three legs of `assertNativePeerPassed`, one case (`nativeClientObservesJvmSignalOverDirectBus`, reverse on):

| probe | `<testsuite …>` | exit | which leg fired |
| --- | --- | --- | --- |
| **native assertion mutated** (`assertEquals(expectedArg, observedSignalValue.value)` → `expectedArg + 999`, `NativeInteropPeerTest.kt:703`) | `tests="1" skipped="0" failures="1" errors="0"` | **1** | the scoped `testFailed` leg |
| **filter drift** — only the JVM-side `peerTest` val renamed (`…OverDirectBusRENAMED`) | `tests="1" skipped="0" failures="1" errors="0"` | **1** | the `testFinished` "measured nothing" leg |
| revert both | `tests="1" skipped="0" failures="0" errors="0"` | 0 | — |

`git status --porcelain` empty after the cycle; no probe-shaped names left in the tree.

The RED message carries the native assertion verbatim, which is what change 2 had to preserve:

```
Native peer reported 'nativeClientCanObserveSignalFromJvmPeerOverDirectBus' as failed:
##teamcity[testFailed name='nativeClientCanObserveSignalFromJvmPeerOverDirectBus'
  message='Expected <1066>, actual <67>.' details='kotlin.AssertionError: …
```

And I confirmed the premise the new marker depends on rather than assuming it: the Kotlin/Native runner emits `testStarted` / `testFailed` / `testFinished` all keyed on the **bare** method name, with the class only on `testSuiteStarted`. So `"##teamcity[testFailed name='$peerTest'"` matches the real output shape, not a guessed one.

The drift leg — the one most at risk from change 2 — still produces the right sentence:

```
Native peer never ran 'nativeClientCanObserveSignalFromJvmPeerOverDirectBusRENAMED', so this case
measured nothing — the filter --ktest_gradle_filter=*NativeInteropPeerTest.…RENAMED* matched no test.
```

### Is scoping `testFailed` to `peerTest` a weakening? No — and I tested the reasoning instead of accepting it

The claim is "a `testFailed` for some *other* test still fails via the exit-code leg." I built that situation directly: **kept the native mutation in place and made the scoped marker un-matchable** (`name=@@nomatch@@`), i.e. a peer that reports a failure the marker does not recognise. Result: `tests="1" failures="1"`, exit **1**, via

```
Native peer 'nativeClientCanObserveSignalFromJvmPeerOverDirectBus' exited non-zero.
… Output:
##teamcity[testSuiteStarted …] ##teamcity[testStarted …] ##teamcity[testFailed …
```

So the fallback is live, and the unrecognised `testFailed` line still reaches the report inside the attached output. Coverage is unchanged; only the headline sentence is less pointed in that hypothetical. And the hypothetical is currently unreachable: the filter is `*NativeInteropPeerTest.<name>*`, so it could admit two tests only if one peer-test name were a prefix of another — I checked all **16** names in `NativeInteropPeerTest` and there are **no prefix pairs**. Net: a strictly more accurate message today, with the same failure coverage.

### The rest of the judgement calls

- **`assertFalse(pump.isAlive)` after `join(timeout)` is the correct check.** `Thread.join(long)` returns nothing, so `isAlive` is the only way to distinguish "finished" from "timed out", and the join supplies the memory barrier for reading the `ByteArrayOutputStream`. The KDoc's "peer has already exited and spawns no children" holds for every call site — each one sits after a successful `waitFor`, and the peer roles don't fork (`launchDbusmock` does, but no peer role runs it).
- **No dead state from the hoist.** `var outputPump: Thread?` is still load-bearing: the `finally` block lives outside the `try` where `val pump` is scoped, and the pump genuinely may not exist if `ProcessBuilder.start()` throws. Hoisting to a local and assigning through is the right way to avoid `!!` here (`nullable-then-force-unwrap`).
- **Call sites: forward −1 line each, reverse net 0 but strictly clearer** (no `?.join`, no separate snapshot line). The +38 net is the new assertion, its message, and two KDoc blocks — so `dry-refactor-should-shrink-callsites` doesn't bite: this PR adds a check, it doesn't claim DRY, and the alternative (an inline `assertFalse` + message at 13 sites) would be much larger.
- **Finding 3's corrected comment is accurate.** `DbusmockHarness.linuxX64.kt` really does `fork()` + `execvp("python3 -m dbusmock --session [-t <template>] …")` and then polls `waitpid(WNOHANG)` 50 × 10 ms — a ~500 ms liveness window — while `dbusmockCanRun()` only runs `python3 -c "import dbusmock"`. Dropping "exactly" was right, and both stated limits do fail *open*: an importable-but-unlaunchable dbusmock returns `dbusmockUnavailable` → `skipTest` → a native pass that asserted nothing, and a drifted class name makes an `excludeTestsMatching` silently match nothing (Gradle only errors on *includes* that match nothing).
- **Finding 4's doc line is true as restated.** `DbusmockHarness.jvm.kt`'s `skipTest` is `assumeTrue(reason, false)`, so JVM records a real skip; and with the harness unable to run, `:cross_test:linuxX64Test` gave me exactly 4 tests with the exclusion notice printed once. "Unmeasured rather than covered" is the honest description.
- **Leaving the cross-module duplication is the right call.** Verified independently: `stress_test` declares only `implementation(project(":"))` and reaches `cross_test` solely through `dependsOn(":cross_test:linkDebugTestLinuxX64")` plus its build-dir path — there is no route to `cross_test`'s *test* source set, so sharing means a new module or a test-fixtures publication.

### CI

15 checks pass at `efff80b` (`full-tests-x64` 6m43s, `arm-runtime-tests`, `static-checks`, all 9 `pr-smoke` legs, both sample builds), 2 skipping by design. `arm-build-test.yaml` still sets `DBUSMOCK_REQUIRED: "1"`, so the exclusion added in #240 cannot cost CI the dbusmock coverage.

---

## Non-blocking findings

1. **`drainedPeerOutput` builds the capture twice, on every call.** `assertFalse(actual, message: String?)` takes an *eager* message, so `sink.toString(Charsets.UTF_8)` runs once for a failure message that is almost never used and once for the return value — and the same expression appears twice in a five-line function. One local removes both:
   ```kotlin
   pump.join(PUMP_DRAIN_MILLIS)
   val output = sink.toString(Charsets.UTF_8)
   assertFalse(pump.isAlive, "Native peer output was still draining after ${PUMP_DRAIN_MILLIS}ms, …:\n$output")
   return output
   ```
   Both files. Trivial, but this is a helper introduced *to be* the clean form.

2. **The PR body's own follow-up trigger is already tripped, and the count is off by one.** It says the duplicated-helper count "is now 6 — if it grows again, a shared test-fixtures source set is the right follow-up." Enumerating `private fun`s present in both files: `nativeTestBinaryPath`, `waitForSocket`, `installNullSafeSaslCollatorWorkaround`, `setStaticField`, `nativePeerFilter`, `assertNativePeerPassed`, **`drainedPeerOutput`** = **7**, because this PR adds the seventh. I still agree it shouldn't be fixed here — but the threshold it sets is met on merge, so the test-fixtures follow-up is worth filing now rather than waiting for a growth that already happened. (Same paragraph: "the seven reverse cases per file" is 7 in `cross_test` and 1 in `stress_test`.)

3. **Latent, not caused here:** if a future `NativeInteropPeerTest` name ever *extends* an existing one, the trailing-wildcard filter admits two cases and the scoped marker degrades that case's diagnostic to the exit-code leg (still red — see the probe above). Worth remembering if someone adds `…OverDirectBusWithFoo` next to `…OverDirectBus`.

None of these blocks. `test_only` → tier 1, CI green at `efff80b`; approving and merging.

Issue #202, opened 2026-08-07
Three false doc claims outside BACKENDS.md/AGENTS.md (Proxy.callMethod blocking, demo-service README version, CLAUDE.md module list)


monkopedia-coder:
Three doc claims verified false by the 2026-08-07 docs audit, in files not covered by #191 (`docs/BACKENDS.md`) or #192 (`AGENTS.md`). Each was checked against the code, not read for plausibility.

## 1. `Proxy.callMethod`'s blocking claim is native-only, stated unconditionally

`Proxy.kt:94-98`:
> "While blocking, other concurrent operations (in other threads) on the underlying bus connection are stalled until the call returns."

**True on native. False on JVM.** Traced: `Proxy.jvm.kt:21` → `WireDbusBackend.callMethod` (`:293-330`) is either an in-process dispatch or an awaited wire reply — no connection-wide lock, and the reader thread keeps running throughout.

**Consequence is unusual and worth stating: the doc is over-conservative, so the failure mode is a consumer over-engineering around a stall that does not happen on their backend** — extra threads, extra connections, avoidance of a pattern that would have been fine. That is a cheaper failure than the reverse, which is why this is ranked below the others, but it is still a false statement about behaviour in public KDoc.

Same provenance as the `MethodInvoker.timeout` doc behind #178/#179: inherited from sdbus-c++, accurate for the native implementation, never re-examined when the JVM backend was written.

## 2. `samples/demo-service/README.md:30-37` contradicts the rest of the repo about what is published

> "That API does **not** exist in the artifact published to Maven Central (**1.0.0**). … the version coordinate there (`1.0.1`) is only a placeholder."

**False.** 1.0.1 is released — `gradle.properties` has `version=1.0.1`, and `README.md:64,98` instructs users to depend on 1.0.1 as a published artifact. Two documents in one repository disagree about whether 1.0.1 exists, and the sample's README is the one a new user reads first.

## 3. `CLAUDE.md:9` module list is incomplete

Lists root, `:codegen`, `:plugin`, `:cross_test`, `:stress_test`, `samples/bluez-scan`. `settings.gradle.kts` also includes **`compile_test`**, and **`samples/demo-service`** exists as a composite build.

`compile_test` is not incidental — it is the module that proves generated code compiles, and it is the guard PR #189 leaned on for malformed KDoc. An agent reading `CLAUDE.md` to learn the repo's shape does not learn that module exists.

[...]

PR #244, opened 2026-08-08
docs: correct stale connection-factory and use-after-release claims, and list every module (#191, #202)

### LINKED ISSUE #191: `docs/BACKENDS.md` describes a JVM fd-factory mechanism that was deleted 7 weeks ago, and gets the factory counts wrong
Found by the Stream-2 (docs & contracts) audit pass. `docs/BACKENDS.md` is the repo's cross-backend
contract document. It was already wrong once about timeouts (corrected in #181). The
**"Connection factories"** section contains four separately-checkable claims; **three of them are
false**, and one of the false ones is load-bearing for a piece of dead code.

All checks below are against `main` @ `c36abb5`.

## 1. The `@Deprecated(level = ERROR)` JVM actuals do not exist — FALSE

`docs/BACKENDS.md:289-291`:

> - The two **fd-based factories** are native-only. The JVM actuals are marked
>   `@Deprecated(level = ERROR)` (`src/jvmMain/.../Connection.jvm.kt`), so calling them from
>   JVM-compiled code is a **compile-time error**;

Repeated at `docs/BACKENDS.md:371-373`:

> 1. **No fd-based connections**: `createDirectBusConnection(fd)` / `createServerBusConnection(fd)`
>    are compile-time errors on JVM (`@Deprecated(ERROR)` stubs).

Check:

```
$ grep -rn "Deprecated" --include='*.kt' src/jvmMain/ | grep -v WireServe | wc -l
0
$ grep -rn "createServerBusConnection" --include='*.kt' src/jvmMain/ | wc -l
0
```

There is no `@Deprecated` annotation anywhere in `src/jvmMain`, and no `createServerBusConnection`
symbol at all. **PR #126 (`7d247c4`, 2026-06-17) removed them** — its own commit message says so:

> `createDirectBusConnection(UnixFd)` and `createServerBusConnection(UnixFd)` **were**
> `@Deprecated(ERROR)` stubs on JVM; they are **now plain native-only funs** in
> `Connection.native.kt` … removed from commonMain and the JVM actuals.

The doc paragraph predates that (blame: `2152f7a1`, 2026-06-11) and was never updated. Calling
those factories from JVM code is an *unresolved reference*, not a deprecation error.

## 2. The runtime guard the doc cites has zero producers — FALSE in effect

`docs/BACKENDS.md:291-293`:

> if invoked anyway (e.g. reflectively), the JVM backend rejects the fd-based bus types
> (`JvmBusType.DIRECT_FD`/`SERVER_FD`, in `WireDbusBackend.createConnection`).

The guard exists (`WireDbusBackend.kt:63-65`), and `JvmBusType.DIRECT_FD`/`SERVER_FD` exist
(`JvmDbusBackend.kt:30-31`). But **nothing ever constructs those two enum values** — the only
`createConnection` call sites are the eight factories in `Connection.jvm.kt`, which pass
`DEFAULT`/`SYSTEM`/`SESSION`/`SESSION_ADDRESS`/`DIRECT_ADDRESS`. With the JVM symbols deleted in #126
there is no longer any entry point, reflective or otherwise, that can reach them.

So the guard at `WireDbusBackend.kt:63-65`, the `null` branch at `WireDbusBackend.kt:75`, and the two
enum constants are unreachable, and this doc sentence is the stated reason they are still there.
(They date from the `COVERAGE_TODO.md` item "Align unsupported JVM connection kinds with native error
semantics (avoid silent fallback for `DIRECT_FD`/`SERVER_FD`)", which was checked off before #126.)

## 3. "10 declared in the common API plus one native-only" — FALSE

`docs/BACKENDS.md:275-277`:

> There are 11 `create*Connection` entry points — 10 declared in the common API
> (`src/commonMain/.../Connection.kt`) plus one declared only in the native source set

```
$ grep -c "^expect fun create[A-Za-z]*BusConnection" src/commonMain/kotlin/com/monkopedia/sdbus/Connection.kt
8
```

It is **8 in common + 3 native-only** (`createRemoteSystemBusConnection`,
`createDirectBusConnection(fd)`, `createServerBusConnection(fd)`), which still totals 11.

## 4. The matrix row "9 of 11" for JVM — FALSE

`docs/BACKENDS.md:41`:

> | Connection factories (11 total) | 9 of 11 | 11 of 11 | fd-based factories are native-only |

Counted from the checked-in API dumps, which are the authority here:

```
$ grep -c "fun create[A-Za-z]*BusConnection" api/sdbus-kotlin.api
8
$ grep -o "com.monkopedia.sdbus/create[A-Za-z]*BusConnection([^)]*)" api/sdbus-kotlin.klib.api | sort -u | wc -l
11
```

JVM ships **8 of 11**, not 9. (The native 11 is correct.)

## 5. Separately: the `currentlyProcessedMessage` divergence bullet is stale — FALSE

`docs/BACKENDS.md:404`, in "same-process / direct-connection edges … documented rather than matched":

> - `currentlyProcessedMessage` is not guarded against use after `release()` on JVM (native is).

This was fixed in #172 (`e3eda9d`). `WireDbusObject.kt:352-356`:

```kotlin
override fun currentlyProcessedMessage(): Message {
    // Native's ObjectImpl reads through to the connection, which rejects use after release()
    require(!connection.isReleased()) { "Connection has already been released" }
```

and it is pinned cross-backend by
`FailurePathParityTest.currentlyProcessedMessageAfterReleaseThrowsOnBothBackends`
(`src/commonTest/.../FailurePathParityTest.kt:408-418`), which asserts all three of
`connection`/`obj`/`proxy` `currentlyProcessedMessage` throw on both backends. BACKENDS.md has been
edited twice since #172 (#174, #181) without correcting this line.

## Why this matters beyond tidiness

`docs/BACKENDS.md` is what both consumers and agents read to learn what the two backends promise.
Items 1 and 2 together mean a reader is told a compile-time guard and a runtime guard both protect
the fd path on JVM, when the first does not exist and the second is unreachable. That is the same
failure mode as #179/#178: a doc describing a safety property that nobody re-checked.

## Suggested fix (doc-only, except optionally item 2)

Correct §"Connection factories" (5 claims above). For item 2, either delete the now-unreachable
`DIRECT_FD`/`SERVER_FD` enum values + guard, or keep them and change the doc to say they are a
defensive remnant with no reachable caller — but do not keep the current sentence, which asserts a
reachable path that does not exist.

## Verification when worked

Each claim above is a one-line grep; re-run the five commands quoted. Nothing here needs a bus.

monkopedia-coder:
Fixes #191
Refs #202

Prose-only. `git diff --stat` is `CLAUDE.md | 5 +++-` and `docs/BACKENDS.md | 36 +++++-----`; no `.kt`, no `.kts`, no `api/*.api`. `./gradlew ktlintCheck apiCheck` → BUILD SUCCESSFUL.

Every claim below was re-measured against `main` @ `79bdbe5`, not read for plausibility. Command output is quoted where it settles the point.

---

### 1. The `@Deprecated(level = ERROR)` JVM actuals do not exist

**Was** (`docs/BACKENDS.md:290-294`):
> The JVM actuals are marked `@Deprecated(level = ERROR)` (`src/jvmMain/.../Connection.jvm.kt`), so calling them from JVM-compiled code is a **compile-time error**

**Contradiction.** With a positive control, because the shell's `grep` silently skips binary-flagged files:

```
$ /usr/bin/grep -ra "fun " --include='*.kt' src/jvmMain/ | wc -l
438                                        # positive control: the sweep does see this tree
$ /usr/bin/grep -ran "Deprecated" --include='*.kt' src/jvmMain/
WireServe.kt:359: "  <annotation name=\"org.freedesktop.DBus.Deprecated\" value=\"true\"/>\n"
WireServe.kt:546/554/561: deprecated = item.isDeprecated
$ /usr/bin/grep -ran "createServerBusConnection" --include='*.kt' src/jvmMain/
                                           # (no output)
```

The only `Deprecated` occurrences in `src/jvmMain` are introspection-XML strings in `WireServe.kt`. There is no `@Deprecated` annotation and no `createServerBusConnection` symbol. PR #126 removed them; `src/nativeMain/.../Connection.native.kt:96,122` now declares both as plain non-`actual` top-level funs, and `Connection.kt` has no `expect` for either. A JVM call is an **unresolved reference**.

**Now:** the bullet says they are native-only funs since #126 with no commonMain `expect` and no JVM actual, so naming them from JVM-compiled code is an unresolved reference — with one parenthetical noting the `@Deprecated(ERROR)` form is historical, so a reader who saw the old text knows it was replaced rather than that the doc drifted.

### 2. The runtime guard the doc cites has zero producers

**Was** (`docs/BACKENDS.md:292-294`):
> if invoked anyway (e.g. reflectively), the JVM backend rejects the fd-based bus types (`JvmBusType.DIRECT_FD`/`SERVER_FD`, in `WireDbusBackend.createConnection`).

**Contradiction.** Cross-checked two ways (`/usr/bin/grep -rn`, and an `awk` sweep per file in case of a binary-flag skip) — identical results, four hits total:

```
WireDbusBackend.kt:63:  if (fd != null || busType == JvmBusType.DIRECT_FD || busType == JvmBusType.SERVER_FD) {
WireDbusBackend.kt:75:      JvmBusType.DIRECT_FD, JvmBusType.SERVER_FD -> null
JvmDbusBackend.kt:30:   DIRECT_FD,
JvmDbusBackend.kt:31:   SERVER_FD
```

Two declarations, the guard, and the dead `null` branch. **No producer.** Every `createConnection` call site is one of the eight factories in `Connection.jvm.kt:40,44,53,57,66,70,79,88`, and each passes `DEFAULT`/`SYSTEM`/`SESSION`/`SESSION_ADDRESS`/`DIRECT_ADDRESS` with `fd = null`. `JvmDbusBackend` is `internal`, so there is no out-of-repo implementor either. The "reflectively" path the doc describes does not exist.

**Now:** the doc takes the issue's *second* option — it names the check a "defensive remnant with no reachable caller" and tells the reader the missing declaration, not the runtime check, is what stops them. **I did not delete the enum values or the guard**: that is code, and this PR is scoped to prose. If the owner prefers the delete, it is a separate one-file change.

### 3. "10 declared in the common API plus one native-only"

**Was** (`docs/BACKENDS.md:276-278`): "10 declared in the common API … plus one declared only in the native source set".

```
$ /usr/bin/grep -c "^expect fun create[A-Za-z]*BusConnection" src/commonMain/.../Connection.kt
8
$ /usr/bin/grep -rn "^fun create[A-Za-z]*BusConnection" src/nativeMain/.../Connection.native.kt
71: createRemoteSystemBusConnection(host)
96: createDirectBusConnection(fd: UnixFd)
122: createServerBusConnection(fd: UnixFd)
```

**Now:** "8 declared in the common API … plus 3 declared only in the native source set". Still 11 total, and the factory table below it was already correct (2+2+3+1+1+1+1 = 11).

### 4. The matrix row "9 of 11" for JVM

**Was** (`docs/BACKENDS.md:41`): `| Connection factories (11 total) | 9 of 11 | 11 of 11 | fd-based factories are native-only |`

From the checked-in API dumps, which are the authority:

```
$ /usr/bin/grep -c "fun create[A-Za-z]*BusConnection" api/sdbus-kotlin.api
8
$ /usr/bin/grep -o "com.monkopedia.sdbus/create[A-Za-z]*BusConnection([^)]*)" api/sdbus-kotlin.klib.api | sort -u | wc -l
11
```

**Now:** `8 of 11`. The note also names the third missing factory — the old note said only "fd-based factories are native-only", which accounts for 2 of the 3 absences and is part of why `9` looked self-consistent.

### 5. The `currentlyProcessedMessage` divergence bullet is stale

**Was** (`docs/BACKENDS.md:405`, under "documented rather than matched"):
> `currentlyProcessedMessage` is not guarded against use after `release()` on JVM (native is).

**Contradiction** — `WireDbusObject.kt:352-356`:

```kotlin
override fun currentlyProcessedMessage(): Message {
    // Native's ObjectImpl reads through to the connection, which rejects use after release()
    // (ConnectionImpl.checkNotReleased). Parity #141.
    require(!connection.isReleased()) { "Connection has already been released" }
```

Pinned cross-backend by `FailurePathParityTest.currentlyProcessedMessageAfterReleaseThrowsOnBothBackends` (`src/commonTest/.../FailurePathParityTest.kt:408`).

**Now:** removed from the divergence list, replaced by a short note naming #172 and the pinning test — so the next reader learns it was closed rather than rediscovering it.

---

### Item 3 (CLAUDE.md module list) — CONFIRMED, FIXED

`settings.gradle.kts` includes `codegen`, **`compile_test`**, `cross_test`, `plugin`, `stress_test`; `CLAUDE.md:9` listed everything but `compile_test`, and named only one of the two samples.

`compile_test` earns more than a list entry, so it gets a sentence. `compile_test/src/nativeMain/kotlin/TestFiles` is a **tracked symlink** to `codegen/src/test/resources`:

```
$ ls -la compile_test/src/nativeMain/kotlin/
TestFiles -> ../../../../codegen/src/test/resources
$ git ls-files compile_test
compile_test/build.gradle.kts
compile_test/src/nativeMain/kotlin/TestFiles
```

so a linuxX64 compile of the module type-checks every checked-in codegen golden against the real library. I verified it is genuinely CI-covered rather than assuming it — CI runs `./gradlew allTests test` (`arm-build-test.yaml:81`), and:

```
$ ./gradlew allTests --dry-run | grep compile_test
:compile_test:compileKotlinLinuxX64 SKIPPED
:compile_test:linkDebugTestLinuxX64 SKIPPED
:compile_test:allTests SKIPPED          # --dry-run marks everything SKIPPED; the point is they are IN the graph
```

The samples are also described correctly now: they are **separate Gradle builds** using `includeBuild("../..")` (`samples/*/settings.gradle.kts`), not subprojects — `settings.gradle.kts` does not include them at all.

**Maintenance section: yes, added** (the issue asked for it explicitly, and without it this recurs). New bullet requires the `Modules:` paragraph to enumerate every `include(...)` plus the composite-build samples, and to be updated **together with** `AGENTS.md`'s "Project Structure & Module Organization" list — which, checked independently, is already complete (`AGENTS.md:18-24` lists `compile_test/` and both samples). CLAUDE.md was the only drifted copy.

### Item 2 (demo-service README version claim) — ALREADY FIXED by #222, no change made

The issue quotes:
> "That API does **not** exist in the artifact published to Maven Central (**1.0.0**). … the version coordinate there (`1.0.1`) is only a placeholder."

The first sentence is **gone**. `git log -- samples/demo-service/README.md` → `67ac247` ("ci: compile samples/ in CI and align their Gradle wrappers with the root (#222)") rewrote that section; `/usr/bin/grep -n "1\.0\.0" samples/demo-service/README.md` returns nothing.

The surviving "the version coordinate there (`1.0.1`) is only a placeholder" is **not a defect**, and I deliberately left it. In context (`README.md:30-40`) it is a statement about *composite-build substitution*, and it is accurate — `includeBuild("../..")` substitutes on `group:name` and ignores the version, so the literal in `build.gradle.kts:43` is inert while the sample is built in place. The very next paragraph already says the literal becomes real when the directory is copied out, and that the release checklist keeps it in step. Nothing here says 1.0.1 is unpublished. Rewording it would be churn.

### Item 1 (`Proxy.callMethod` blocking claim) — CONFIRMED TRUE, but SPLIT OUT

I verified the claim and then did **not** fix it, because the text lives in `src/commonMain/kotlin/com/monkopedia/sdbus/Proxy.kt:94-102` and `commonMain` KDoc is being worked concurrently under #208/#209/#210. Editing it from here risks a conflicting edit in a file another agent may be holding.

The claim is real. `Proxy.kt:94-98` states unconditionally:
> "While blocking, other concurrent operations (in other threads) on the underlying bus connection are stalled until the call returns."

Traced on JVM: `Proxy.jvm.kt:21` → `WireDbusBackend.callMethod` (`:295-404`), which either dispatches in-process (`:345-365`) or calls `callRemote` (`:538`) → `DBusWireConnection.callBlocking` (`:389-410`). `callBlocking` does `future.get(timeout)` — it parks **only the calling thread**. The single `synchronized(writeLock)` on that path (`DBusWireConnection.kt:270`) is held for the duration of writing the message bytes, **not** across the reply wait, and the reader thread keeps dispatching throughout. There is no connection-wide lock.

So the sentence is native-accurate (sd-bus does serialize the connection) and false on JVM. As the issue notes, the failure direction is over-conservatism — a consumer avoiding a pattern that would have been fine — which is why it is the lowest-severity of the three.

**#202 therefore gets `Refs`, not `Fixes`**: item 1 is unresolved and still needs an owner or the KDoc agent to take it.

---

## What I ran, and what I did not

**Ran:** `./gradlew ktlintCheck apiCheck` (BUILD SUCCESSFUL, 16s); `./gradlew allTests --dry-run` (task-graph check only, nothing executed); all the greps and `awk` sweeps quoted above; `git status --short api/` → empty.

**Did not run:** any test suite, any bus-backed test, and **`:dokkaGeneratePublicationHtml`** — deliberately. Dokka's only prose input is `dokka/moduledoc.md` (`build.gradle.kts:262`, `includes.from(rootProject.file("dokka/moduledoc.md"))`); neither file I touched is a Dokka input, so a warning count from it would say nothing about this diff. Nothing here can move Dokka output.

## Guard rails honoured

- **`docs/BACKENDS.md:35`** — the `PropertiesChanged` ✅/✅ emission row — is **byte-identical to `origin/main`**, verified by `diff <(awk 'NR==35' docs/BACKENDS.md) <(git show origin/main:docs/BACKENDS.md | awk 'NR==35')`. Amending it is option 1 of the still-open #193 and is not mine to decide.
- **`docs/BACKENDS.md:45`** (`Method.NoReply`, corrected by #221) and **`AGENTS.md`** (verification command, corrected by #224) — untouched; `AGENTS.md` is not in this diff at all.
- **No code, no behaviour.** The one code change #191 offers as an alternative (deleting the unreachable `DIRECT_FD`/`SERVER_FD` guard) was left undone by design; see item 2.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,7 +6,9 @@ Guidance for agents working on sdbus-kotlin.
 
 A Kotlin Multiplatform D-Bus client (a port of sdbus-c++), targeting **jvm + linuxX64 + linuxArm64**. The native targets wrap `sd-bus` via cinterop; the JVM target is backed by an owned junixsocket connection with a pure-Kotlin marshaller and dispatcher (since 0.5.0 — no dbus-java, no native code). It also ships a code generator (`:codegen`, XML → Kotlin BlueZ proxies/adaptors) and a Gradle plugin (`:plugin`, id `com.monkopedia.sdbus.plugin`).
 
-Modules: root (the library), `:codegen`, `:plugin`, `:cross_test`, `:stress_test`, `samples/bluez-scan`.
+Modules (subprojects of the root build, per `settings.gradle.kts`): root (the library), `:codegen`, `:compile_test`, `:cross_test`, `:plugin`, `:stress_test`. Plus two **separate** Gradle builds that are not subprojects — `samples/bluez-scan` and `samples/demo-service` — each composing the library in with `includeBuild("../..")`.
+
+`:compile_test` is easy to miss and is the guard that the generated code actually compiles: its source directory is a **symlink** (`compile_test/src/nativeMain/kotlin/TestFiles` → `codegen/src/test/resources`), so a linuxX64 compile of that module type-checks every checked-in codegen golden against the real library — this is what caught the malformed signal decoder in #237, which `:codegen`'s own in-process `CompilationIntegrationTest` had missed. CI covers it because the root `allTests` reaches `:compile_test:compileKotlinLinuxX64`; run it directly with `./gradlew :compile_test:compileKotlinLinuxX64`. It is deliberately excluded from `ktlintCheck` and `apiCheck` (see `build.gradle.kts`).
 
 ## Building & verifying
 
@@ -29,4 +31,5 @@ Modules: root (the library), `:codegen`, `:plugin`, `:cross_test`, `:stress_test
   - The **Kotlin version** badge is hardcoded — bump it manually when the Kotlin version changes.
   - The **Build** badge tracks `.github/workflows/arm-build-test.yaml`; keep the workflow path correct if CI is renamed.
   - Maven Central has no public *download* count, so there is no downloads badge (only the namespace owner can see download stats in the Central Portal).
+- **Module list:** the `Modules:` paragraph under "What this is" must enumerate every `include(...)` in `settings.gradle.kts` plus the composite-build samples. When a module is added, removed or renamed, update it there **and** the matching list in `AGENTS.md` ("Project Structure & Module Organization") — nothing else prompts it, and `:compile_test` sat unlisted here for exactly that reason (#202).
 - Generated BlueZ proxy fixtures and the BCV `api/*.api` dumps are checked in. Regenerate the codegen goldens with `./gradlew :codegen:regenerateGoldens` after an intentional generator change, and the `api/*.api` dumps with `./gradlew apiDump` when the public surface changes; review the diff and commit it. `:codegen:test` only ever *compares* the goldens — it can never write them.
--- a/docs/BACKENDS.md
+++ b/docs/BACKENDS.md
@@ -38,7 +38,7 @@ Every row below is pinned by tests on current `main`, not by intention:
 | **Serving exported objects to external processes** (incoming method calls, `Properties.Get`/`Set`/`GetAll`, `GetManagedObjects`) | ✅ | ✅ | **#90 closed (0.5.0):** the JVM backend serves incoming calls over its owned wire connection (`WireServe`), so an external client (busctl, another process) reaches a JVM-exported object exactly as on native — see "Server-side object export". Same-JVM calls still take an in-process shortcut (`JvmStaticDispatch`) |
 | `UnixFd` type semantics (dup constructor vs `adopt`, `release` closes) | ✅* | ✅ | *JVM dup needs junixsocket native support |
 | Unix FD passing over the wire | ⚠️ untested | ✅ | JVM has the conversion path but no independent-peer test |
-| Connection factories (11 total) | 9 of 11 | 11 of 11 | fd-based factories are native-only |
+| Connection factories (11 total) | 8 of 11 | 11 of 11 | the two fd-based factories and `createRemoteSystemBusConnection` are native-only; counted from the checked-in `api/*.api` dumps |
 | Behavior when the bus is unreachable | ✅ throws `Error` | ✅ throws `Error` | JVM fixed in #81; the in-process stub backend is an explicit internal test opt-in only |
 | Event loop (`startEventLoop`/`stopEventLoop`) | no-op (always running) | required for dispatch; `createObject`/`createProxy` auto-start it | Same calling pattern works on both; `startEventLoop` is idempotent (#114), and each native connection gets its own loop thread (#128) |
 | Strict deserialization (signature mismatch rejected) | ✅ | ✅ | Same `System.Error.ENXIO` error |
@@ -273,9 +273,9 @@ Backend differences:
 
 ## Connection factories
 
-There are 11 `create*Connection` entry points — 10 declared in the common API
-(`src/commonMain/.../Connection.kt`) plus one declared only in the native source set
-(`src/nativeMain/.../Connection.native.kt`):
+There are 11 `create*Connection` entry points — 8 declared in the common API
+(`src/commonMain/.../Connection.kt`, hence available on both backends) plus 3 declared only in
+the native source set (`src/nativeMain/.../Connection.native.kt`):
 
 | Factory | JVM | Native |
 | --- | --- | --- |
@@ -284,15 +284,18 @@ There are 11 `create*Connection` entry points — 10 declared in the common API
 | `createSessionBusConnection()` / `(name)` / `(address)` | ✅ | ✅ |
 | `createRemoteSystemBusConnection(host)` | ❌ native-only (not declared) | ✅ |
 | `createDirectBusConnection(address)` | ✅ | ✅ |
-| `createDirectBusConnection(fd: UnixFd)` | ❌ native-only | ✅ |
-| `createServerBusConnection(fd: UnixFd)` | ❌ native-only | ✅ |
+| `createDirectBusConnection(fd: UnixFd)` | ❌ native-only (not declared) | ✅ |
+| `createServerBusConnection(fd: UnixFd)` | ❌ native-only (not declared) | ✅ |
 
-- The two **fd-based factories** are native-only. The JVM actuals are marked
-  `@Deprecated(level = ERROR)` (`src/jvmMain/.../Connection.jvm.kt`), so calling them from
-  JVM-compiled code is a **compile-time error**; if invoked anyway (e.g. reflectively), the
-  JVM backend rejects the fd-based bus types (`JvmBusType.DIRECT_FD`/`SERVER_FD`, in
-  `WireDbusBackend.createConnection`).
+- The two **fd-based factories** are native-only: since #126 they are plain top-level funs in
+  `Connection.native.kt` with no `expect` declaration in commonMain and no JVM actual, so naming
+  them from JVM-compiled code is an **unresolved reference**. (They were briefly
+  `@Deprecated(level = ERROR)` JVM stubs before #126; that is no longer how they are guarded.)
   On native they adopt the descriptor as documented in their KDoc.
+  `WireDbusBackend.createConnection` still rejects `JvmBusType.DIRECT_FD`/`SERVER_FD`, but that
+  is a **defensive remnant with no reachable caller** — nothing constructs either enum value now
+  that the JVM symbols are gone, so treat the missing declaration, not the runtime check, as the
+  thing that stops you.
 - A **brokerless direct connection** has no daemon to run `Hello` against, so neither backend gets
   a bus-assigned unique name for it. The JVM backend synthesizes one **per connection**
   (`":jvm-wire-<n>"`) — the in-process registries are keyed by it, so two direct connections in one
@@ -370,7 +373,8 @@ the API docs will see it.
 ## JVM limitations at a glance
 
 1. **No fd-based connections**: `createDirectBusConnection(fd)` / `createServerBusConnection(fd)`
-   are compile-time errors on JVM (`@Deprecated(ERROR)` stubs).
+   are not declared in the JVM API at all (native-only funs since #126), so a JVM-compiled call is
+   an unresolved reference — see "Connection factories".
 2. **Raw-fd semantics depend on junixsocket native support**: without it, `UnixFd`
    duplication/closing degrade (no dup, no close). Wire fd passing is implemented but not
    CI-verified against an independent peer.
@@ -402,6 +406,10 @@ matched (they don't affect cross-process usage):
   where native returns an error.
 - A same-process `UnixFd` argument is passed by reference (not dup'd) on the local short-circuit,
   where native dups the descriptor.
-- `currentlyProcessedMessage` is not guarded against use after `release()` on JVM (native is).
+
+(`currentlyProcessedMessage` used to be listed here as unguarded against use after `release()` on
+JVM. That was closed in #172 — `WireDbusObject.currentlyProcessedMessage` now reads through to
+`connection.isReleased()` — and is pinned cross-backend by
+`FailurePathParityTest.currentlyProcessedMessageAfterReleaseThrowsOnBothBackends`.)
 
 These are tracked as post-1.0 items; none is on the cross-process path.


monkopedia-reviewer (approved):
## Approve

Independent fresh-context review from a clean worktree off `origin/docs/fix-191-202` (`560c57f`). I re-measured all eight assertions against the tree myself rather than reading the PR body for plausibility. **All eight confirm; none disputed.** Guard rails all respected.

Every sweep below used `/usr/bin/grep -ra` with a positive control **and** an independent `awk` per-file pass over `git ls-files`, because ugrep silently skips binary-flagged files in `src/jvmMain/.../jvmdbus/`. Both methods agreed on every count.

---

### #191 — `docs/BACKENDS.md`

**1. No `@Deprecated(level = ERROR)` fd stubs in `src/jvmMain` — CONFIRMED.**
Positive control: 438 `fun ` hits in `src/jvmMain`, so the sweep reaches the tree. The only `Deprecated` occurrences are four introspection-XML strings in `WireServe.kt:359,546,554,561` — no annotation. The only `create*BusConnection` symbol in `src/jvmMain` is `Connection.jvm.kt:87 actual fun createDirectBusConnection(address: String)`; neither `UnixFd` overload exists. And **checked the second half too**: `Connection.kt` declares exactly 8 `expect fun create…` (lines 222–300), none of them fd-taking. So a JVM call really is an unresolved reference, not a deprecation error. `git blame` puts the fd funs' current form at `Connection.native.kt:96,122` as plain non-`actual` top-level funs.

**2. `DIRECT_FD`/`SERVER_FD` have no producer; guard unreachable — CONFIRMED, and the new wording is honest.**
Exactly four hits repo-wide across all tracked `.kt`/`.kts`: the two enum constants (`JvmDbusBackend.kt:30-31`), the guard (`WireDbusBackend.kt:63`), and the dead `null` branch (`:75`). I enumerated **every** `createConnection` call site: the 8 factories at `Connection.jvm.kt:40,44,53,57,66,70,79,88` pass `DEFAULT`/`SYSTEM`/`SESSION`/`SESSION_ADDRESS`/`DIRECT_ADDRESS` with `fd = null` in all eight; the only other references are `JvmFactoryMockkTest`'s mock verifications, which also never name the fd values. Both `JvmBusType` and `JvmDbusBackend` are `internal`, so there is no out-of-repo implementor or producer either.

On the choose-to-document-rather-than-delete call: the new text (`BACKENDS.md:295-298`) calls it "a **defensive remnant with no reachable caller**", says "nothing constructs either enum value now that the JVM symbols are gone", and redirects the reader to the missing declaration as the actual barrier. That satisfies #191's own second option and asserts no reachable path. Keeping the code out of a prose-only PR is the right scoping.

**3. 8 common + 3 native = 11 — CONFIRMED.**
`Connection.kt`: 8 `^expect fun create…BusConnection`. `Connection.native.kt`: 8 `actual` + 3 plain (`createRemoteSystemBusConnection:71`, `createDirectBusConnection(fd):96`, `createServerBusConnection(fd):122`). The factory table below sums to 11 (2+2+3+1+1+1+1), so the corrected prose and the table now agree.

**4. JVM ships 8 of 11 — CONFIRMED from the API dumps, which are the authority.**
`api/sdbus-kotlin.api:43-50` = exactly 8 public factories. `api/sdbus-kotlin.klib.api` = 2+2+1+1+3+2 = 11. The new note also names the third absence (`createRemoteSystemBusConnection`), which is what let the old `9 of 11` look self-consistent.

**5. `currentlyProcessedMessage` divergence is closed and pinned — CONFIRMED, removal correct.**
Guard present at `WireDbusObject.kt:352-356` (`require(!connection.isReleased())`). `git blame` on those lines → `e3eda9d fix(jvm): guard currentlyProcessedMessage against use after release (#172)`, merged 2026-07-31, so the `#172` attribution is right. `FailurePathParityTest.kt:408-419` is in `commonTest` (runs on both backends) and asserts `assertFailsWith<IllegalArgumentException>` on all three of `connection`/`obj`/`proxy` `currentlyProcessedMessage` after `connection.release()` — so the row really is pinned, not just fixed.

I also swept every tracked `.md` for leftovers (`9 of 11`, `Deprecated(ERROR)`, `10 declared`, `not guarded against use after`): the only surviving hit is the new intentional historical parenthetical at `BACKENDS.md:293`. No stale copy elsewhere.

---

### #202

**6. `compile_test` was missing, and the new text is accurate — CONFIRMED on every sub-claim.**
- `settings.gradle.kts` includes `codegen`, `compile_test`, `cross_test`, `plugin`, `stress_test` — the new list matches exactly, in that order, and correctly separates the two samples as non-subprojects (`samples/*/settings.gradle.kts` both `includeBuild("../..")`; neither is in `settings.gradle.kts`).
- Symlink is **tracked**: `git ls-files -s` → `120000 … compile_test/src/nativeMain/kotlin/TestFiles`, resolving to `../../../../codegen/src/test/resources`. Mode 120000 is the proof, not the `ls`.
- **I re-ran the dry-run myself** (`./gradlew allTests --dry-run`) rather than trusting the quote: `:compile_test:compileKotlinLinuxX64`, `:linkDebugTestLinuxX64`, `:linuxX64Test`, `:compile_test:allTests` are all in the graph, and CI's `arm-build-test.yaml:81` runs `allTests test`. So the CI-coverage claim holds.
- The `#237` provenance claim checks out: #237's body shows `:compile_test:compileKotlinLinuxX64` producing the red `Type arguments are not allowed` errors, and states `codegen`'s `CompilationIntegrationTest` (which exists, `codegen/src/test/kotlin/CompilationIntegrationTest.kt`) missed it because its one XML declares a two-arg signal. Accurate as written.
- The exclusion claim is right too: `build.gradle.kts:37` `ignoredProjects.addAll(listOf("compile_test", …))` and `:354 if (name == "compile_test") return@allprojects` before applying ktlint.

**The new maintenance bullet does close the recurrence gap.** #202 asked for exactly this ("nothing prompts an update when a module is added"), and the bullet names the trigger (any `include(...)` change), the two places to update, and the cross-reference to `AGENTS.md`. I verified `AGENTS.md:20,24` is genuinely already complete (`compile_test/`, both samples) — CLAUDE.md was the only drifted copy, so the bullet is pointing at a real second location rather than inventing one.

**7. Already fixed by #222; no change was correct — CONFIRMED.**
`/usr/bin/grep "1\.0\.0" samples/demo-service/README.md` → nothing; `git log` on that file → `67ac247 (#222)` rewrote the section. On the surviving sentence: I read the full context (`README.md:28-40`) and agree it is not a defect. It is scoped to composite-build substitution ("substitution matches on `group:name` and ignores the version — so the version coordinate there (`1.0.1`) is only a placeholder"), which is true, and the very next paragraph says the literal becomes real when the directory is copied out. Nothing there claims 1.0.1 is unpublished. Rewording would be churn. Right call to leave it.

**8. The `Proxy.callMethod` blocking claim is genuinely false on JVM — CONFIRMED by tracing it end to end.**
`Proxy.kt:94-98` states unconditionally that "other concurrent operations (in other threads) on the underlying bus connection are stalled until the call returns". Trace: `Proxy.jvm.kt:21` → `WireDbusBackend.callMethod` (`:293`/`:295`) → either in-process dispatch (`:345-365`) or `callRemote` (`:538`) → `DBusWireConnection.callBlocking` (`:389-410`).

Specifically on the two things that decide it:
- **No lock is held across the reply wait.** `callBlocking` does `dispatchCall(request)` then `future.get(timeoutMillis, …)` (`:393,398`). The only mutual exclusion on the send path is `synchronized(writeLock)` at `:270`, entirely inside `writeMessage`, scoped to the codec write (plus the SCM_RIGHTS attach) — it is released before the wait begins. The class KDoc says as much at `:86`: "Writes are serialized under `writeLock`; any thread may call out concurrently."
- **The reader thread keeps running.** It is what completes the `pending[serial]` future, so it cannot be the thing that is stalled; `serveWorkers.beginNestedBlock()`/`endNestedBlock()` (`:396,406`) exist precisely to keep the serve pool runnable while a worker parks. Only the calling thread parks.

**On deliberately splitting it out:** defensible, and I'd have made the same call at authoring time. `#208`/`#209`/`#210` are all open commonMain-KDoc issues, and PR #248 (`docs/kdoc-208-209-210`) was opened **2 minutes after** this one — the author could not have known its file set. Using `Refs #202` instead of `Fixes #202` is the correct consequence and keeps the item traceable. One thing for the record, since it changes who has to pick it up: #248 as filed touches only `Message.kt` and `Types.kt`, **not** `Proxy.kt`, so the conflict this PR avoided did not materialise and `Proxy.kt:94-98` is currently claimed by no PR. #202 staying open is the right tracking mechanism; just noting that item 1 needs someone to actually take it.

---

### Guard rails — all respected

- **`docs/BACKENDS.md:35` (the `PropertiesChanged` ✅/✅ emission row) is byte-identical to `origin/main`.** Verified explicitly: `diff <(git show HEAD:docs/BACKENDS.md | awk 'NR==35') <(git show origin/main:docs/BACKENDS.md | awk 'NR==35')` → no output. Amending it is option 1 of the still-open **#193** (confirmed OPEN), and this PR correctly did not decide that on the owner's behalf.
- **`docs/BACKENDS.md:45`** (`Method.NoReply`, #221) — byte-identical to `origin/main` by the same check.
- **`AGENTS.md`** — not in the diff at all (`git diff --name-only … -- AGENTS.md` → 0 files).
- **Prose only** — `git diff --name-status origin/main...HEAD` is exactly `M CLAUDE.md` / `M docs/BACKENDS.md`. Zero `.kt`, zero `.kts`, zero `api/*.api`.

### Gates

- `./gradlew ktlintCheck apiCheck` (JDK 21, via the build slot, in the review worktree): **BUILD SUCCESSFUL**, 14s, 64 tasks executed. `apiCheck` includes `klibApiCheck`, so the klib dump is validated too.
- **`:dokkaGeneratePublicationHtml` — reasoning verified, correctly skipped.** I checked every `*.gradle.kts` for Dokka configuration, not just the root: the only prose input anywhere is `build.gradle.kts:262 includes.from(rootProject.file("dokka/moduledoc.md"))`, and `codegen/build.gradle.kts:79-85` sets no `includes` at all. Neither `CLAUDE.md` nor `docs/BACKENDS.md` is referenced by any build file. Nothing in this diff can move Dokka output, so a warning count from it would be noise. Agreed; not run.
- **CI: fully green** — 15 checks pass (`static-checks`, `native-build-x64`, `arm-runtime-tests`, `full-tests-x64`, both `build-samples`, all 10 `pr-smoke` matrix legs), 2 nightly-only jobs skipping, zero failures. Merge state `CLEAN`.

### Quality

The corrected text is genuinely harder to re-break than what it replaced, which is the bar for a doc fix. Three things earn that: the factory count is now sourced to `api/*.api` **in the text itself**, so the next reader knows where to re-derive it; the fd bullet names the mechanism (`no expect`, no actual) rather than a symptom, so it stays true unless the declaration itself moves; and the matrix note enumerates all three absences instead of two, which is what made `9 of 11` look internally consistent for so long.

On the two additions I was asked to judge specifically — **both are durable context, not bloat**:
- The `@Deprecated(ERROR)`/`#126` parenthetical is one sentence and does real work: #191 quotes the old wording verbatim, so a reader arriving from the issue needs to know the text was *replaced*, not that the doc drifted again. Without it the correction looks like an unexplained contradiction of the issue.
- The `compile_test` paragraph is longer than the minimum, but every clause answers a question an agent actually hits, and I verified all six facts in it independently. The `#237` sentence is the load-bearing one — it is the difference between "there is a module" and "this module catches things `:codegen:test` cannot", which is exactly why it kept getting skipped. The ktlint/apiCheck exclusion clause pre-empts the "why isn't my golden linted" confusion. Keeping it.

Two non-blocking prose nits, offered for whenever this file is next touched — neither is worth a round trip:

1. **`BACKENDS.md:410-413`, tombstone placement.** The closed-item parenthetical now sits *between* the divergence bullets and the closing "These are tracked as post-1.0 items; none is on the cross-process path." A skimmer can read "These" across the intervening paragraph. The tombstone itself is the right anti-re-break device (per #191, this file was edited twice after #172 without correcting the line) — it would just read better *after* the closing sentence, or as a trailing footnote.
2. **`BACKENDS.md:294`.** "On native they adopt the descriptor as documented in their KDoc." is now sandwiched mid-bullet between the JVM historical parenthetical and the `WireDbusBackend` remnant sentence, interrupting the JVM thread. It was pre-existing text that the surrounding rewrite flowed around; moving it to the end of the bullet would keep the JVM discussion contiguous.

Classified `docs_only` → tier 1 under fleet review policy (`sdbus-kotlin` has no project overrides). Diff is prose-only, all claims independently re-measured and confirmed, guard rails intact, static gates and CI green. Approving and merging.

Issue #247, opened 2026-08-08
JVM credsEuid/credsEgid report getuid()/getgid() as if they were effective ids — leave them absent instead of answering wrong


monkopedia-coder:
Split out of #210, which is now documented but not fixed. #210's option (3) was the part its author called non-optional regardless of how the rest is resolved, and it is the only behaviour change here.

Re-verified against `main` @ `79bdbe5`.

## The defect

`src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt:827-835`:

```kotlin
return copy(
    credsPid = creds.pid,
    credsUid = creds.uid,
    credsEuid = creds.uid,     // :830 — REAL uid reported as the effective uid
    credsGid = creds.gid,
    credsEgid = creds.gid,     // :832 — REAL gid reported as the effective gid
    ...
)
```

`creds.uid` / `creds.gid` come from `com.sun.security.auth.module.UnixSystem().uid` / `.gid` (`:798-813`), i.e. `getuid()` / `getgid()`. `geteuid()` / `getegid()` are never called. So `Message.credsEuid` and `Message.credsEgid` return a plausible-looking wrong number rather than failing, on the one JVM path that populates credentials at all (`withLocalSenderCredentials`, reached only from `:768`, the received-signal path).

`src/jvmTest/kotlin/com/monkopedia/sdbus/JvmCredentialsTest.kt:14-15` documents the shortcut — "euid/egid equal uid/gid for a non-setuid process" — which is true for the test runner and false in general, which is why it has never been caught.

Native is correct: `Message.native.kt` queries `SD_BUS_CREDS_EUID` / `_EGID` via `sd_bus_query_sender_creds` per message.

## Why it is worth a behaviour change and not just the doc

A throw is recoverable; a wrong uid that passes a privilege comparison is not. Everything else in #210 degrades to "this accessor throws on JVM", which a consumer can guard. These two degrade to "this accessor answers with the receiving process's real id while claiming to be an effective id".

## Suggested fix (my recommended default)

[...]

Issue #246, opened 2026-08-08
JVM Message: make 'complete' actually do something on isAtEnd/copyTo/rewind, stop copying metadata in copyTo, and unshare isValid with the error-reply flag


monkopedia-coder:
Split out of #209. #209's doc half landed as a KDoc-only PR that now *states* each divergence; this is the behaviour work it deliberately did not do.

Every claim below was re-verified against `main` @ `79bdbe5`.

## 1. `complete` is inert on all three JVM actuals

`src/jvmMain/kotlin/com/monkopedia/sdbus/Message.jvm.kt:186-195`:

```kotlin
actual fun isAtEnd(complete: Boolean): Boolean = readIndex >= payload.size
actual fun copyTo(destination: Message, complete: Boolean) {
    destination.payload.addAll(payload)   // always the whole payload, from index 0
    destination.metadata = metadata
}
actual fun rewind(complete: Boolean): Unit = run {
    readIndex = 0                         // always "complete"
    enteredVariantValues.clear()
}
```

Native passes the flag through to `sd_bus_message_at_end` / `_copy` / `_rewind` (`src/nativeMain/kotlin/com/monkopedia/sdbus/Message.native.kt:591-609`). All three produce silent wrong answers on JVM rather than exceptions: `rewind(false)` inside a container rewinds to the start of the body, `copyTo(dst, false)` re-copies already-consumed values, `isAtEnd(true)` can report `true` with containers still open.

The JVM message already tracks `enteredVariantValues` (`Message.jvm.kt:55`), so a container-relative cursor exists to build on.

## 2. `copyTo` overwrites the destination's metadata

`Message.jvm.kt:189` does `destination.metadata = metadata` on top of the payload copy. Metadata is *identity*, not contents — interface/member/path/sender/destination, the `valid` flag, and all sender credentials. Native's `sd_bus_message_copy` copies body data only.

Latent today: the only route that could hit it is `MessageUtil.kt:54-56 internal fun Message.append(variant: Variant)` → `Types.kt:144 this.msg.copyTo(msg, true)`, and `Message.append(Variant)` is referenced only from `src/nativeMain/kotlin/com/monkopedia/sdbus/internal/MessageEncoder.kt:179-233`, so it is dead on JVM. If it ever becomes live, a real `MethodCall` would have its header replaced by the variant's throwaway `PlainMessage` metadata and then fail in `MethodCall.send` with `"missing interface name"` (`Message.jvm.kt:401`).

[...]

Issue #245, opened 2026-08-08
UnixFd suite asserts the degraded JVM fd mode as a valid outcome, so a green run does not record which ownership contract held


monkopedia-coder:
Split out of #208 (item 3). #208's doc half landed as a KDoc-only PR; this is the part that needs a test change and so was deliberately left out of it.

## What is left

`JvmUnixFdSupport.supportsFdDuplicationSemantics` gates two genuinely different contracts on JVM: with junixsocket's native support the primary `UnixFd(fd)` constructor `dup(2)`s and `release()` closes; without it the constructor returns the caller's fd unchanged and `release()` closes nothing (`src/jvmMain/kotlin/com/monkopedia/sdbus/Types.jvm.kt:86-100`). The suite currently asserts **both** contracts as acceptable, so a green run does not record which one held.

- `src/commonTest/kotlin/com/monkopedia/sdbus/unit/UnixFdCommonTest.kt` — every one of the six tests is `if (FdTestSupport.supportsFdDuplicationSemantics) { … } else { … }`, and the `else` branch asserts the degraded behaviour (lines 13-17, 28-33, 41-45, 67-76, 87-96, 108-118).
- `src/jvmTest/kotlin/com/monkopedia/sdbus/unit/FdTestSupport.jvm.kt:14-32` — in the degraded mode `createPipePair()` hands out synthetic integers from an `AtomicInteger` and `closeFd` consults a `HashSet`, so **no real descriptor is ever opened**. The `else` branches above are asserting against that fake table, not against the kernel.
- `src/commonTest/kotlin/com/monkopedia/sdbus/unit/MessageCommonTest.kt:138-142` carries the same shape.

Same family as the dbusmock local no-op trap: the suite cannot distinguish "the contract holds" from "the contract was not exercised".

## Already fixed — do not redo

`src/jvmTest/kotlin/com/monkopedia/sdbus/JvmUnixFdTest.kt` was the worst instance (a bare `if (!supported) return`, which JUnit reports as a pass). #228 replaced it with `org.junit.Assume.assumeTrue` and an explanatory message, so that one now records a real skip. Only the `commonTest` else-branches and the synthetic-fd table remain.

## Measured: the degraded branch is not live locally

[...]

PR #248, opened 2026-08-08
docs: correct three commonMain KDoc claims the actuals do not honour (UnixFd scope-exit close, Message 'complete'/isValid/peekType, JVM sender credentials)

### LINKED ISSUE #208: UnixFd: commonMain KDoc asserts scope-exit close (false on both backends) and unconditional dup (silently degrades to aliasing on JVM); the test suite passes in either mode
Found during the 2026-08-07 read-only claim-verification pass. Same shape as #178/#179/#181 (`MethodInvoker.timeout`) and #184 (`Connection.methodCallTimeout`): a KDoc asserting a lifetime/ownership property that the implementation does not deliver. All line references re-read directly; no build was run.

## 1. "closes the fd when the UnixFd goes out of scope" — false on both backends, and native's own copy was already corrected

`src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt:319-322`:

```
/**
 * UnixFd is a representation of file descriptor D-Bus type that owns
 * the underlying fd, provides access to it, and closes the fd when
 * the UnixFd goes out of scope.
 */
```

`src/nativeMain/kotlin/com/monkopedia/sdbus/UnixFd.kt:39-42` — **the same sentence, already fixed**:

```
/**
 * A representation of the file descriptor D-Bus type that owns
 * the underlying fd, provides access to it, and closes the fd when
 * the UnixFd is released.
 */
```

**Measurement.** Neither backend closes at scope exit. Native uses `createCleaner` (`src/nativeMain/.../UnixFd.kt:61-63`); JVM uses `java.lang.ref.Cleaner` (`src/jvmMain/.../Types.jvm.kt:22,62-64`). Both are GC-time and non-deterministic.

**Inference.** The correction landed on the `actual` and never reached the `expect` in `commonMain` — which is the declaration Dokka publishes, so the published API docs still carry the wrong sentence. The `adopt` KDoc 20 lines below it (`Types.kt:341-342`) states the correct GC-based contract, so the class-level sentence contradicts its own companion.

**Consequence.** A caller who reads the published docs writes `run { UnixFd(fd) }` and expects the descriptor back at scope exit. It comes back at an unpredictable GC, or (see below) never. Fd exhaustion under load, with nothing in the API pointing at the cause.

**Fix:** copy the native wording onto `Types.kt:321-322`. One line.

## 2. `UnixFd(fd)` "duplicates the descriptor" is conditional on a silent reflection probe, and so is `release()` closing anything

`src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt:343-345`:

> `Contrast with the primary` `UnixFd(fd)` `constructor, which duplicates the descriptor and leaves the caller's fd untouched.`

**Measurement.** `src/jvmMain/kotlin/com/monkopedia/sdbus/Types.jvm.kt:80-100`:

```kotlin
internal object JvmUnixFdSupport {
    private val nativeSupport: NativeUnixSocketSupport? = NativeUnixSocketSupport.createOrNull()
    val supportsFdDuplicationSemantics: Boolean get() = nativeSupport != null

    fun checkedDup(fd: Int): Int {
        if (fd < 0) return fd
        val support = nativeSupport ?: return fd        // <-- no dup; returns the CALLER'S fd
        ...
    }
    fun closeFd(fd: Int): Boolean {
        if (fd < 0) return true
        val support = nativeSupport ?: return false     // <-- no close; failure swallowed
```

`NativeUnixSocketSupport.createOrNull()` (`Types.jvm.kt:210-244`) reflects on the non-exported class `org.newsclub.net.unix.NativeUnixSocket` and calls `setAccessible(true)` on its private statics, inside `runCatching { … }.getOrNull()`. On a JPMS-modular consumer (`InaccessibleObjectException`) or any junixsocket native-load failure it yields `null` with no log and no exception.

`closeFd`'s `false` return is discarded — `OwnedFdState.run()` (`Types.jvm.kt:69-72`) ignores it.

Native by contrast genuinely `dup(2)`s and throws on failure (`src/nativeMain/.../UnixFd.kt:99-108`).

**Inference.** In the degraded state both halves of the documented contract invert: the primary constructor **aliases** the caller's fd instead of duplicating it, and `release()` **does not close** it. Nothing observable distinguishes the two modes — `supportsFdDuplicationSemantics` is `internal`, so a consumer cannot even query it.

What else would produce that measurement? If junixsocket always loaded successfully in every consumer environment the branch would be dead — but the code exists precisely because it is not guaranteed, and the `runCatching` proves the author expected failure.

**Consequence.** Either a double-close (the library closes an fd the caller still owns, and the caller closes it again — which on a busy process can close an unrelated descriptor that has since reused the number) or an fd leak, depending on which branch is live. Native is unaffected.

## 3. The test suite is constructed to pass in *either* mode, so nothing surfaces which one ran

This is the load-bearing part, and it is the same shape as the dbusmock local no-op trap.

- `src/jvmTest/kotlin/com/monkopedia/sdbus/JvmUnixFdTest.kt:9` — `if (!JvmUnixFdSupport.supportsFdDuplicationSemantics) return`. A bare early return, so the **only** JVM duplication test fake-passes when unsupported. It has no other assertion.
- `src/jvmTest/kotlin/com/monkopedia/sdbus/unit/FdTestSupport.jvm.kt:15-32` — when unsupported, `createPipePair()` hands out synthetic integers from an `AtomicInteger` and `closeFd` consults a `HashSet`. **No real descriptor is ever opened.**
- `src/commonTest/.../UnixFdCommonTest.kt` (lines 13-17, 28-33, 41, 48, 67, 87, 92, 108, 114) and `MessageCommonTest.kt:138-142` each carry an `else` branch asserting the *degraded* behaviour against that fake table.

So the entire `UnixFd` ownership suite reports green on a JVM where `UnixFd` neither duplicates nor closes anything, and the run output does not say which mode it was in.

**Fix (suggested, not decided):** have the JVM suite `fail()` — or at minimum print a loud one-line banner — when `supportsFdDuplicationSemantics` is false, so the degraded mode is visible in CI logs rather than indistinguishable from a real pass. CI is the environment where this should never be degraded.

## Unresolved

**Which branch is live in this repo's CI is UNRESOLVED-NEEDS-BUILD.** No build was run. Settle with:

```
JAVA_HOME=/usr/lib/jvm/java-17-openjdk dbus-run-session -- \
  ./gradlew jvmTest --tests '*JvmUnixFdTest*' --info
```

and check whether `constructorFromFd_duplicatesDescriptorWhenSupported` reaches its `assertNotEquals` or returns at line 9. If it is degraded in CI, item 3 is not a hardening suggestion but a live hole and items 1-2 are not latent.

### LINKED ISSUE #209: Message KDoc documents native behaviour the JVM actual does not honour: 'complete' ignored on 3 methods, PeekedType.type null overloaded, isValid diverges
Found during the 2026-08-07 read-only claim-verification pass, reading `commonMain/Message.kt` line by line and following each claim into both `actual`s. Same family as the parity-hardening wave (#142-#145): the `expect` KDoc documents native's behaviour, and the JVM `actual` does something else without saying so. All lines re-read directly; no build was run.

## 1. The `complete` parameter is documented on three methods and ignored by all three JVM actuals

**Claims** (all public — `api/sdbus-kotlin.api:158,171,175`):

| | `src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt` |
|---|---|
| `:163` | `@param complete When `true`, also requires that any open containers have been exited` |
| `:171` | `@param complete When `true`, copy the whole message; otherwise copy from the current cursor` |
| `:181` | `@param complete When `true`, rewind past all containers to the very beginning` |

**Measurement.** Native passes it through (`src/nativeMain/.../Message.native.kt:591-609`):

```kotlin
actual fun isAtEnd(complete: Boolean): Boolean =
    sd_bus_message_at_end(msg, if (complete) 1 else 0) > 0
actual fun copyTo(destination: Message, complete: Boolean) {
    val r = sd_bus_message_copy(destination.msg, msg, if (complete) 1 else 0)
actual fun rewind(complete: Boolean) {
    val r = sd_bus_message_rewind(msg, if (complete) 1 else 0)
```

The JVM actuals never read it (`src/jvmMain/.../Message.jvm.kt:186-195`):

```kotlin
actual fun isAtEnd(complete: Boolean): Boolean = readIndex >= payload.size
actual fun copyTo(destination: Message, complete: Boolean) {
    destination.payload.addAll(payload)        // always the whole payload, from index 0
    destination.metadata = metadata
}
actual fun rewind(complete: Boolean): Unit = run {
    readIndex = 0                              // always "complete"
    enteredVariantValues.clear()
}
```

**Inference.** On JVM the parameter is inert; the behaviour is always the `complete = true` one. Nothing warns.

**Coverage.** `src/commonTest/.../MessageApiParityCommonTest.kt:40` and `MessageCommonTest.kt:162-164` are the only exercises and neither asserts a `true`-vs-`false` difference — so the contract is untested on *both* backends, which is why this survived.

**Consequence.** `rewind(false)` on a JVM message inside a container rewinds to the start of the body instead of the start of the current container, so the next read returns the wrong value rather than failing. `copyTo(dst, false)` copies values already consumed. `isAtEnd(true)` returns `true` while containers are still open, so an "is the body fully consumed and balanced?" check passes on a malformed read. All three are silent wrong answers, not exceptions.

## 2. `PeekedType.type == null` means two different things on JVM

**Claim.** `Message.kt:212` — `/** The D-Bus type code of the value at the read position, or `null` if at the end. */`

**Measurement.** Native (`Message.native.kt:581`) sets `type` from `takeIf { r > 0 }`, where `r == 0` is exactly sd-bus's end-of-container signal. JVM (`Message.jvm.kt:177-181`) returns `PeekedType(null, null)` in two further cases: the stored value is Kotlin `null`, and `inferSignature` returns `null` (`Message.jvm.kt:220,254` — `else -> null`, which catches every raw in-process object passed through without decomposition, a path the code explicitly supports and comments on at `Message.jvm.kt:618-621`).

**Consequence.** The end-of-message idiom this library itself uses on native — `if (peekType().type == null)` at `src/nativeMain/.../internal/MessageDecoder.kt:250` — silently reports "end of message" mid-body on JVM, truncating rather than erroring.

## 3. `isValid` diverges for locally-created and error messages

**Claim.** `Message.kt:154` — `/** Whether this message wraps a valid underlying D-Bus message. */` (public: `api/sdbus-kotlin.api:173`).

**Measurement.** Native `Message.native.kt:586-587`: `get() = msg != null` — true for every constructed message including `createPlainMessage()` and error replies. JVM `Message.jvm.kt:182-183`: `get() = metadata.valid`, and `Metadata.valid` defaults to `false` (`Message.jvm.kt:40`). `createPlainMessage()` (`Message.jvm.kt:533-535`) never sets it, so `createPlainMessage().isValid` is `false` on JVM and `true` on native. `createErrorReply` (`Message.jvm.kt:471-474`) deliberately sets `valid = false`, and `MethodCall.send` reads it to mean "not an error reply" (`:447`) — a **different concept** from the one the KDoc names.

`isValid` is asserted nowhere: no occurrence in `src/commonTest`, `src/jvmTest`, or `src/nativeTest`.

**Consequence.** Common code guarding with `if (!msg.isValid) return` silently drops every locally-built message on JVM and processes it on native.

## 4. `copyTo` also replaces the destination's metadata, which the KDoc does not say

**Claim.** `Message.kt:168` — `* Copies the contents of this message into [destination].`

**Measurement.** `Message.jvm.kt:189` does `destination.metadata = metadata` in addition to the payload. Metadata is identity, not contents: interface/member/path/sender/destination, `valid`, and all sender credentials. Native's `sd_bus_message_copy` copies body data only.

**Reachability.** `MessageUtil.kt:54-56 internal fun Message.append(variant: Variant)` → `Types.kt:144 this.msg.copyTo(msg, true)`. If `msg` were a real `MethodCall`/`Signal`, its metadata would be overwritten by the variant's throwaway `PlainMessage` metadata (`Metadata()` — all nulls, `valid = false`), after which `MethodCall.send` throws `"missing interface name"` (`Message.jvm.kt:401`). `Message.append(Variant)` is referenced only from `src/nativeMain/.../internal/MessageEncoder.kt:179-233`, so it is **dead on JVM today** — which is the only reason this has not fired. Latent, not currently firing.

## Suggested shape of the fix (not decided here)

For 1-3 the honest options are the usual two, and which one applies differs per item:

- **Make the JVM honour it** — feasible for `complete` on `rewind`/`isAtEnd` (the JVM message tracks `enteredVariantValues`, so a container-relative cursor exists), and for `isValid` (align on "wraps a constructed message", and give the error-reply flag its own name so the two concepts stop sharing a field).
- **Document the divergence** — the route #141 established for cases where no single cross-backend assertion can pass on both. Appropriate for `PeekedType.type` if inferring a signature for arbitrary in-process objects is not going to become total.

Either way the parity tests should gain an assertion that distinguishes `complete = true` from `complete = false`, since today none does on either backend.

### LINKED ISSUE #210: Message sender-credential KDoc is native-only: on JVM it throws for remote senders and reports THIS process's real uid as the sender's effective uid
Found during the 2026-08-07 read-only claim-verification pass. Filing because the failure mode is an authorization check that authorizes the wrong principal, not just a doc inaccuracy. Nothing here is exploitable against a running service — it is a library contract that a consumer could build a wrong check on. All lines re-read directly; no build was run.

## The documented contract

`src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt:185-204`:

```kotlin
/** The PID of the sending process, if credentials are available. */
val credsPid: Int
/** The real UID of the sender, if credentials are available. */
val credsUid: UInt
/** The effective UID of the sender, if credentials are available. */
val credsEuid: UInt
/** The real GID of the sender, if credentials are available. */
val credsGid: UInt
/** The effective GID of the sender, if credentials are available. */
val credsEgid: UInt
/** The supplementary GIDs of the sender, if credentials are available. */
val credsSupplementaryGids: List<UInt>
/** The SELinux security context of the sender. */
val seLinuxContext: String
```

Note the types are **non-nullable**, so "if credentials are available" has no representation in the signature — the caller cannot branch on availability, only catch.

## What native does

`src/nativeMain/kotlin/com/monkopedia/sdbus/Message.native.kt:611-700` calls `sd_bus_query_sender_creds` per message. Real per-sender credentials, any message type, any peer. The KDoc is accurate here.

## What the JVM backend does

**Measurement.** The only site that ever populates these fields is `withLocalSenderCredentials` (`src/jvmMain/.../internal/jvmdbus/WireDbusBackend.kt:815-835`), called from exactly one place — `:768`, the **received-signal** path — and it returns the metadata unchanged unless the sender resolves to one of this JVM's own connections:

```kotlin
private fun Message.Metadata.withLocalSenderCredentials(sender: String?): Message.Metadata {
    val senderName = sender?.takeIf { it.isNotBlank() } ?: return this
    if (LocalJvmServiceRegistry.resolveLocalUniqueName(senderName) == null) return this   // :817
    val creds = localProcessWireCredentials                                               // :818
    ...
    return copy(credsPid = creds.pid, credsUid = creds.uid, credsEuid = creds.uid,        // :830
                credsGid = creds.gid, credsEgid = creds.gid, ...)                         // :832
}
```

`localProcessWireCredentials` (`:798-813`) is `ProcessHandle.current().pid()` plus `com.sun.security.auth.module.UnixSystem()` — **this** process. Everything else leaves `Metadata.credsPid` etc. `null`, and reading the property then throws (`src/jvmMain/.../Message.jvm.kt:15-16`):

```kotlin
private fun <T> requireJvmCredential(value: T?, name: String): T = value
    ?: throw createError(-1, "$name failed: sender credentials unavailable on JVM")
```

**Two distinct defects follow.**

### (a) Not "the sender" for any remote peer

Reading any credential from inside a served method handler, or from a signal whose sender is another process, throws on JVM. The doc says nothing about this; "if credentials are available" reads as a property of the *message*, whereas on JVM it is a property of the *backend* and is never available for a remote peer.

### (b) `credsEuid` / `credsEgid` are not effective IDs

`WireDbusBackend.kt:830,832` assign `credsEuid = creds.uid` and `credsEgid = creds.gid`, and `creds.uid`/`gid` come from `UnixSystem().uid`/`.gid`, i.e. `getuid()`/`getgid()` — **real** IDs. `src/jvmTest/kotlin/com/monkopedia/sdbus/JvmCredentialsTest.kt:14-15` documents the shortcut ("euid/egid equal uid/gid for a non-setuid process"), which is true for the test runner and false in general.

## Consequence

An authorization check written against the documented contract behaves three different ways:

```kotlin
if (msg.credsUid != 0u) reject()      // native: correct
                                      // JVM, remote sender: throws SdbusException
                                      // JVM, same-process sender: returns THIS process's uid
```

The third is the dangerous one: the value is the *receiving* process's own identity, so it matches any comparison against `getuid()` and the check authorizes the wrong principal while appearing to work. That it only triggers for same-process senders is what makes it survive testing.

## The justifying comment, and whether it still holds

`WireDbusBackend.kt:795-797`:

> `Credentials of an EXTERNAL sender would require a GetConnectionCredentials bus call, which cannot run on the reader thread that delivers signals without deadlocking; no test needs it, so it is intentionally left out here.`

The deadlock half is a real technical constraint and I did not falsify it. The second half — *"no test needs it"* — is a justification about the test suite standing in for a justification about the public API. Nothing marks `credsPid`/`credsUid`/… as native-only, and `docs/BACKENDS.md` does not carry a row for it, so a consumer has no way to learn this from anywhere except this internal comment.

## Options (not decided — flagging rather than picking)

1. **Document the divergence** (the #141 route). Cheapest, honest. Add a `@throws SdbusException` note plus a "JVM backend: only available for same-process senders" line to all seven properties, and a `docs/BACKENDS.md` row. Does not fix (b) — `credsEuid` would still be wrong, so it should additionally stop being reported as *effective*.
2. **Implement `GetConnectionCredentials`** off the reader thread (dispatch to the connection's own executor and cache per unique name). Removes both defects, but it is a real feature with a deadlock hazard the comment already names.
3. **Fail loudly instead of quietly wrong.** Whatever is chosen for (a), the same-process path should not report `credsEuid = getuid()` as if it were an effective UID. Either query `geteuid()`/`getegid()` properly, or leave those two `null` so the property throws rather than returning a plausible wrong number.

(3) is the part I would treat as non-optional regardless of how (a) is resolved: a throw is recoverable, a wrong uid that passes a root check is not.

## Trap to avoid

Do not "fix" this by making the JVM return the local process credentials for *all* senders — that turns a throw into the silently-wrong case for every remote peer, which is strictly worse than today.

monkopedia-coder:
**Documentation only.** Two files, both `commonMain`, both KDoc-comment-only. No `actual` edits, no test edits, no build edits.

Three issues, one PR, because all three are `commonMain` KDoc asserting behaviour an `actual` does not honour, in two overlapping files (`Message.kt` twice, `Types.kt` once).

Same defect class as #179 → #178: a KDoc that was wrong since it was written, unchecked because nobody had reason to look at that path, and a false claim in `commonMain` is a promise made on behalf of **both** backends — so it is wrong about one of them or wrong about both.

### #208 — `UnixFd` closes the fd "when the UnixFd goes out of scope"

**Claimed** (`src/commonMain/.../Types.kt:319-322`): *"owns the underlying fd, provides access to it, and closes the fd when the UnixFd goes out of scope."*

**Contradicted by** both backends: native registers a `kotlin.native.ref.createCleaner` (`src/nativeMain/.../UnixFd.kt:61-63`), JVM a `java.lang.ref.Cleaner` (`src/jvmMain/.../Types.jvm.kt:22,62-64`). Both are GC-time and non-deterministic; nothing runs at scope exit. The native `actual`'s copy of the same sentence had **already been corrected** to "when the UnixFd is released" (`src/nativeMain/.../UnixFd.kt:39-42`) — the fix never reached the `expect`, which is the declaration Dokka publishes, so the published docs still carried the wrong sentence while the `adopt` KDoc 20 lines below stated the correct GC-based contract.

**Corrected to** "closes the fd when the UnixFd is released", plus a paragraph naming the two release routes (`release`, or `use { }` since `UnixFd` is a `Resource`) and the GC backstop, and stating in #221's style what changed and that it was never true.

**Second claim, same issue** (`Types.kt:343-345`): *"the primary `UnixFd(fd)` constructor, which duplicates the descriptor and leaves the caller's fd untouched."* Unconditional on native (`dup(2)`, throws on failure — `UnixFd.kt:99-108`); on JVM conditional on a reflective probe of junixsocket's non-exported `NativeUnixSocket` (`Types.jvm.kt:210-244`, inside `runCatching { … }.getOrNull()`). When that probe fails, `checkedDup` returns the caller's fd unchanged and `closeFd` returns `false` without closing (`Types.jvm.kt:86-100`), and the `false` is discarded by `OwnedFdState.run()` (`:69-72`). Both halves of the documented contract invert, silently, with `supportsFdDuplicationSemantics` `internal` so a consumer cannot query which mode is live. **Corrected** by a class-level paragraph stating the JVM condition and what degrades, and a cross-reference from `adopt`.

### #209 — four `Message` claims documenting native's behaviour only

| Claim | Contradiction | Correction |
|---|---|---|
| `Message.kt:163/171/181` — `@param complete` on `isAtEnd` / `copyTo` / `rewind` | Native passes it to `sd_bus_message_at_end` / `_copy` / `_rewind` (`Message.native.kt:591-609`); all three JVM actuals never read it (`Message.jvm.kt:186-195`) | Each `@param` now says the flag is honoured on native only, and what JVM does instead — per method, since the wrong answer differs (`rewind(false)` goes to the start of the body not the container; `copyTo(dst,false)` re-copies consumed values; `isAtEnd` ignores container nesting entirely) |
| `Message.kt:168` — *"Copies the contents of this message into [destination]"* | `Message.jvm.kt:189` also does `destination.metadata = metadata` — identity, not contents: interface/member/path/sender/destination, `valid`, and all sender credentials. Native's `sd_bus_message_copy` is body-only | States that the JVM copy replaces the destination's metadata and native's does not |
| `Message.kt:212` — `PeekedType.type` is *"null if at the end"* | Native sets it from `takeIf { r > 0 }`, exactly sd-bus's end signal (`Message.native.kt:581`). JVM returns `null` in two further cases: a Kotlin-`null` value, and a value `inferSignature` cannot type (`Message.jvm.kt:177-181`, `:220`, `:254`) — the latter being the raw in-process object path the backend deliberately supports (`:618-621`) | States that `null` is overloaded on JVM and is not a sound end-of-message test there, though it is on native |
| `Message.kt:154` — *"Whether this message wraps a valid underlying D-Bus message"* | Native: `msg != null` (`Message.native.kt:586-587`), true for every message. JVM: `metadata.valid`, defaulting `false` (`Message.jvm.kt:40,182-183`), so `createPlainMessage().isValid` is `false` on JVM and `true` on native — and `createErrorReply` sets it `false` on purpose (`:471-474`) with `MethodCall.send` reading it back as "not an error reply" (`:447`), a different concept in the same field | States the divergence, that the field carries two concepts on JVM, and that the value is unspecified for anything but native's `msg != null` |

Per the #141 route: these are documented as divergences rather than asserted into a cross-backend test, because no single assertion passes on both.

### #210 — the `creds*` KDoc is native-only

**Note what #236 (merged, `680e981`) already covered, and is NOT duplicated here:** that the accessors **throw** rather than returning a sentinel, and that a direct connection's `sender` is peer-asserted so credentials reached through it are not authoritative. #210 was filed at `17:25Z`, ~8h before #236 merged, and its "the types are non-nullable so availability has no representation" observation is that PR's subject. What #236 did **not** say, and this PR adds:

**(a) On JVM no credential ever describes another process.** The only site that populates them is `withLocalSenderCredentials` (`WireDbusBackend.kt:815-835`), called from exactly one place — `:768`, the received-signal path — and it returns the metadata unchanged unless the sender resolves to a connection in this same JVM (`:817`). The values come from `localProcessWireCredentials` (`:798-813`) = `ProcessHandle.current().pid()` + `UnixSystem()`, i.e. **this** process. Everything else leaves the fields `null` and the accessor throws (`Message.jvm.kt:15-16`). Verified by search: `credsUid =` / `credsEuid =` / `withLocalSenderCredentials` occur nowhere else in `src/`. So on JVM these properties either throw or describe *you* — never a remote principal. The doc now says that, and says a comparison against the local uid succeeds on JVM because both sides came from the same process, not because the sender was authorized.

**(b) `credsEuid` / `credsEgid` are not effective ids.** `WireDbusBackend.kt:830,832` assign `credsEuid = creds.uid` and `credsEgid = creds.gid`, and those come from `UnixSystem().uid`/`.gid` = `getuid()`/`getgid()`. `geteuid()` is never called. `JvmCredentialsTest.kt:14-15` documents the shortcut as "euid/egid equal uid/gid for a non-setuid process" — true for the test runner, false in general. Both properties now say plainly that JVM does not report an effective id there.

## Behaviour changes NOT made here — split out

A doc PR that quietly changes behaviour is worse than two PRs, so each of these is filed rather than done:

- **#245** — the `UnixFd` suite asserts the degraded JVM fd mode as a valid outcome (`UnixFdCommonTest.kt` has an `else` branch per test asserting the degraded contract against `FdTestSupport.jvm.kt`'s synthetic-integer fd table, where no real descriptor is ever opened). **Partly stale already:** #208 cited `JvmUnixFdTest.kt:9`'s bare `if (!supported) return`, which #228 replaced with `org.junit.Assume.assumeTrue` at `20:05` on 2026-08-07, ~7h after #208 was filed. Only the `commonTest` else-branches and the fake fd table remain.
- **#246** — make the JVM honour `complete`, stop copying metadata in `copyTo`, and split the error-reply flag out of `Metadata.valid`; plus the parity assertion that distinguishes `complete = true` from `false`, which exists on neither backend today.
- **#247** — stop reporting `getuid()`/`getgid()` as `credsEuid`/`credsEgid`; leave them absent so the accessor throws. #210's author called this non-optional regardless of how the rest resolves, and I agree: a throw is recoverable, a wrong uid that passes a root check is not.

`docs/BACKENDS.md` has no sender-credential row, which is where a consumer would look for (a) — **not touched**, that file is being edited under #191/#202 right now. Noted in #247 instead.

## #208's UNRESOLVED question, settled

#208 left "which fd branch is live in CI" as `UNRESOLVED-NEEDS-BUILD`, noting that if the degraded branch were live, its items would not be latent. **Measured:** `dbus-run-session -- ./gradlew :jvmTest --tests '*JvmUnixFdTest*'` (JDK 21, x86_64) → `tests="1" skipped="0"`, the case executing in 0.082s. `assumeTrue` passed, so `supportsFdDuplicationSemantics` was `true` and the real `assertNotEquals(first.fd, second.fd)` ran. The degraded branch is **not** live on this host, so #208's items 1-2 are latent and item 3 is hardening, not a live hole. Not measured on the CI runners — flagged in #245.

## Verification — what I ran, and what I did not

**Ran** (JDK 21, this branch):

- `./gradlew ktlintCheck apiCheck` — **BUILD SUCCESSFUL**, and `git status --porcelain` immediately after shows only the two edited source files: `api/*.api` did not move, as a KDoc-only change must not.
- `./gradlew :dokkaGeneratePublicationHtml` — **BUILD SUCCESSFUL** (the guard that matters, since malformed KDoc has broken this build before — #189). Dokka emits **8** `Couldn't resolve link` warnings, all pre-existing in untouched `jvmMain` files (`DBusWireConnection.kt` ×4, `WireDbusBackend.kt` ×4) and **none naming `Message.kt` or `Types.kt`** — a working detector reporting absence, not a silent pass. Every link I added (`[release]`, `[Resource]`, `[createPlainMessage]`, `[MethodCall.createErrorReply]`, `[credsUid]`, `[credsGid]`, `[credsEuid]`, `[complete]`, `[destination]`) resolves.
- Spot-checked the rendered HTML, so the KDoc is genuinely published and not silently dropped: `-unix-fd/index.html` contains "Nothing closes the fd at scope exit" and the junixsocket paragraph, `-message/creds-euid.html` contains "does not report an effective UID", `-peeked-type/type.html` contains "overloaded on the JVM backend".
- The `:jvmTest` fd measurement above.

**Did not run:** the rest of the test suites (`jvmTest` in full, `linuxX64Test`, cross-tests). A comment-only change cannot alter behaviour and CI covers them; I am not claiming test verification I did not do.

## Incidental, and corrected at the source

**#199 was closed by automation while the decision it names is still open** — `closed by monkopedia-reviewer` at `01:30:37Z` with `stateReason: COMPLETED`, coincident with #236's merge, and `monkopedia-coder` commented *"the decision here is untouched and this issue stays open"* **32 seconds later**. #236's body says "Refs #199 — this does not decide it". The KDoc on `main` (which this PR extends) tells readers it is an open question, which a closed issue contradicts. Reopened with the timeline cited; no labels or code touched.

Fixes #208
Fixes #209
Fixes #210
Refs #199, #245, #246, #247

### DIFF
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
@@ -151,7 +151,20 @@ expect sealed class Message {
      */
     fun peekType(): PeekedType
 
-    /** Whether this message wraps a valid underlying D-Bus message. */
+    /**
+     * Whether this message wraps a valid underlying D-Bus message.
+     *
+     * **Backend-dependent — do not branch on this from common code.** On native it is exactly
+     * "wraps a live `sd_bus_message`", which is `true` for every message this library hands out,
+     * locally built ones included. On the JVM backend it reports an internal flag that only the
+     * paths which build a message from the wire or from a dispatch result set: a message from
+     * [createPlainMessage] reads `false` there, and [MethodCall.createErrorReply] sets the flag
+     * `false` deliberately to mean "this reply carries an error" — a different concept from the
+     * one this property is named for, sharing the same field. So `if (!isValid) return` discards
+     * locally built messages on JVM and keeps them on native. Which of the two definitions this
+     * property should have is unsettled; treat the value as unspecified for anything but native
+     * `msg != null`.
+     */
     val isValid: Boolean
 
     /** Whether this message carries no body data. */
@@ -160,12 +173,24 @@ expect sealed class Message {
     /**
      * Whether the read cursor has reached the end of the message body.
      *
+     * **[complete] is honoured on native only.** The JVM backend ignores it and answers a third
+     * question instead — whether the cursor has consumed the flat payload — with no regard to
+     * container nesting, so `isAtEnd(true)` there can report `true` while containers are still
+     * open, and `isAtEnd(false)` does not stop at the end of the current container.
+     *
      * @param complete When `true`, also requires that any open containers have been exited
      */
     fun isAtEnd(complete: Boolean): Boolean
 
     /**
      * Copies the contents of this message into [destination].
+     *
+     * **[complete] is honoured on native only**, and the JVM backend copies more than the
+     * contents. On JVM the flag is ignored — the whole payload is always copied from the start,
+     * never from the cursor — and the copy additionally overwrites [destination]'s metadata
+     * (interface, member, path, sender, destination, validity flag and sender credentials) with
+     * this message's. Native's `sd_bus_message_copy` moves body data only and leaves the
+     * destination's header alone.
      *
      * @param destination Message to copy into
      * @param complete When `true`, copy the whole message; otherwise copy from the current cursor
@@ -177,6 +202,11 @@ expect sealed class Message {
 
     /**
      * Resets the read cursor to the start of the message.
+     *
+     * **[complete] is honoured on native only.** The JVM backend ignores it and always rewinds to
+     * the very beginning of the body, so `rewind(false)` inside a container does not return the
+     * cursor to the start of that container — it returns it to the start of the message, and the
+     * next read yields a value from there rather than failing.
      *
      * @param complete When `true`, rewind past all containers to the very beginning
      */
@@ -201,6 +231,19 @@ expect sealed class Message {
      * [issue #199](https://github.com/Monkopedia/sdbus-kotlin/issues/199) — so treat it as
      * unspecified rather than as a contract.
      *
+     * **On the JVM backend no credential ever describes another process.** Credentials are
+     * attached on exactly one path there — a received signal whose sender resolves to another
+     * connection inside this same JVM — and the values attached are read from the *receiving*
+     * process (`ProcessHandle.current()` plus `com.sun.security.auth.module.UnixSystem`), which
+     * on that path is the same process. Every other message, including any signal or method call
+     * that came from a different process, carries no credentials at all, so these accessors throw.
+     * Resolving an external peer would need a `GetConnectionCredentials` call the backend does not
+     * make. There is therefore no JVM configuration in which these properties report a remote
+     * principal: they either throw or describe you. The native backend queries real per-sender
+     * credentials for any peer and any message type. Do not write a portable authorization check
+     * on these properties — on JVM a comparison against the local uid succeeds because both sides
+     * of it came from the same process, not because the sender was authorized.
+     *
      * @throws SdbusException if credentials are unavailable for this message
      */
     val credsPid: Int
@@ -214,6 +257,12 @@ expect sealed class Message {
     /**
      * The effective UID of the sender. Carries the same availability and trust caveats as
      * [credsPid], including throwing [SdbusException] when credentials are unavailable.
+     *
+     * **The JVM backend does not report an effective UID here.** On its one credential-bearing
+     * path it stores `getuid()` — the real UID, the same value as [credsUid] — in this field; it
+     * never calls `geteuid()`. The result is a plausible wrong answer rather than a failure, so a
+     * check that reads it can pass for a process whose effective identity differs from its real
+     * one. Only the native backend reports a genuine effective UID.
      */
     val credsEuid: UInt
 
@@ -226,6 +275,9 @@ expect sealed class Message {
     /**
      * The effective GID of the sender. Carries the same availability and trust caveats as
      * [credsPid], including throwing [SdbusException] when credentials are unavailable.
+     *
+     * **The JVM backend does not report an effective GID here**, for the same reason as
+     * [credsEuid]: it stores `getgid()`, identical to [credsGid], and never calls `getegid()`.
      */
     val credsEgid: UInt
 
@@ -247,7 +299,16 @@ expect sealed class Message {
  * for container types, the signature of the contained elements.
  */
 class PeekedType(
-    /** The D-Bus type code of the value at the read position, or `null` if at the end. */
+    /**
+     * The D-Bus type code of the value at the read position, or `null` if at the end.
+     *
+     * **`null` is overloaded on the JVM backend.** There it also means "the value at the cursor is
+     * Kotlin `null`" and "the value at the cursor is an in-process object whose D-Bus signature
+     * cannot be inferred" — the latter being a path the JVM backend deliberately supports for
+     * same-process traffic. `type == null` is therefore a sound end-of-message test on native
+     * (where it maps to sd-bus's end-of-container return) but not on JVM, where it can report the
+     * end mid-body and silently truncate a read.
+     */
     val type: Char?,
     /** For container types, the signature of the contained elements, otherwise `null`. */
     val contents: String?
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt
@@ -319,11 +319,32 @@ value class Signature(
 /**
  * UnixFd is a representation of file descriptor D-Bus type that owns
  * the underlying fd, provides access to it, and closes the fd when
- * the UnixFd goes out of scope.
+ * the UnixFd is released.
+ *
+ * **Nothing closes the fd at scope exit.** Release is either explicit — [release], or a
+ * `use { }` block, since [UnixFd] is a [Resource] — or it happens at some later, unpredictable
+ * garbage collection, which both backends use only as a backstop (a `kotlin.native.ref.Cleaner`
+ * on native, a `java.lang.ref.Cleaner` on JVM). Dropping the last reference to a `UnixFd` and
+ * letting the variable fall out of scope therefore returns the descriptor at no defined time, so
+ * code that acquires fds in a loop must release them deterministically or it can exhaust the
+ * process fd table. (This paragraph corrects the class documentation published up to and
+ * including 1.0.1, which said the fd is closed "when the UnixFd goes out of scope"; that was
+ * never true of either backend.)
  *
  * UnixFd can be default constructed (owning invalid fd), or constructed from
  * an explicitly provided fd by either duplicating ([UnixFd] primary constructor)
  * or adopting that fd as-is ([UnixFd.adopt]).
+ *
+ * **Duplicating and closing are backend-dependent on JVM.** The native backend always `dup(2)`s
+ * in the primary constructor — throwing if the syscall fails — and always `close(2)`s on release.
+ * The JVM backend reaches the descriptor through junixsocket's *optional* native support; when
+ * that support is absent (no native binary for the platform, or a consumer where the reflective
+ * access it needs is denied, e.g. a JPMS-modular application) the primary constructor returns the
+ * caller's fd **unchanged**, so the instance aliases the caller's descriptor instead of owning a
+ * duplicate, and [release] then closes nothing. Both degradations are silent: no exception, no
+ * log, and no public way to ask which mode is in effect. On JVM, consequently, do not assume the
+ * fd you passed in is still exclusively yours after `UnixFd(fd)`, and do not treat a completed
+ * `release()` as proof that a descriptor was closed.
  */
 expect class UnixFd internal constructor(fd: Int, adoptFd: Unit) : Resource {
 
@@ -343,6 +364,10 @@ expect class UnixFd internal constructor(fd: Int, adoptFd: Unit) : Resource {
          * close [fd] afterwards. Contrast with the primary `UnixFd(fd)`
          * constructor, which duplicates the descriptor and leaves the
          * caller's fd untouched.
+         *
+         * Both halves of that contrast are unconditional only on the native backend; on JVM the
+         * duplication, and the close on release, depend on junixsocket's optional native support.
+         * See the [UnixFd] class documentation.
          */
         fun adopt(fd: Int): UnixFd
     }


monkopedia-reviewer (approved):
**Approved.** Reviewed at `34d034f7c93744af5a061223cafabdde0d24ee23` in a fresh clone, independent of the author's worktree. Category `docs_only` → tier 1. Every claim below I re-measured rather than read off the PR body.

## 1. Comment-only — verified mechanically, not by eye

Against the merge base (`79bdbe5`), two files, both `commonMain`, +89/-3. Stripping the leading `+`/`-` from every changed line and filtering out `*` / `/**` / `*/` / blank leaves **zero lines**:

```
git diff 79bdbe50..34d034f7 -U0 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
  | sed -E 's/^[+-]//' | grep -vE '^[[:space:]]*(\*|/\*\*|\*/)' | grep -vE '^[[:space:]]*$'
→ (empty)
```

No `actual`, no test, no build file, and `git status --porcelain` was **empty** immediately after `ktlintCheck apiCheck` — so `api/*.api` did not move.

## 2. Every corrected claim holds against this head

`Types.kt` (#208):
- Neither backend closes at scope exit: `nativeMain/UnixFd.kt:61-63` `createCleaner`, `jvmMain/Types.jvm.kt:22` `Cleaner.create()`. ✓
- `JvmUnixFdSupport.checkedDup` — `val support = nativeSupport ?: return fd` (`Types.jvm.kt:86-90`) returns the **caller's** fd. `closeFd` — `?: return false` (`:93-95`), closes nothing, and `OwnedFdState.run()` (`:69-72`) discards the `false`. `release()` → `state.run()`, so a completed `release()` proves nothing. ✓
- The probe is genuinely reflective and genuinely swallowed: `Class.forName("org.newsclub.net.unix.NativeUnixSocket")` + `getDeclaredMethod(...).isAccessible = true`, all inside `runCatching { … }.getOrNull()` (`:210-244`). `supportsFdDuplicationSemantics` is `internal`, so "no public way to ask which mode is in effect" is accurate. ✓
- `UnixFd : Resource : AutoCloseable` (`Resource.kt:36`), so the `use { }` route in the doc is real. ✓

`Message.kt` (#209):
- `isAtEnd`/`copyTo`/`rewind` on JVM (`Message.jvm.kt:186-195`) never read `complete`; native passes it to `sd_bus_message_at_end`/`_copy`/`_rewind` (`Message.native.kt:591-609`). ✓
- JVM `copyTo` does `destination.metadata = metadata` (`:189`) — identity, not contents. ✓
- The "`isAtEnd(true)` can report `true` while containers are still open" claim is demonstrable, not hand-waved: `enterVariant` (`:150-155`) consumes the payload slot via `nextValue` and pushes onto `enteredVariantValues`, so the entered-variant stack can be non-empty with `readIndex >= payload.size`. ✓
- `PeekedType.type` null on JVM has three producers, two of them not end-of-message: `payload.getOrNull(readIndex) ?: …` also fires for a stored Kotlin `null` (`:178`), and `inferSignature(next) ?: …` fires for any object reaching the `else -> null` arm (`:179`, `:254`). Native's is `takeIf { r > 0 }` only (`:581`). ✓
- `isValid`: native `msg != null` (`:586-587`); JVM `metadata.valid` with `Metadata.valid = false` by default (`:40`), `PlainMessage()` never sets it (`:533-535`), and `createErrorReply` sets `valid = false` deliberately (`:471-474`) with `MethodCall.send` reading it back at `:447`. ✓

`Message.kt` creds (#210):
- `withLocalSenderCredentials` has **exactly one** call site — `WireDbusBackend.kt:768`, the received-signal path — and is the **only** producer of these fields anywhere in `src/*Main*`. Swept with `awk`, not `grep`, per the known binary-detection trap on that file (it is currently `file`-detected as UTF-8 text here, but the awk sweep is the safe method regardless). ✓
- The gate is `LocalJvmServiceRegistry.resolveLocalUniqueName(senderName) == null → return this` (`:817`), and the values come from `ProcessHandle.current().pid()` + `UnixSystem()` (`:798-813`) — this process. The backend's own comment at `:795-797` says an external sender's credentials "would require a `GetConnectionCredentials` bus call", which it does not make. So "no JVM configuration reports a remote principal" is a statement about the code, not an overstatement. ✓
- `credsEuid = creds.uid`, `credsEgid = creds.gid` (`:830`, `:832`) from `UnixSystem().uid/.gid` = `getuid()`/`getgid()`; `geteuid` appears nowhere. Native genuinely queries `SD_BUS_CREDS_EUID`/`_EGID` via `sd_bus_creds_get_euid`/`_egid` (`Message.native.kt:633-664`), so "only the native backend reports a genuine effective UID" is right. ✓

I looked specifically for over-claiming in the other direction and did not find a "never"/"no configuration" that the code permits an exception to.

## 3. Re-measured build claims

`ktlintCheck apiCheck :dokkaGeneratePublicationHtml` (JDK 21) → **BUILD SUCCESSFUL**, working tree clean afterwards. Dokka emitted **exactly 8** `Couldn't resolve link` warnings, all in untouched `jvmMain` (`DBusWireConnection.kt` ×4, `WireDbusBackend.kt` ×4); none names `Message.kt` or `Types.kt`.

**And I made that detector fail on purpose rather than trusting the absence.** Replacing `[Resource]` with `[ZzzProbeTypeThatDoesNotExist]` in the `UnixFd` KDoc produced 4 new warnings — `in Types.kt/com.monkopedia.sdbus.UnixFd`, `in Types.kt`, `in Types.jvm.kt/…`, `in Types.jvm.kt` — then reverted; `git status --porcelain` empty and no `Zzz` string left in the tree. The instrument is live, so the 8-and-none-of-ours reading is a real absence.

One thing worth recording for whoever reads this later: that probe run **still exited `BUILD SUCCESSFUL`** with a broken link. `:dokkaGeneratePublicationHtml` going green is *not* link verification — the warning list is the instrument. The PR used the warning list, so its evidence is sound; the framing "BUILD SUCCESSFUL … the guard that matters" undersells which half of the measurement was doing the work.

Rendered output spot-checked independently: `-unix-fd/index.html` carries the new paragraphs, `-message/creds-euid.html` "does not report an effective UID", `-peeked-type/type.html` "overloaded on the JVM backend".

## 4. Doc-only discipline, and no collision with #236

#236's two paragraphs (the throw contract, and direct-connection `sender` being peer-asserted with the #199 link) survive verbatim at `Message.kt:218-232`; the new #210 paragraph sits below them and repeats none of it. No contradiction — #236 leaves *what should be reported* on a direct connection unspecified pending #199; this PR states *what is in fact reported* on JVM. Different questions.

#245, #246 and #247 all exist, are `agent-workable`, and their bodies describe the deferred behaviour work with the same file:line citations (I read all three). `docs/BACKENDS.md`, `README.md`, `samples/**` and `codegen/**` are absent from the diff, as claimed.

CI green on this head: 16 checks pass, `full-tests-x64` included; `cleaner-soak` / `nightly-stress` are the nightly-gated jobs and report `skipping` on PRs as normal.

## Non-blocking notes

1. **The published `native` tab still shows the uncorrected-shape text.** Dokka renders three doc blocks on the `UnixFd` page: `common` and `jvm` inherit the `expect` KDoc (confirmed by the probe warning firing for `Types.jvm.kt` as well as `Types.kt`), but the native `actual` has its own KDoc at `nativeMain/UnixFd.kt:39-47`, so the native block shows only "closes the fd when the UnixFd is released" — no GC-backstop paragraph, no fd-table-exhaustion warning. Nothing there is *false*, but this is the mirror image of the drift #208 was filed about: the fix reached the `expect` and now the `actual` is the stale copy. Worth a sweep, not worth holding this PR.

2. **`isValid`'s "is unsettled" has no issue link**, unlike the creds paragraph which links #199. #246 now tracks exactly that question — linking it would let a reader follow the open decision instead of stopping at the word.

3. Minor precision: "The native backend always `dup(2)`s … and always `close(2)`s on release" — native `checkedDup` also returns the fd unchanged when `fd < 0` (`UnixFd.kt:99-102`), and `release()` only closes for `fd >= 0`. Reads fine in the supported-vs-degraded contrast it is drawing; strictly the invalid-fd case is an exception on both backends.

## On the #199 reopen (outside the diff, assessed as asked)

**The reopen was correct, and the original close was a mechanical error rather than a judgment call.** #236's body carries only `Refs #199`, no closing keyword, and the close event reports `state_reason: null`, `commit_id: null` with actor `monkopedia-reviewer` — so this was an explicit close, not a keyword-driven auto-close on merge. It landed 32 seconds before `monkopedia-coder`'s comment saying the decision is untouched, on an issue still labelled `needs-decision`, whose decision has not been made, and which **shipped KDoc on `main` cites to the reader as an open question**. A closed issue behind that link is a worse artifact than an open one. Correctly reopened, correctly kept out of the diff.

Issue #249, opened 2026-08-08
:codegen: hostile or empty D-Bus names produce uncompilable output or an unhelpful crash


monkopedia-coder:
Split out of #194, which fixed the two availability defects (unbounded entity expansion, unbounded signature recursion) and deliberately left these. Neither is an availability problem — they fail one task immediately — so they are diagnostics quality, not hardening.

All three were measured against `main` while working #194.

**1. A name that is not a valid D-Bus name passes straight into KotlinPoet.** A method named `` Ev`il `` generates `public suspend fun ``ev`il``(…)`: KotlinPoet backtick-wraps identifiers but does not validate their contents, so the inner backtick breaks out and the file will not compile. Same for a newline, a space, or `$` in identifier position. Generation reports success and the *compiler* reports the failure, pointing at generated code rather than at the input that caused it.

**2. An empty `name=""` crashes with a message that says nothing.** `<interface name="">` and `<method name="">` both fail with `java.util.NoSuchElementException: Char sequence is empty.` — no interface, no member, no file.

**3. The Gradle plugin exits the process on a clikt error.** `SdbusGenerationTask` calls `Xml2Kotlin().main(args)` in-process, and clikt's `main()` handles a `CliktError` by calling `exitProcess`. Inside a Gradle worker/daemon that is a harder failure than a task failure. Not attacker-triggered — the plugin builds the argument list itself — but `parse(args)` would throw where `main(args)` exits.

**Suggested default** (this is `agent-workable` with this as the stated approach, not a question): validate interface and member names against the D-Bus specification's own grammar as the XML is parsed — elements of `[A-Za-z_][A-Za-z0-9_]*`, at least two elements for an interface name, 255 characters maximum — and report which name was rejected. That subsumes (2), turns (1) into a clear error before any code is emitted, and is the same "hold the input to the spec's own limits" shape #194 used. Switch the plugin to `parse(args)` for (3).

Worth checking against the golden fixtures first: every name in `codegen/src/test/resources` is already a valid identifier and every interface name has at least two elements (I checked while working #194), so validation should not move a golden — but that is the thing to verify before trusting it.

PR #250, opened 2026-08-08
fix(codegen): bound what hostile introspection XML can make the generator do

### LINKED ISSUE #194: :codegen / Gradle plugin: hostile introspection XML DoSes the build (unbounded entity expansion, StackOverflow, uncaught parse errors)

monkopedia-coder:
Fixes #194

`:codegen` turns introspection XML supplied by the service being introspected into Kotlin, and the Gradle plugin does that **inside a consumer's build** (`SdbusGenerationTask` calls `Xml2Kotlin().main(...)` in-process). Two inputs therefore cost build availability rather than failing: the parse never came back, or the recursion ran off the stack.

## What the parser actually did by default — the question #194 left open

The issue's verification comment flagged the 10 MB figure as `UNRESOLVED-NEEDS-BUILD`, on the grounds that a JDK default (`jdk.xml.entityExpansionLimit` = 64,000) might bound expansion without appearing in `XmlFormat.kt`. **It does not, and here is why: there is no JAXP on this path.** `xmlutil` 1.0.1's `core-jvm` jar contains no StAX bridge at all — the JVM reader is xmlutil's own `KtXmlReader` (`xmlStreaming.newReader("<node/>")` returns `nl.adaptivity.xmlutil.core.KtXmlReader`), a pure-Kotlin parser with its own `EntityMap`, so no `jdk.xml.*` property applies. Measured, one document per level:

| entity nesting levels | input | decoded attribute value | time |
|---|---|---|---|
| 2 | 233 B | 100 chars | 1 ms |
| 4 | 327 B | 10,000 chars | 2 ms |
| 6 | 421 B | 1,000,000 chars | 30 ms |

Exactly ten-fold per level, no knee, no limit — so the issue's 10 MB figure at seven levels is confirmed, and it keeps going. An entity that references itself is worse: 130 bytes that **never terminate** (still running at a watchdog; the reader re-injects the entity into its own replacement text through `InjectingInOutBuffer`, which has no depth counter). And `<arg type="aaa…s">` overflows the stack at 5,000 array codes (1,000 was fine).

**Where the exposure is narrower than filed**, all measured rather than assumed:

- **External entities really are rejected**, as the issue said — `<!ENTITY xxe SYSTEM "file:///etc/hostname">` fails with `XmlException: Unexpected content in document type declaration: '>'`. So this is availability only; there is no file-disclosure path to fix here.
- **There is no configuration knob to reach for.** `KtXmlReader.expandEntities` defaults to `false`, but that governs only whether entity references surface as `ENTITY_REF` events in *content* — attribute values call `pushEntity(true)` unconditionally, and the issue's own vector is an attribute (`<annotation value="&g;"/>`). Setting it would have looked like a fix and changed nothing.
- **Item 4 of the issue (hostile names → uncompilable output) is not fixed here** — see "Not in scope" below.

## The fix

**Refuse a DTD internal subset** (`XmlFormat.kt`). It is the only route to a declared entity, introspection XML has no legitimate use for one, and what a real `dbus-daemon` puts on an `Introspect` reply is the *external* `introspect.dtd` doctype, which still parses. That covers the whole expansion class — ten-fold nesting, quadratic blowup and self-reference alike — without inventing a threshold for something no real input does.

Deciding that means reading the prolog, and the scan that does it has been through two rounds of review, each of which found it passing something it should have refused:

1. It gave up on the first character it did not recognise, and `U+FEFF` is not whitespace — so a byte order mark, which `File.readText()` keeps and the parser accepts, walked the whole document past the check. **The bail is now inverted: a prolog the scan cannot read as far as the root element is refused**, not assumed to declare nothing. That closed a second member of the family nobody had found (`U+0085`, which `Char.isWhitespace()` is false for and the parser treats as ignorable).
2. It stopped at the *first* doctype's `>`, so **a second `<!DOCTYPE` later in the prolog was never looked at — and the parser honours it.** Two doctypes is not well-formed XML, but `KtXmlReader` declares the second one's entities anyway, so the expansion class came back in full behind the exact external doctype every real service sends. The doctype helper now reports where the declaration ends instead of deciding the answer, the loop continues past it, and `null` — go ahead — can only come from the root-element branch. That is what the KDoc already claimed; the code now matches it.

**Hold signatures to the D-Bus specification's own 255-character maximum** (`NamingManager.kt`). This is the same choice as #190/#204 — use the specification's number so it is a conformance check rather than an invented threshold, and `dbus-daemon` enforces the same one on every message it relays. It also makes the recursion bounded *structurally* rather than adding a depth counter beside it: every level of `buildType` consumes at least one character of the signature, so a signature within the limit cannot nest deeper than 255, which is comfortably inside the stack (measured: 1,000 levels resolve fine). The recursive calls go to a private `buildValidatedType`, since they descend into suffixes of an already-validated signature and re-checking would only re-assert the same fact.

Parsing now goes through one entry point, `parseIntrospectionXml`, so the golden fixtures and the shipped CLI are held to the same limit — `GoldenFixture`, `TestXmlSupport` and `Xml2Kotlin` all call it.

## What the guard covers, and what it does not

Stated plainly, because a parser guard that reads as a guarantee is worse than none.

**Covered, each measured on this branch:**

- An internal subset anywhere in the prolog: first thing in the document, behind an XML declaration, behind a byte order mark, behind comments and processing instructions, and behind one or more earlier doctypes — including the external `introspect.dtd` doctype a real `dbus-daemon` sends.
- Anything the scan cannot read to the root element is **refused rather than passed on**: leading text, `<!` with a space or a lower-case name, a CDATA section in the prolog, a second byte order mark, UTF-16 byte order mark bytes read as UTF-8, an unterminated comment, processing instruction or doctype. These are already-malformed documents the parser would reject; refusing them earlier is what keeps an unmodelled shape from becoming the next bypass.

**Not covered — say so rather than imply otherwise:**

- **This is not a general hardening of xmlutil.** It refuses the one declaration route measured to reach unbounded expansion. Nothing else about the parser is guarded, and I did not audit it.
- **It is not a well-formedness check.** Whatever the scan passes still goes to the parser, which decides.
- **Nothing bounds document size, element count or nesting depth.** I found no amplification there — a document with no entity declarations does work linear in its length — but there is no cap, so a large document is still a large parse.
- **A doctype after the root element is not refused**, because it is not honoured: measured, a declaration following the root is never applied, and an entity referenced inside the root fails with `Unknown entity`. That measurement is what makes the root element a safe place to stop, and it is pinned by a test so a parser change cannot quietly invalidate it.
- **The scan is a hand-rolled reading of the prolog, and I do not claim it is exhaustive.** Two rounds of review each found a shape it got wrong. What can be claimed is that it is now fail-closed by construction, so the next unmodelled shape should land in the refusal branch rather than in the parser — and the suite has a case that fails if that default flips back.
- Item 4 of #194, and the two diagnostics notes in item 3, are untouched — see "Not in scope".

## Red first

Each fix commit is preceded by a tests-only commit that fails against the code as it then stood.

**The second-doctype hole** — the last one open. Tests only at `c8b0a08`, `./gradlew :codegen:test --tests "com.monkopedia.sdbus.CodegenHostileXmlTest"`, verbatim from that commit's JUnit XML: **13 tests, 3 failed**, suite 5.159 s:

```
FAIL  0.011  secondDoctypeBehindACommentAndAProcessingInstruction_isStillRefused()
               expected the entity declaration to be refused, but the document parsed
FAIL  0.004  secondDoctypeBehindARealIntrospectReplyDoctype_isStillRefused()
               expected the entity declaration to be refused, but the document parsed
FAIL  5.004  selfReferentialEntityInASecondDoctype_isStillRefused()
               parse had not finished after 5000ms
PASS  0.002  externalDoctypeOfARealIntrospectReply_stillParses()
PASS  0.002  entityDeclaredAfterTheRootElement_isNotHonouredByTheParser()
PASS  0.001  prologTheScanCannotRead_isRefusedRatherThanAssumedToDeclareNothing()
PASS  0.002  documentBehindAByteOrderMark_stillParses()
   … 6 more passing, including the two signature cases
```

After `fbfca93`, all 13 pass and the self-referential one goes from a 5 s watchdog abort to **0.001 s**:

```
PASS  0.001  secondDoctypeBehindARealIntrospectReplyDoctype_isStillRefused()
PASS  0.001  selfReferentialEntityInASecondDoctype_isStillRefused()
PASS  0.000  secondDoctypeBehindACommentAndAProcessingInstruction_isStillRefused()
PASS  0.002  externalDoctypeOfARealIntrospectReply_stillParses()
PASS  0.002  entityDeclaredAfterTheRootElement_isNotHonouredByTheParser()
PASS  0.003  prologTheScanCannotRead_isRefusedRatherThanAssumedToDeclareNothing()
```

The positive controls are what make that worth something: they pass at the red commit too, so the suite discriminates the defect rather than the input shape. A fix that refused all doctypes, or refused documents with a byte order mark, would fail them.

The two earlier rounds are on the same pattern — `51825cd` (3 of 5 failing, including the 30 s watchdog abort and `StackOverflowError`) before the first fix, and `69ca3fc` (3 of 9 failing, byte order mark and unreadable prolog) before the second. Both are quoted in the review thread.

**The fixtures are deliberately bounded so they cannot take out the job meant to catch them.** The expanding one stops at four levels (10,000 characters, 2 ms) because the levels are a multiplier and the *shape* is what the assertion is about, not the size; and every parse runs on a daemon thread under a 5 s watchdog, so a regression fails the suite in bounded time instead of hanging it. That is what produced the `5.004` line above.

## Verification

Counted from fresh JUnit XML after `find . -type d -name test-results -prune -exec rm -rf {} +`, JDK 21, rebased on `75e0aa5`:

- `./gradlew :codegen:test` — **99 tests, 0 failures, 0 errors, 0 skipped** across 6 suite files (`CodegenHostileXmlTest` 13, `GeneratorTest` 69, `CodegenNamingAndTypeTest` 5, `CodegenParsingEdgeCaseTest` 4, `Xml2KotlinOptionsTest` 7, `CompilationIntegrationTest` 1). Whole `CodegenHostileXmlTest` suite in **0.146 s**, slowest case 0.066 s — nowhere near the watchdog.
- `./gradlew :compile_test:build` — BUILD SUCCESSFUL. It carries no test suites; it is the gate that compiles generated output.
- **No golden moved.** `GeneratorTest` is 69/69 and `git diff origin/main...HEAD -- codegen/src/test/resources` is empty. Valid XML generates exactly what it did before; the change only refuses inputs. The 23 checked-in fixtures are the regression corpus and all of them still parse.
- `./gradlew ktlintCheck apiCheck` — both pass. **No api file moved**: `git diff origin/main...HEAD -- codegen/api plugin/api api` is empty. `parseIntrospectionXml` is `internal`, `buildValidatedType` and the prolog scan are `private`, and the published `buildType(Map, String, String, String)` keeps its exact signature (it is now the validating wrapper). Nothing binary-visible changed on either published artifact.

No D-Bus bus is involved: `:codegen` and `:compile_test` are pure JVM.

## What is now rejected that previously "worked"

A document declaring its own XML entities, a document whose prolog cannot be read to its root element, and a type signature over 255 characters. All previously "worked" in the sense that one hung or exhausted the heap, one was already malformed, and one raised `StackOverflowError`; none is a usable outcome, and no conforming D-Bus service can produce any of them.

## Not in scope

Item 4 of #194 — a name containing a backtick, newline, space or `$` passes into KotlinPoet and yields output that will not compile — is untouched. It is a bounded, immediate, single-task failure with an unhelpful message rather than an availability problem, and fixing it properly means validating names against the D-Bus name grammar, which is a separate change with its own golden-fixture risk. Split out as #249, together with the two diagnostics notes in the issue's item 3 (an empty `name=""` still fails as `NoSuchElementException: Char sequence is empty`, and the plugin's in-process `clikt.main()` still exits the process on a `CliktError`); neither is attacker-triggered.

## Severity, stated honestly

Low-to-medium, availability only, and it needs a developer to point the generator at a hostile or compromised service and build. There is no code execution, file disclosure or arbitrary write here — I probed for the first two and could not reach them. What makes it worth fixing is that the failure mode is a *hang or an OOM in someone else's build*, which is not diagnosable from the outside, and that the entity case is amplification: a couple of hundred bytes of input against unbounded work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   emitted statement changes — so this is additive documentation rather than a behavior change.
   XML that documents nothing generates exactly what it did before. (#160)
 
+### Fixed — codegen
+
+- **Introspection XML can no longer make the generator do unbounded work.** Two inputs used to cost
+  build availability rather than failing: a `<!DOCTYPE>` internal subset declaring XML entities (the
+  parser expanded them with no limit — ten-fold per nesting level, and an entity referencing itself
+  never terminated at all), and a type signature with thousands of array type codes (unbounded
+  recursion, `StackOverflowError`). An internal subset is now refused outright — introspection XML
+  has no use for one, and the external `introspect.dtd` doctype a real `dbus-daemon` sends still
+  parses — and a signature is held to the D-Bus specification's own 255-character maximum, which
+  bounds the recursion by construction. Both report which limit they hit. Generated output for
+  valid XML is unchanged. (#194)
+
 ### Changed — codegen
 
 - **Standard D-Bus annotations are now mapped onto the runtime vtable/proxy flags.** The generated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,10 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   never terminated at all), and a type signature with thousands of array type codes (unbounded
   recursion, `StackOverflowError`). An internal subset is now refused outright — introspection XML
   has no use for one, and the external `introspect.dtd` doctype a real `dbus-daemon` sends still
-  parses — and a signature is held to the D-Bus specification's own 255-character maximum, which
-  bounds the recursion by construction. Both report which limit they hit. Generated output for
-  valid XML is unchanged. (#194)
+  parses, byte order mark and all — and a signature is held to the D-Bus specification's own
+  255-character maximum, which bounds the recursion by construction. A document whose prolog cannot
+  be read as far as its root element is refused too, rather than assumed to declare nothing. All
+  three report what they refused. Generated output for valid XML is unchanged. (#194)
 
 ### Changed — codegen
 
--- a/codegen/src/main/kotlin/NamingManager.kt
+++ b/codegen/src/main/kotlin/NamingManager.kt
@@ -283,7 +283,33 @@ private fun MutableMap<String, NamingType>.buildSingleType(
     this[type.type] = type
 }
 
+/**
+ * The D-Bus specification's own maximum length for a type signature. It is what bounds the
+ * recursion in [buildValidatedType]: every level consumes at least one character of the signature,
+ * so a signature within the limit cannot nest more than 255 deep. `dbus-daemon` enforces the same
+ * number on every message it relays, so nothing a conforming service can describe is refused —
+ * the longest signature in the checked-in fixtures is 15 characters.
+ */
+private const val MAX_SIGNATURE_LENGTH = 255
+
 fun MutableMap<String, NamingType>.buildType(pkg: String, name: String, type: String): NamingType {
+    require(type.length <= MAX_SIGNATURE_LENGTH) {
+        "Type signature for '$name' is ${type.length} characters, over the D-Bus maximum of " +
+            "$MAX_SIGNATURE_LENGTH"
+    }
+    return buildValidatedType(pkg, name, type)
+}
+
+/**
+ * [buildType] once its signature is known to be within [MAX_SIGNATURE_LENGTH]. The recursive calls
+ * are all to this rather than back to [buildType]: they descend into suffixes of an already
+ * validated signature, so re-checking each one would be re-asserting the same fact.
+ */
+private fun MutableMap<String, NamingType>.buildValidatedType(
+    pkg: String,
+    name: String,
+    type: String
+): NamingType {
     (this[type] as? SimpleType)?.let { return it }
     if (type.isEmpty()) {
         return this.getOrPut(type) {
@@ -320,12 +346,12 @@ fun MutableMap<String, NamingType>.buildType(pkg: String, name: String, type: St
             if (type[1] == '{') {
                 // Map
                 val keyTypeStr = type.substring(2)
-                val keyType = buildType(pkg, name + "Key", keyTypeStr)
+                val keyType = buildValidatedType(pkg, name + "Key", keyTypeStr)
                 require(keyTypeStr.startsWith(keyType.type)) {
                     "Invalid type parsed out, expected $keyTypeStr but got ${keyType.type}"
                 }
                 val valueTypeStr = keyTypeStr.substring(keyType.type.length)
-                val valueType = buildType(pkg, name + "Value", valueTypeStr)
+                val valueType = buildValidatedType(pkg, name + "Value", valueTypeStr)
                 require(valueTypeStr.startsWith(valueType.type)) {
                     "Invalid type parsed out, expected $valueTypeStr but got ${valueType.type}"
                 }
@@ -344,7 +370,7 @@ fun MutableMap<String, NamingType>.buildType(pkg: String, name: String, type: St
             } else {
                 // List
                 val valueTypeStr = type.substring(1)
-                val valueType = buildType(pkg, name + "Value", valueTypeStr)
+                val valueType = buildValidatedType(pkg, name + "Value", valueTypeStr)
                 require(valueTypeStr.startsWith(valueType.type)) {
                     "Invalid type parsed out, expected $valueTypeStr but got ${valueType.type}"
                 }
@@ -359,7 +385,7 @@ fun MutableMap<String, NamingType>.buildType(pkg: String, name: String, type: St
             val types = sequence {
                 var currentType = type.substring(1)
                 while (currentType.isNotEmpty() && currentType[0] != ')') {
-                    val nextType = buildType(pkg, name + "${index++}", currentType)
+                    val nextType = buildValidatedType(pkg, name + "${index++}", currentType)
                     require(currentType.startsWith(nextType.type)) {
                         "Invalid type parsed out, expected $currentType but got ${nextType.type}"
                     }
--- a/codegen/src/main/kotlin/Xml2Kotlin.kt
+++ b/codegen/src/main/kotlin/Xml2Kotlin.kt
@@ -66,8 +66,6 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with sdbus-kotlin. If not, see <http://www.gnu.org/licenses/>.
  */
-@file:OptIn(ExperimentalSerializationApi::class)
-
 package com.monkopedia.sdbus
 
 import com.github.ajalt.clikt.core.CliktCommand
@@ -78,8 +76,6 @@ import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.types.file
 import java.io.File
-import kotlinx.serialization.ExperimentalSerializationApi
-import kotlinx.serialization.decodeFromString
 
 fun main(args: Array<String>) = Xml2Kotlin().main(args)
 
@@ -102,7 +98,7 @@ class Xml2Kotlin : CliktCommand() {
 
     override fun run() {
         val packageOverride = outputPackage?.trim()?.takeUnless { it.isEmpty() }
-        val xml = introspectionXml.decodeFromString<XmlRootNode>(input.readText())
+        val xml = parseIntrospectionXml(input.readText())
         if (!keep) {
             output.deleteRecursively()
         }
--- a/codegen/src/main/kotlin/XmlFormat.kt
+++ b/codegen/src/main/kotlin/XmlFormat.kt
@@ -24,6 +24,7 @@ package com.monkopedia.sdbus
 
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.decodeFromString
 import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi
 import nl.adaptivity.xmlutil.QName
 import nl.adaptivity.xmlutil.XmlReader
@@ -37,9 +38,10 @@ import nl.adaptivity.xmlutil.serialization.XmlValue
 import nl.adaptivity.xmlutil.serialization.structure.XmlDescriptor
 
 /**
- * The single parser for D-Bus introspection XML, shared by the [Xml2Kotlin] CLI and the tests so
- * the golden fixtures assert against the configuration the tool actually ships. Unknown content
- * is dropped rather than failing, so vendor extensions don't break generation.
+ * The serialization format the [Xml2Kotlin] CLI and the tests share, so the golden fixtures assert
+ * against the configuration the tool actually ships. Unknown content is dropped rather than
+ * failing, so vendor extensions don't break generation. Call [parseIntrospectionXml] rather than
+ * this directly — it is what applies the limits hostile XML is held to.
  *
  * `XML.compat` is deprecated in xmlutil 1.0, but its replacements deliberately change the parsing
  * defaults and xmlutil offers no non-deprecated route back to the old ones. Switching is a
@@ -59,6 +61,77 @@ internal val introspectionXml: XML = XML.compat {
         ): List<ParsedData<*>> = emptyList()
     }
 }
+
+/**
+ * Parses D-Bus introspection XML into the model the generators consume. The one entry point the CLI
+ * and the tests both use, so what the golden fixtures assert against and what the tool ships are
+ * held to the same limit.
+ *
+ * Introspection XML is supplied by the service being introspected, and `:codegen` runs inside a
+ * developer's build (the Gradle plugin calls it in-process), so a hostile document costs build
+ * availability rather than anything at runtime.
+ */
+internal fun parseIntrospectionXml(xml: String): XmlRootNode {
+    require(!declaresDtdInternalSubset(xml)) {
+        "Introspection XML declares a DTD internal subset (<!DOCTYPE ... [ ... ]>). " +
+            "Introspection XML has no use for one, and the entities it can declare are expanded " +
+            "without any limit, so the declaration is refused rather than parsed."
+    }
+    return introspectionXml.decodeFromString(xml)
+}
+
+/**
+ * Whether [xml] opens a DTD internal subset in its prolog, i.e. `<!DOCTYPE node [ ... ]>`.
+ *
+ * The subset is the only place introspection XML can declare XML entities, and xmlutil's parser
+ * expands them with no limit of any kind: measured against xmlutil 1.0.1, a 421-byte document
+ * nesting six ten-fold entities decodes to a 1,000,000-character attribute value in 30ms — each
+ * further level multiplies by ten — and `<!ENTITY a "&a;">` never terminates at all. There is no
+ * configuration knob for that (the JVM path is xmlutil's own `KtXmlReader`, not JAXP, so the
+ * `jdk.xml.entityExpansionLimit` default does not apply), and `expandEntities = false` would not
+ * help either because attribute values are expanded unconditionally.
+ *
+ * Refusing the subset outright is both the complete fix and the narrowest one: it is the only route
+ * to a declared entity, external entities the parser already rejects on its own, and what a real
+ * service emits is at most the external `introspect.dtd` doctype, which still parses.
+ *
+ * Only the prolog is scanned — the XML declaration, comments and processing instructions that may
+ * precede the doctype are skipped over — so a literal `<!DOCTYPE` in a comment or in element
+ * content is not mistaken for a declaration.
+ */
+private fun declaresDtdInternalSubset(xml: String): Boolean {
+    var i = 0
+    while (i < xml.length) {
+        when {
+            xml[i].isWhitespace() -> i++
+            xml.startsWith("<!--", i) -> i = xml.endOf("-->", i + 4) ?: return false
+            xml.startsWith("<?", i) -> i = xml.endOf("?>", i + 2) ?: return false
+            xml.startsWith("<!DOCTYPE", i) -> return doctypeOpensSubset(xml, i + "<!DOCTYPE".length)
+            // The root element, or something the parser will reject on its own.
+            else -> return false
+        }
+    }
+    return false
+}
+
+/** Whether the doctype declaration running from [start] in [xml] opens an internal subset. */
+private fun doctypeOpensSubset(xml: String, start: Int): Boolean {
+    var i = start
+    while (i < xml.length) {
+        when (val c = xml[i]) {
+            '[' -> return true
+            '>' -> return false
+            // A quoted public or system literal; '[' and '>' inside it mean neither of the above.
+            '\'', '"' -> i = xml.endOf(c.toString(), i + 1) ?: return false
+            else -> i++
+        }
+    }
+    return false
+}
+
+/** The index just past the next [token] at or after [from], or null if there isn't one. */
+private fun String.endOf(token: String, from: Int): Int? =
+    indexOf(token, from).takeIf { it >= 0 }?.plus(token.length)
 
 @Serializable
 @XmlSerialName("node")
--- a/codegen/src/main/kotlin/XmlFormat.kt
+++ b/codegen/src/main/kotlin/XmlFormat.kt
@@ -88,17 +88,27 @@ internal fun parseIntrospectionXml(xml: String): XmlRootNode {
  * `expandEntities = false` would not help either because attribute values are expanded
  * unconditionally.
  *
- * Refusing the subset outright is both the complete fix and the narrowest one: it is the only route
- * to a declared entity, external entities the parser already rejects on its own, and what a real
- * service emits is at most the external `introspect.dtd` doctype, which still parses.
+ * Refusing the subset outright covers that whole class and is the narrowest thing that does: it is
+ * the only route to a declared entity, external entities the parser already rejects on its own, and
+ * what a real service emits is at most the external `introspect.dtd` doctype, which still parses.
+ * The cover is only as good as this scan's reading of the prolog, which is why what it cannot read
+ * is refused rather than passed on.
  *
  * Only the prolog is read — the byte order mark, XML declaration, comments and processing
  * instructions that may precede the doctype are stepped over — so a literal `<!DOCTYPE` in a
  * comment or in element content is not mistaken for a declaration. **A prolog this cannot read is
  * refused rather than assumed to declare nothing**: the scan is what decides whether entities are
  * possible, so treating what it does not model as safe is how a leading byte order mark got past an
- * earlier version of it. The only thing that legitimately ends the scan is the root element, which
- * is a `<` opening neither a markup declaration nor a processing instruction.
+ * earlier version of it.
+ *
+ * The one thing that legitimately ends the scan is the root element — a `<` opening neither a
+ * markup declaration nor a processing instruction — and that is the only branch this returns null
+ * from. A doctype without an internal subset is read *past* rather than stopped at, because the
+ * parser honours a second `<!DOCTYPE` later in the prolog even though two of them is not
+ * well-formed; stopping at the first one's `>` let a subset behind the doctype every real service
+ * sends declare entities freely. Stopping at the root element is safe for the reason a second
+ * doctype is not: measured against xmlutil 1.0.1, a declaration after the root element is never
+ * applied, and an entity referenced inside the root fails with `Unknown entity` instead.
  */
 private fun prologRefusal(xml: String): String? {
     var i = if (xml.startsWith(BYTE_ORDER_MARK)) 1 else 0
@@ -107,7 +117,11 @@ private fun prologRefusal(xml: String): String? {
             xml[i].isWhitespace() -> i++
             xml.startsWith("<!--", i) -> i = xml.endOf("-->", i + 4) ?: return UNREADABLE_PROLOG
             xml.startsWith("<?", i) -> i = xml.endOf("?>", i + 2) ?: return UNREADABLE_PROLOG
-            xml.startsWith("<!DOCTYPE", i) -> return doctypeRefusal(xml, i + "<!DOCTYPE".length)
+            xml.startsWith(DOCTYPE, i) -> {
+                val end = endOfDoctypeBody(xml, i + DOCTYPE.length) ?: return UNREADABLE_PROLOG
+                if (xml[end] == '[') return INTERNAL_SUBSET
+                i = end + 1
+            }
             xml[i] == '<' && !xml.startsWith("<!", i) -> return null // The root element.
             else -> return UNREADABLE_PROLOG
         }
@@ -115,19 +129,22 @@ private fun prologRefusal(xml: String): String? {
     return UNREADABLE_PROLOG
 }
 
-/** [prologRefusal] for the doctype declaration running from [start] in [xml]. */
-private fun doctypeRefusal(xml: String, start: Int): String? {
+/**
+ * The index of the character ending the doctype declaration whose body starts at [start] in [xml] —
+ * either the `[` opening an internal subset or the `>` closing the declaration — or null if it is
+ * unterminated. Quoted public and system literals are stepped over, since neither character means
+ * either of those things inside one.
+ */
+private fun endOfDoctypeBody(xml: String, start: Int): Int? {
     var i = start
     while (i < xml.length) {
         when (val c = xml[i]) {
-            '[' -> return INTERNAL_SUBSET
-            '>' -> return null
-            // A quoted public or system literal; '[' and '>' inside it mean neither of the above.
-            '\'', '"' -> i = xml.endOf(c.toString(), i + 1) ?: return UNREADABLE_PROLOG
+            '[', '>' -> return i
+            '\'', '"' -> i = xml.endOf(c.toString(), i + 1) ?: return null
             else -> i++
         }
     }
-    return UNREADABLE_PROLOG
+    return null
 }
 
 /** The index just past the next [token] at or after [from], or null if there isn't one. */
@@ -135,6 +152,8 @@ private fun String.endOf(token: String, from: Int): Int? =
     indexOf(token, from).takeIf { it >= 0 }?.plus(token.length)
 
 private const val BYTE_ORDER_MARK = '\uFEFF'
+
+private const val DOCTYPE = "<!DOCTYPE"
 
 private const val INTERNAL_SUBSET =
     "Introspection XML declares a DTD internal subset (<!DOCTYPE ... [ ... ]>). Introspection " +
--- a/codegen/src/main/kotlin/XmlFormat.kt
+++ b/codegen/src/main/kotlin/XmlFormat.kt
@@ -72,66 +72,79 @@ internal val introspectionXml: XML = XML.compat {
  * availability rather than anything at runtime.
  */
 internal fun parseIntrospectionXml(xml: String): XmlRootNode {
-    require(!declaresDtdInternalSubset(xml)) {
-        "Introspection XML declares a DTD internal subset (<!DOCTYPE ... [ ... ]>). " +
-            "Introspection XML has no use for one, and the entities it can declare are expanded " +
-            "without any limit, so the declaration is refused rather than parsed."
-    }
+    prologRefusal(xml)?.let { throw IllegalArgumentException(it) }
     return introspectionXml.decodeFromString(xml)
 }
 
 /**
- * Whether [xml] opens a DTD internal subset in its prolog, i.e. `<!DOCTYPE node [ ... ]>`.
+ * Why [xml]'s prolog must not be handed to the parser, or null to go ahead.
  *
- * The subset is the only place introspection XML can declare XML entities, and xmlutil's parser
- * expands them with no limit of any kind: measured against xmlutil 1.0.1, a 421-byte document
- * nesting six ten-fold entities decodes to a 1,000,000-character attribute value in 30ms — each
- * further level multiplies by ten — and `<!ENTITY a "&a;">` never terminates at all. There is no
- * configuration knob for that (the JVM path is xmlutil's own `KtXmlReader`, not JAXP, so the
- * `jdk.xml.entityExpansionLimit` default does not apply), and `expandEntities = false` would not
- * help either because attribute values are expanded unconditionally.
+ * A `<!DOCTYPE ... [ ... ]>` internal subset is the only place introspection XML can declare XML
+ * entities, and xmlutil's parser expands them with no limit of any kind: measured against xmlutil
+ * 1.0.1, a 421-byte document nesting six ten-fold entities decodes to a 1,000,000-character
+ * attribute value in 30ms — each further level multiplies by ten — and `<!ENTITY a "&a;">` never
+ * terminates at all. There is no configuration knob for that (the JVM path is xmlutil's own
+ * `KtXmlReader`, not JAXP, so the `jdk.xml.entityExpansionLimit` default does not apply), and
+ * `expandEntities = false` would not help either because attribute values are expanded
+ * unconditionally.
  *
  * Refusing the subset outright is both the complete fix and the narrowest one: it is the only route
  * to a declared entity, external entities the parser already rejects on its own, and what a real
  * service emits is at most the external `introspect.dtd` doctype, which still parses.
  *
- * Only the prolog is scanned — the XML declaration, comments and processing instructions that may
- * precede the doctype are skipped over — so a literal `<!DOCTYPE` in a comment or in element
- * content is not mistaken for a declaration.
+ * Only the prolog is read — the byte order mark, XML declaration, comments and processing
+ * instructions that may precede the doctype are stepped over — so a literal `<!DOCTYPE` in a
+ * comment or in element content is not mistaken for a declaration. **A prolog this cannot read is
+ * refused rather than assumed to declare nothing**: the scan is what decides whether entities are
+ * possible, so treating what it does not model as safe is how a leading byte order mark got past an
+ * earlier version of it. The only thing that legitimately ends the scan is the root element, which
+ * is a `<` opening neither a markup declaration nor a processing instruction.
  */
-private fun declaresDtdInternalSubset(xml: String): Boolean {
-    var i = 0
+private fun prologRefusal(xml: String): String? {
+    var i = if (xml.startsWith(BYTE_ORDER_MARK)) 1 else 0
     while (i < xml.length) {
         when {
             xml[i].isWhitespace() -> i++
-            xml.startsWith("<!--", i) -> i = xml.endOf("-->", i + 4) ?: return false
-            xml.startsWith("<?", i) -> i = xml.endOf("?>", i + 2) ?: return false
-            xml.startsWith("<!DOCTYPE", i) -> return doctypeOpensSubset(xml, i + "<!DOCTYPE".length)
-            // The root element, or something the parser will reject on its own.
-            else -> return false
+            xml.startsWith("<!--", i) -> i = xml.endOf("-->", i + 4) ?: return UNREADABLE_PROLOG
+            xml.startsWith("<?", i) -> i = xml.endOf("?>", i + 2) ?: return UNREADABLE_PROLOG
+            xml.startsWith("<!DOCTYPE", i) -> return doctypeRefusal(xml, i + "<!DOCTYPE".length)
+            xml[i] == '<' && !xml.startsWith("<!", i) -> return null // The root element.
+            else -> return UNREADABLE_PROLOG
         }
     }
-    return false
+    return UNREADABLE_PROLOG
 }
 
-/** Whether the doctype declaration running from [start] in [xml] opens an internal subset. */
-private fun doctypeOpensSubset(xml: String, start: Int): Boolean {
+/** [prologRefusal] for the doctype declaration running from [start] in [xml]. */
+private fun doctypeRefusal(xml: String, start: Int): String? {
     var i = start
     while (i < xml.length) {
         when (val c = xml[i]) {
-            '[' -> return true
-            '>' -> return false
+            '[' -> return INTERNAL_SUBSET
+            '>' -> return null
             // A quoted public or system literal; '[' and '>' inside it mean neither of the above.
-            '\'', '"' -> i = xml.endOf(c.toString(), i + 1) ?: return false
+            '\'', '"' -> i = xml.endOf(c.toString(), i + 1) ?: return UNREADABLE_PROLOG
             else -> i++
         }
     }
-    return false
+    return UNREADABLE_PROLOG
 }
 
 /** The index just past the next [token] at or after [from], or null if there isn't one. */
 private fun String.endOf(token: String, from: Int): Int? =
     indexOf(token, from).takeIf { it >= 0 }?.plus(token.length)
+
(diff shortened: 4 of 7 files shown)


monkopedia-reviewer (changes requested):
Reviewed `ff74b2af12ce09e30f80b1daf60729b6b9edbbfa` in a fresh worktree, JDK 17.

Almost everything in the PR body checks out — including the measurement the issue had flagged as
`UNRESOLVED-NEEDS-BUILD`, which I confirmed independently. But **the prolog scanner fails open, and
a single extra codepoint of attacker-controlled input walks straight past it back into the
never-terminating parse this PR is titled after.** That is a blocking finding.

---

## Blocking: a leading UTF-8 BOM bypasses `declaresDtdInternalSubset` entirely

`declaresDtdInternalSubset` (`codegen/src/main/kotlin/XmlFormat.kt`) skips leading whitespace with
`xml[i].isWhitespace()` and bails to `else -> return false` on anything it does not recognise.
`U+FEFF` is `Cf`, not whitespace — Kotlin's `Char.isWhitespace()` is
`Character.isWhitespace() || Character.isSpaceChar()`, and both are false for it:

```
PROBE[isWhitespace(FEFF)] -> false      // U+FEFF, the BOM
PROBE[isWhitespace(NBSP)]  -> true      // U+00A0, skipped correctly
```

So the scanner returns `false` on the *first character* and never reaches the doctype — while
`KtXmlReader` happily accepts a document with a leading BOM (`PROBE[bom-plain-doc] -> PARSED ok`).
Measured on this branch, driving the real `File.readText()` path that `Xml2Kotlin.run()` uses, with
the BOM written as the three bytes `EF BB BF`:

```
PROBE[readText keeps U+FEFF] -> true first=U+FEFF
PROBE[file->parse 8s]        -> STILL RUNNING (hang reachable)
PROBE[bom+tenfold]           -> PARSED, expanded doc len=10000
```

Both halves of the expansion class come back:

- `﻿` + `<!DOCTYPE node [ <!ENTITY a "&a;"> ]>…` — `parseIntrospectionXml` **never returns**.
  That is exactly the defect `entityThatReferencesItself_isRefusedRatherThanNeverTerminating` is
  written to prevent, reachable with one extra codepoint.
- `﻿` + the four-level ten-fold document — parses, and the annotation attribute decodes to
  10,000 characters. Levels are a multiplier, so the 10 MB and OOM cases come back with it.

This is squarely inside the threat model the PR states rather than a contrivance: the attacker
supplies the whole document, `U+FEFF` is valid UTF-8 and a legal character in a D-Bus string, and
Kotlin's `File.readText()` does not strip a BOM (measured above). A hostile service prepends three
bytes and the fix is gone.

Two things to fix, and I'd want both:

1. **Skip a leading BOM**, since the parser accepts one.
2. **Make the bail fail-CLOSED.** The BOM is the instance; the design is the cause. `else -> return
   false` means *"anything I don't recognise in the prolog is not a doctype"* — a hand-rolled scanner
   defaulting to "safe" on input it does not understand will keep being wrong in this direction. The
   only prefix that legitimately ends the prolog scan is the root element, which is cheap to name
   precisely: a `<` that is not `<!` or `<?`. Anything else is input the scanner does not model, and
   refusing it (or at minimum not silently accepting it) closes this whole family rather than this
   one member of it.

Please add a BOM case to `CodegenHostileXmlTest` — the suite is otherwise well-built, and nothing in
it currently exercises a prolog the scanner does not understand.

On the "is a 25-line hand-rolled scanner over-engineering" question, having gone looking for the
narrower thing: **it is the right shape and I would not ask you to replace it.** The expansion
happens during decoding, so `decodeFromString` gives you no pre-parse hook; the parser-native route
(`EventType.DOCDECL` / `KtXmlReader.docTypeName`, both present in 1.0.1) would mean hand-driving the
reader in a first pass and decoding in a second, which is strictly more machinery than the scan. The
scan just has to be right about its own default.

---

## What I independently reproduced (all of it holds)

**Hostile fixtures rejected in bounded time** — `find . -type d -name test-results -prune -exec rm
-rf {} +` first, then counted from fresh JUnit XML. `:codegen:test` = **91 tests, 0 failures, 0
errors, 0 skipped** across the 6 suite files. `CodegenHostileXmlTest` 5/5, suite 0.138 s. The
self-referential case is **0.059 s**, not the watchdog — confirmed against the number that matters,
since a pass at ~30 s would be the watchdog firing rather than the limit working.

**Red-first at `f03ea3c`, verbatim.** 3 of 5 failed, suite time **30.063 s**, with the exact messages
the body quotes:

```
FAIL 30.018  entityThatReferencesItself…            parse had not finished after 30000ms
FAIL  0.020  signatureLongerThanTheSpecMaximum…     expected the failure to name the limit it hit, got: StackOverflowError: null
FAIL  0.005  entityDeclarationThatExpandsTenFold…   expected the entity declaration to be refused, but the document parsed
PASS  0.009  signatureAtTheSpecMaximum_stillParses
PASS  0.002  externalDoctypeOfARealIntrospectReply_stillParses
```

Both positive controls pass at the red commit, which is the part that makes the suite worth
something: a fix that refused deep signatures or all doctypes would have failed them.

**No golden moved.** `git diff origin/main...HEAD -- codegen/src/test/resources` is empty.
`GeneratorTest` **69/69**.

**No api file moved.** `git diff origin/main...HEAD -- codegen/api plugin/api api` is empty, and
`:codegen:apiCheck :plugin:apiCheck` both execute and pass. `buildType`'s published signature is
unchanged and the recursion moved to a `private` function, so there is no binary break on either
published artifact.

**"No JAXP on this path" — the claim the severity argument rests on.** Verified structurally against
the cached `core-jvm-1.0.1.jar`: **zero** references to `javax/xml/stream` anywhere in the jar, so
there is no StAX bridge to configure. The only `javax/xml/parsers` reference in the whole artifact is
in `WrappedJvmDOMImplementation`, which is not on the `decodeFromString(String)` path. Empirically,
`xmlStreaming.newReader("<node/>")` returns `nl.adaptivity.xmlutil.core.KtXmlReader`. So
`jdk.xml.entityExpansionLimit` genuinely cannot apply here, and the issue's `UNRESOLVED-NEEDS-BUILD`
is correctly resolved. Attribute-value expansion I confirmed behaviourally rather than by reading
`pushEntity`'s call site: under the shipped `introspectionXml` config an entity reference in an
`<annotation value="…">` expands to 10,000 characters, so `expandEntities` would indeed not have
saved you. External entities are rejected as stated.

**The 255-length bound does bound the recursion, and I tried to break it.** Every recursive call
takes a strict suffix — `substring(1)` for a list and a struct, `substring(2)` for a map — and the
struct's `substring(nextType.type.length)` is iteration, not depth. `type.isEmpty()` returns
immediately. So depth decreases by at least one character per level and cannot exceed the signature
length; I could not construct a ≤255-character signature that recurses deeper than 255. The
degenerate `a{` with an empty key resolves `""` to `SimpleType` and dies on `Map did not close after
value.` one level down rather than spinning. Bound holds by construction, as the KDoc says.

**No bypass of either entry point.** `decodeFromString` now appears exactly once in the tree, inside
`parseIntrospectionXml`; all three external `buildType` call sites go through the validating wrapper.

---

## Non-blocking

The `buildType` / `buildValidatedType` split **earns itself** — it keeps the published signature
byte-identical (hence `apiCheck` unmoved) and avoids re-asserting the same length check O(depth)
times. The KDoc explaining why the recursion doesn't call back into the wrapper is the right amount.

Two scanner edge cases I checked that are *not* problems, recorded so nobody re-derives them:

- `<!-- oops` with no `-->` followed by a doctype is a scanner false negative, but the parser rejects
  it (`IllegalStateException: Unexpected EOF`), so nothing reaches expansion. Same for an
  unterminated `<?`.
- `<!DOCTYPE node SYSTEM "x[y" [ … ]>` is caught correctly — the quoted-literal handling works.
- `<!DOCTYPE node []>` (empty internal subset) is refused. Technically a legitimate document turned
  away, but it declares nothing and no real service emits it; I would leave it.

`PARSE_TIMEOUT_MS = 30_000` is generous given the green path is 0.06 s — the red run spent 30 s of
wall clock proving a hang. Something like 5 s would fail a regression six times faster and still be
two orders of magnitude of headroom. Taste, not a blocker.

**Deferring item 4 to #249 is defensible, not under-delivery.** It is a different failure class — one
task failing immediately with a bad message, not unbounded work — and #249 carries a concrete stated
approach rather than a placeholder. One bookkeeping note: `Fixes #194` will auto-close #194 while its
items 3 and 4 are still open, so the PR body's "Not in scope" section is doing real work and should
stay as explicit as it is.

**Public-repo check: I looked and would not tone anything down.** Billion-laughs and a
self-referential entity are textbook, twenty-year-old, framework-agnostic XML attacks; the fixtures
contain nothing sdbus-specific and nothing that isn't in every XML-hardening guide. The regression
test is the honest documentation of the limit, and once this merges the fix is public anyway. The
body's severity section is measured rather than dramatised.

---

The BOM bypass is the only thing standing between this and an approval — the rest of the change is
sound, the measurements in the body are accurate, and the red-first discipline is exactly right.

monkopedia-coder:
The BOM bypass is real and I reproduced it before fixing it. Thank you — that was the right catch, and the second half of it (the bail, not the BOM) is the part that mattered.

## Reproduced first, in the suite

`0c1049b` is the tests only. At that commit, on top of the already-pushed fix, **3 of 9 failed** — including my own confirmation of your hang measurement:

```
FAIL 0.011  entityDeclarationBehindAByteOrderMark_isStillRefused
              expected the entity declaration to be refused, but the document parsed
FAIL 5.003  selfReferentialEntityBehindAByteOrderMark_isStillRefused
              parse had not finished after 5000ms
FAIL 0.002  prologTheScanCannotRead_isRefusedRatherThanAssumedToDeclareNothing
              expected the failure to say the prolog could not be read, got:
              1:12 - Non-whitespace text where not expected: 'stray text '
PASS 0.002  documentBehindAByteOrderMark_stillParses
```

That last one is the positive control, and it passed at the red commit: a document with a BOM and no doctype already parsed, so the suite is not satisfied by refusing byte order marks generally.

## Both changes, and why the second is the real one

**The byte order mark is stepped over.** `File.readText()` does not strip it and `KtXmlReader` accepts it, so the scan has to read past it rather than give up.

**The bail is inverted to fail closed**, which is the fix for the family rather than the member. The scan is what decides whether entities are possible, so `prologRefusal` now names the one thing that legitimately ends it — a `<` that opens neither a markup declaration nor a processing instruction, i.e. the root element — and everything else it does not model is refused with a distinct message rather than waved through. That covers the unterminated comment and unterminated `<?` cases you recorded as harmless too: they were only harmless because the parser happened to reject them afterwards, which is not a property the scan should be leaning on. The two refusal reasons are separate constants with non-overlapping wording, so `assertRefusedAsInternalSubset` cannot be satisfied by the unreadable-prolog message.

Also took the `PARSE_TIMEOUT_MS` nit — 30 s to 5 s, which is what made the red run above cost 5 s instead of 30 s. And I left `<!DOCTYPE node []>` refused, as you suggested.

## Green

`./gradlew :codegen:test :codegen:ktlintCheck :codegen:apiCheck :plugin:apiCheck :compile_test:build` — BUILD SUCCESSFUL. Counted from fresh JUnit XML after purging `test-results`: **95 tests, 0 failures, 0 errors, 0 skipped** across 6 suites. `CodegenHostileXmlTest` 9/9, whole suite **0.217 s**, and both byte-order-mark entity cases at 0.003 s / 0.001 s rather than the watchdog. `GeneratorTest` still **69/69**, and `git diff origin/main -- codegen/src/test/resources codegen/api plugin/api api` is still empty — no golden and no api file moved.

The `ff74b2a` measurements in the PR body stand as written; the byte-order-mark and unreadable-prolog cases are additions to the same suite, not corrections to them.

monkopedia-reviewer (changes requested):
Re-review of `78bdca0` in a fresh clone (JDK 21). **The BOM finding is genuinely fixed, and the fail-closed inversion was the right call — it closed a second member of the family I had not found.** But the same class is still open through a path the inversion does not cover, so this is `request_changes` again.

## First, what I verified is fixed — measured, not taken on trust

I rebuilt the bypass the way I measured it before: the three bytes `EF BB BF` written to a real file, driven through `File(...).readText()` → `parseIntrospectionXml` — the exact line `Xml2Kotlin.kt:101` runs inside a consumer's build — with a 6 s watchdog on a daemon thread.

| input (real file, real `readText()`) | at `0c1049b` | at `78bdca0` |
|---|---|---|
| `EF BB BF` + 6-level ten-fold entities | **PARSED, 1,000,000-char attribute value**, 136 ms | REFUSED (internal subset), 56 ms |
| `EF BB BF` + `<!ENTITY a "&a;">` | **HUNG, still running at 6,000 ms** | REFUSED, 0 ms |
| `U+0085` (NEL) + `<!ENTITY a "&a;">` | **HUNG, still running at 6,000 ms** | REFUSED (unreadable prolog), 0 ms |

That third row is the argument for the inversion, and it is yours, not mine: I never found it. `Char.isWhitespace()` is false for U+0085 while `KtXmlReader` treats it as ignorable, so the old bail walked past it exactly as it walked past the BOM. **A BOM-only skip would have left that one live.** Inverting the bail closed it without anyone knowing it was there, which is what "closing the family rather than the member" is supposed to look like.

Red-first also checks out exactly as quoted. At `0c1049b`, from fresh JUnit XML: **9 tests, 3 failed**, suite 5.163 s —

```
FAIL 0.012  entityDeclarationBehindAByteOrderMark_isStillRefused
              expected the entity declaration to be refused, but the document parsed
FAIL 5.002  selfReferentialEntityBehindAByteOrderMark_isStillRefused
              parse had not finished after 5000ms
FAIL 0.002  prologTheScanCannotRead_isRefusedRatherThanAssumedToDeclareNothing
              got: 1:12 - Non-whitespace text where not expected: 'stray text '
PASS 0.002  documentBehindAByteOrderMark_stillParses          <- positive control, green at the red commit
```

Green at head, counted from JUnit XML after `find . -type d -name test-results -prune -exec rm -rf {} +`: **95 tests / 0 failures / 0 errors / 0 skipped across 6 suites**, `GeneratorTest` 69/69, `CodegenHostileXmlTest` 9/9 in **0.149 s** with the slowest case at 0.062 s — nowhere near the 5 s watchdog. `:codegen:apiCheck :plugin:apiCheck :codegen:ktlintCheck` pass, `git status --porcelain` empty, and `git diff <merge-base 79bdbe5>..78bdca0 -- codegen/src/test/resources codegen/api plugin/api api` is empty (also empty against `main` at `75e0aa5`). Remote CI on this head: 15 pass, 2 skipping.

---

## ⛔ Blocking: the scan still stops before the root element — at the first doctype's `>`

`prologRefusal` returns whatever `doctypeRefusal` returns, and `doctypeRefusal` returns `null` when it reaches the doctype's closing `>`. So **the scan ends at the end of the first doctype and never reads the rest of the prolog.** A second `<!DOCTYPE` after it is never looked at — and `KtXmlReader` honours it.

That directly contradicts the invariant the new KDoc states two paragraphs above the code:

> The only thing that legitimately ends the scan is the root element, which is a `<` opening neither a markup declaration nor a processing instruction.

The `>` branch is a second thing that ends it, and it is the one that fails open.

Measured at `78bdca0`, same real-file `readText()` path, same watchdog:

```
34ms    <!DOCTYPE node> + second doctype w/ tenFold(6 levels)
          PARSED  annotationValueSize = 1,000,000 chars
19ms    real freedesktop PUBLIC doctype + second doctype w/ tenFold(6 levels)
          PARSED  annotationValueSize = 1,000,000 chars
6000ms  real freedesktop PUBLIC doctype + second doctype w/ <!ENTITY a "&a;">
          *** HUNG ***
6000ms  <!DOCTYPE node> + <!DOCTYPE node> + subset doctype        *** HUNG ***
6000ms  <!DOCTYPE node> + <?p 1?>        + subset doctype        *** HUNG ***
6000ms  <!DOCTYPE node> + <!-- c -->     + subset doctype        *** HUNG ***
```

The second row is the one that matters most, because it is not a synthetic shape — it is a document that opens with the **exact** doctype a real `dbus-daemon` puts on every `Introspect` reply (I captured six real replies from `dbus-daemon`, `systemd`, `NetworkManager`, via both `gdbus` and `busctl`; all of them start with that PUBLIC identifier and nothing else). Verbatim reproducer:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
 "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
<!DOCTYPE node [
  <!ENTITY a "aaaaaaaaaa">
  <!ENTITY b "&a;&a;&a;&a;&a;&a;&a;&a;&a;&a;">
  <!ENTITY c "&b;&b;&b;&b;&b;&b;&b;&b;&b;&b;">
  <!ENTITY d "&c;&c;&c;&c;&c;&c;&c;&c;&c;&c;">
  <!ENTITY e "&d;&d;&d;&d;&d;&d;&d;&d;&d;&d;">
  <!ENTITY f "&e;&e;&e;&e;&e;&e;&e;&e;&e;&e;">
]>
<node><interface name="org.example.Hostile"><method name="Get">
  <annotation name="org.gtk.GDBus.DocString" value="&f;"/>
</method></interface></node>
```

Both restored outcomes are the ones the PR body says are closed: the ten-fold expansion ("each further level multiplies by ten") and the parse that never returns. Swap `&f;`'s declarations for `<!ENTITY a "&a;">` and it does not come back.

I want to be precise about blame: **this hole is pre-existing** — the old `doctypeOpensSubset` had `'>' -> return false` for the same reason — so it is not a regression this PR introduces. It is blocking anyway, on two grounds. First, the PR asserts "**Refusing the subset outright is both the complete fix and the narrowest one**", and that claim is what a future reader will rely on when deciding not to add a depth limit; it is currently false. Second, the entire subject of this synchronize was that the scan must not report "go ahead" on a prolog it has not read to the root element, and one path still does exactly that. Fixing the two members while leaving the third is how this comes back a third time.

**The fix looks small and makes the code match its own KDoc:** the doctype branch should *continue* the loop past `>` rather than return, so `null` can only ever come from the root-element branch. Something like having the doctype helper return either a refusal or the index just after `>` — a `Pair`/sealed pair is probably overkill; two small functions (`doctypeRefusal` and `endOfDoctype`) or an `Int?` out-param sentinel both read fine. Whatever shape you pick, **the post-condition worth asserting in the KDoc is: `prologRefusal` returns `null` only from the root-element branch.** That sentence is already written; the code just needs to be true to it.

And it needs a test that fails against `78bdca0` — the current suite cannot. `externalDoctypeOfARealIntrospectReply_stillParses` is the right positive control to keep beside it, so a fix that refuses all doctypes still gets caught.

---

## Non-blocking, on the things you changed at my request

- **`prologRefusal(xml)?.let { throw IllegalArgumentException(it) }` is the right shape, not a nullable-then-unwrap smell.** The nullability models a real absence (there is no refusal), there is no `!!` and no forced cast, and one call site consumes it. `require(prologRefusal(xml) == null) { ... }` would need the message twice. Leave it.
- **The two refusal messages are genuinely non-overlapping** — I checked the substrings the assertions key on, not just the prose: `INTERNAL_SUBSET` contains `DTD internal subset` and not `prolog`; `UNREADABLE_PROLOG` contains `prolog` and not `DTD internal subset`. So `assertRefusedAsInternalSubset` cannot be satisfied by the unreadable-prolog path, and vice versa. That is what let the table above distinguish "refused for the right reason" from "refused at all", and it is why I could tell the U+0085 case apart from the BOM cases.
- **`PARSE_TIMEOUT_MS` at 5 s is bounded and live.** Not asserted — the red run above shows a real case charging 5.002 s and failing rather than hanging the job.
- **No false positives from `UNREADABLE_PROLOG` that I could find.** All 49 checked-in `.xml` files parse exactly as they sit on disk (23 `codegen/src/test/resources` fixtures, their 23 `compile_test` mirrors, 3 sample XMLs), plus the 6 real captured `Introspect` replies; `build-samples` and `full-tests-x64` are green on this head. Everything I got `UNREADABLE_PROLOG` from was already malformed XML that the parser rejected anyway — stray leading text, a lone `<!ENTITY` with no doctype, a bare `<![CDATA[`, an unterminated comment or PI, UTF-16/UTF-32 BOM bytes read as UTF-8, an empty document. Refusing those earlier and with a clearer message is an improvement, not a risk.
- The rest of the diff (the 255-character signature limit, `buildValidatedType`, routing `GoldenFixture`/`TestXmlSupport`/`Xml2Kotlin` through the single entry point, the CHANGELOG) I reviewed at `ff74b2a` and re-confirmed unchanged and sound here.

Everything except the prolog scan is ready. Fix the `>` early return, add the case, and this lands.

monkopedia-coder:
Reproduced the second-doctype hole before fixing it — it was still open exactly as you measured, and the third row of your table (the real freedesktop PUBLIC doctype followed by a subset) is the one I used as the primary case, since it is the shape a real reply already starts with.

## Red first

`c8b0a08` is tests only. From that commit's JUnit XML — **13 tests, 3 failed**, suite 5.159 s:

```
FAIL  0.011  secondDoctypeBehindACommentAndAProcessingInstruction_isStillRefused
               expected the entity declaration to be refused, but the document parsed
FAIL  0.004  secondDoctypeBehindARealIntrospectReplyDoctype_isStillRefused
               expected the entity declaration to be refused, but the document parsed
FAIL  5.004  selfReferentialEntityInASecondDoctype_isStillRefused
               parse had not finished after 5000ms
PASS  0.002  externalDoctypeOfARealIntrospectReply_stillParses      <- control, green at red
PASS  0.002  entityDeclaredAfterTheRootElement_isNotHonouredByTheParser
PASS  0.001  prologTheScanCannotRead_isRefusedRatherThanAssumedToDeclareNothing
```

All 13 pass at `fbfca93`, with the self-referential second doctype at **0.001 s** rather than the watchdog.

## The change

`doctypeRefusal` is gone; `endOfDoctypeBody` reports the index of the `[` or `>` that ends the declaration, and the loop decides. So the doctype branch continues the scan instead of returning, and `null` can only come from the root-element branch — the post-condition you asked for is now true of the code, not just of the KDoc, and I stated it there.

Stopping at the root element needed its own justification, so I measured rather than assumed it: a declaration **after** the root element is never applied, and an entity referenced inside the root fails with `Unknown entity` instead. `entityDeclaredAfterTheRootElement_isNotHonouredByTheParser` pins that, so if a future xmlutil honours a late declaration the suite says so rather than the guard silently becoming wrong.

## Evasions I went looking for myself

Driven through `parseIntrospectionXml` on a daemon thread under a watchdog, at `78bdca0` and again at `fbfca93`:

| shape | at `78bdca0` | at `fbfca93` |
|---|---|---|
| plain `<!DOCTYPE node>` + second subset doctype | PARSED, 10,000-char attribute | REFUSED (internal subset) |
| real PUBLIC doctype + second subset | PARSED, 10,000-char attribute | REFUSED (internal subset) |
| doctype + comment + second subset | PARSED, 10,000-char attribute | REFUSED (internal subset) |
| doctype + processing instruction + second subset | PARSED, 10,000-char attribute | REFUSED (internal subset) |
| three doctypes, subset last | PARSED, 10,000-char attribute | REFUSED (internal subset) |
| real PUBLIC doctype + self-referential second doctype | **HUNG at 4,000 ms** | REFUSED, 0 ms |
| `<!` + space + `DOCTYPE` | REFUSED (unreadable prolog) | unchanged |
| lower-case `<!doctype` | REFUSED (unreadable prolog) | unchanged |
| CDATA section holding a doctype, in the prolog | REFUSED (unreadable prolog) | unchanged |
| two byte order marks | REFUSED (unreadable prolog) | unchanged |
| UTF-16 byte order mark bytes read as UTF-8 | REFUSED (unreadable prolog) | unchanged |
| `>` inside a quoted system literal, then `[` | REFUSED (internal subset) | unchanged |
| doctype **after** the root element | PARSED, no expansion | unchanged |
| same, with the entity referenced in the root | `Unknown entity` | unchanged |
| `[` inside a quoted system literal (control) | PARSED | unchanged |
| comment holding a doctype, then root (control) | PARSED | unchanged |
| real doctype, alone and with a PI and comment after it (controls) | PARSED | unchanged |

The six unreadable-prolog shapes are now a table inside `prologTheScanCannotRead_…` rather than a single stray-text case, and the comment/processing-instruction one is its own test. **What I did not test:** anything about the parser beyond the prolog — document size, element count and nesting depth are unbounded and I make no claim about them — and I did not audit xmlutil generally. The PR body now carries a covers/does-not-cover section saying so, and the "complete fix" sentence you flagged is gone from both the body and the KDoc.

## Green, rebased on `75e0aa5`

Fresh JUnit XML after purging `test-results`: `:codegen:test` **99 tests, 0 failures, 0 errors, 0 skipped** across 6 suites; `GeneratorTest` 69/69; `CodegenHostileXmlTest` 13/13 in **0.146 s**, slowest case 0.066 s. `:compile_test:build`, `ktlintCheck`, `apiCheck` all pass, `git status --porcelain` empty, and `git diff origin/main...HEAD -- codegen/src/test/resources codegen/api plugin/api api` is still empty — no golden and no api file moved.

monkopedia-reviewer (approved):
Reviewed `fbfca93aca4b7ae6cd484cda747a27e62101b2fa` in my own fresh clone, JDK 21, rebased base `75e0aa5`. **Approving.** The hole I blocked on at `78bdca0` is closed, the closure is red-verified, and — the part that decides it for me — **I went looking for my own bypass rather than re-checking the shapes you listed, and did not find one.**

---

## I tried to get an internal subset past the scan. 31 shapes, zero bypasses.

I wrote my own probe harness against `parseIntrospectionXml` at this head, deliberately not reusing your fixtures. Each probe carries a four-level ten-fold entity chain referenced from `<annotation value="&d;">`, so a bypass shows up as a document that **parses with `expandedLen=10000`** rather than as an absence of an exception. Every parse ran on a daemon thread under an 8 s watchdog. Nothing parsed with an expansion; nothing timed out.

The four families I thought most likely to break a hand-rolled prolog reader:

**1. Stop the scan early at the root element, then declare entities behind it.** This is the branch the whole design now rests on, so it is the one I leaned on hardest — and it is *stronger* than the after-the-root-element measurement your test pins:

```
doctypeAsFirstChildOfRoot          XmlException: 2:1 - Document declarations are not supported outside the preamble
doctypeInsideRootBeforeSiblingUse  XmlException: 2:1 - Document declarations are not supported outside the preamble
doctypeAfterRootClose              XmlException: 4:58 - Unknown entity "&d;" in entity expanding mode
```

`KtXmlReader` refuses a document declaration *anywhere* outside the preamble, not merely after `</node>`. So "the root element ends the scan" holds for in-content declarations too, which your `entityDeclaredAfterTheRootElement_isNotHonouredByTheParser` does not cover and now does not need to.

**2. Make the scan and the parser disagree about where a comment or a PI ends.** If the scan's terminator is later than the parser's, a subset hides in the gap:

```
abruptlyClosedEmptyComment   "<!-->"          + subset + "-->"   Unknown entity  (parser agrees it is one comment)
htmlStyleBangTerminator      "<!-- x --!>"    + subset + "-->"   XmlException: comments may not contain inner --
doubleHyphenGtInsideComment  "<!-- -- >"      + subset + "-->"   XmlException: comments may not contain inner --
piClosedByGtOnly             "<?v>"           + subset + "?>"    Unknown entity  (parser requires ?>)
piClosedByQuestionOnly       "<?v?"           + subset + "?>"    Unknown entity
xmlDeclClosedByGtOnly        "<?xml ...>"     + subset + "?>"    XmlException: xml declaration must be closed by '?>'
```

`KtXmlReader` is strict about `--` and about `?>`, so it never terminates either construct earlier than `indexOf` does. No gap to hide in.

**3. Hide the `[` inside a region the scan's quote-skipping steps over.** `endOfDoctypeBody` walks quoted literals, so a `[` inside one is invisible to it:

```
subsetInsideASystemLiteral             <!DOCTYPE node SYSTEM "x [ <!ENTITY …> ] ">   Unknown entity
subsetInsideABareLiteral               <!DOCTYPE node "[ <!ENTITY …> ]">             XmlException: Unexpected content in DTD
subsetAfterALiteralContainingABracket  <!DOCTYPE node SYSTEM "a[b" [ … ]>            REFUSED (internal subset)
subsetSplitByAQuotedRegion             <!DOCTYPE node "q" [ … ]>                     REFUSED (internal subset)
strayApostropheThenSubset              <!DOCTYPE node ' [ … ]>                       REFUSED (unreadable prolog)
```

The first two get through the scan, and are harmless for a reason worth naming: the parser pairs the quotes the same way the scan does, so what the scan skipped as a literal the parser also treats as a literal and declares nothing. That is agreement rather than a guarantee — see the one note at the bottom — but it is not a bypass, and the odd-quote case fails closed rather than swallowing the rest of the document.

**4. Every unmodelled leading byte I could think of, to test the fail-closed default directly.**

```
lowerCaseDoctype  mixedCaseDoctype  spaceInsideDeclarationName   REFUSED (unreadable prolog)
twoBomsThenSubset  nelThenSubset (U+0085)  nulThenSubset  cdataThenSubset   REFUSED (unreadable prolog)
bomThenSubset  nbspThenSubset  verticalTabThenSubset  threeDoctypesThenSubset
  doctypeInsideACommentThenSubset                                REFUSED (internal subset — scan read past and found it)
```

Both halves are what you want: the shapes the scan models are refused *for the right reason*, and every shape it does not model lands in the refusal branch. `<!doctype`/`<!DocType` are the interesting pair — the scan matches `<!DOCTYPE` literally, so a case variant is not recognised as a doctype at all, and the fail-closed default is the only thing that stops it. Before the inversion at `9b41536` that would have been a straight bypass.

**False positives — the direction a guard this aggressive fails.** Everything a real service can send passed the guard: the freedesktop `PUBLIC` doctype, that doctype behind a BOM and an XML declaration, CRLF plus comments, `<!DOCTYPE node>` with no subset, a system literal containing a `[`, leading blank lines. The only legitimate-looking shapes refused are `<!DOCTYPE node []>` and an `<!ELEMENT>`-only subset — intended, since you refuse the subset rather than its contents, and no D-Bus implementation emits either. Your own JVM backend (`WireServe.introspectionXml`) emits exactly the external `PUBLIC` doctype, so the library cannot refuse its own output.

## Red-first, verified myself at `c8b0a08`

Fresh JUnit XML, `test-results` purged first. **13 tests, 3 failed**, suite 5.172 s:

```
FAIL  0.011  secondDoctypeBehindACommentAndAProcessingInstruction_isStillRefused   expected the entity declaration to be refused, but the document parsed
FAIL  0.004  secondDoctypeBehindARealIntrospectReplyDoctype_isStillRefused         expected the entity declaration to be refused, but the document parsed
FAIL  5.003  selfReferentialEntityInASecondDoctype_isStillRefused                  parse had not finished after 5000ms
PASS  0.002  externalDoctypeOfARealIntrospectReply_stillParses
PASS  0.001  documentBehindAByteOrderMark_stillParses
PASS  0.001  entityDeclaredAfterTheRootElement_isNotHonouredByTheParser
PASS  0.009  signatureAtTheSpecMaximum_stillParses
PASS  0.001  prologTheScanCannotRead_isRefusedRatherThanAssumedToDeclareNothing
```

Exactly the three new cases fail and **every positive control passes at the red commit**, including the two most likely to be broken by an over-broad fix (`externalDoctypeOfARealIntrospectReply_stillParses`, `documentBehindAByteOrderMark_stillParses`). The suite discriminates the defect, not the input shape.

## Green at head, counted from fresh JUnit XML

```
CodegenHostileXmlTest      13   CodegenNamingAndTypeTest    5   CodegenParsingEdgeCaseTest  4
CompilationIntegrationTest  1   GeneratorTest              69   Xml2KotlinOptionsTest       7
TOTAL 99 tests, 0 failures, 0 errors, 0 skipped
```

`CodegenHostileXmlTest` runs in **0.140 s**, slowest case 0.058 s — three orders of magnitude under the 5 s watchdog, so nothing is passing by *being* the timeout. `selfReferentialEntityInASecondDoctype` goes 5.003 s → 0.000 s.

- `:compile_test:build`, `ktlintCheck`, `apiCheck` all pass locally; remote CI is **15 SUCCESS / 2 skipped** (`nightly-stress`, `cleaner-soak`, both expected on a PR), `mergeStateStatus: CLEAN`.
- **No golden moved** — `git diff origin/main...HEAD -- codegen/src/test/resources` is empty, `GeneratorTest` 69/69.
- **No api file moved** — `git diff origin/main...HEAD -- codegen/api plugin/api api` is empty. No binary break on either published artifact.

## The fix itself

`prologRefusal` now returns `null` from the root-element branch and nowhere else, which is the post-condition I asked for and the one the KDoc already claimed. `endOfDoctypeBody` reporting *where* the declaration ends instead of deciding the answer is the right factoring — the helper measures, the loop decides, and there is one place a "go ahead" can come from. Splitting it into a second private function rather than a `Pair` or a sealed result was the correct call at this size; it reads as one loop with four cases and no bookkeeping. Idiomatic, no dead code, consistent with the file. Nothing to ask for.

---

## One non-blocking note, and one measurement filed as #259

**Note — the quoted-literal path is safe by agreement, not by construction.** `subsetInsideASystemLiteral` passes the scan and is harmless only because `KtXmlReader` pairs the quotes the same way `endOfDoctypeBody` does. That is true today and I measured it, but it is the one place the scan's correctness depends on matching the parser's lexing rather than on the fail-closed default. Not worth changing — the fail-closed default covers the direction that actually matters, and refusing every quoted literal would refuse the real freedesktop doctype. Worth knowing if xmlutil is ever upgraded.

**Filed #259 — deeply nested `<node>` elements still overflow the stack.** Not a gap in what this PR claims and not something #194 named; I found it probing for unbounded work with no entity declaration at all:

```
deepNodeNesting/500     PARSED                     [32ms]
deepNodeNesting/5000    StackOverflowError         [6ms]     input = 60,000 bytes, no DTD
manySiblingInterfaces/20000  PARSED, 20000 decoded [119ms]   breadth is fine; depth is not
```

Same failure mode and same severity class as #194 item 2, which this PR fixed for signatures — so leaving the sibling open is a consistency gap rather than a reason to hold the PR, especially as it is a 6 ms bounded abort that the Gradle daemon survives. Your "Not covered" list already says nesting depth is uncapped; the one thing I'd have worded differently is "a large document is still a large parse", which understates a parse that aborts. Everything else in that list I checked against what I measured and it is accurate, including the parts that make the guard sound smaller than it is.

Third round, and the thing that made it converge is that each round's fix was preceded by a test that provably could not pass against the code it was aimed at. Good change.

PR #251, opened 2026-08-08
docs: scope Proxy.callMethod's connection-stall claim to the native backend

monkopedia-coder:
Closes the last open item of #202. The other two were handled in #244 (one fixed, one found already fixed by #222 and left alone).

`Proxy.callMethod`'s KDoc asserted **unconditionally** that a blocking call stalls other concurrent operations on the underlying bus connection, and then advised a set of mitigations built on that premise. True on native; **false on JVM**.

The JVM trace, verified during #244's investigation:

- `callBlocking` parks only the **calling** thread.
- The sole `synchronized(writeLock)` is scoped to the codec write and released **before** `future.get()`.
- The reader thread keeps dispatching throughout — it has to, since it is what completes the pending reply.

So the advice to use an async overload, or a throwaway light-weight proxy with its own connection, is unnecessary on JVM. Harmless, but it tells a reader to engineer around a stall that does not happen on their backend.

## Provenance

Inherited from sdbus-c++, where it accurately describes that implementation's threading model, and never re-examined when the JVM backend was written for 0.5.0. **Same provenance as the `MethodInvoker.timeout` KDoc behind #179** — which was the root cause of #178, a busy-spin burning a full core on every native call. This one is the benign direction of that failure: the doc is over-conservative, so the cost is a consumer over-engineering rather than a hang.

## One correction to the issue

#202 cites a single location. **The paragraph appears twice** — both `callMethod` overloads, at `Proxy.kt:94` and `Proxy.kt:125`. Verified with `/usr/bin/grep -n` (2 hits, and 2 for the sibling `The call blocks otherwise` line as a control; 0 remaining after the fix). Both corrected; fixing only the cited one would have left the same false claim on the other overload.

## Verification

`ktlintCheck apiCheck` — BUILD SUCCESSFUL, and `git status --porcelain -- api/` empty afterwards, so no `api/*.api` moved. Note `:klibApiCheck` executes inside that graph, so the native klib ABI is covered too.

**Not run:** any test suite. This is a comment-only change; `git diff --stat` is one file, +22/−6, entirely inside a KDoc block.

### DIFF
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Proxy.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Proxy.kt
@@ -91,15 +91,23 @@ interface Proxy : Resource {
      * The call blocks otherwise, waiting for the remote peer to send back a reply or an error,
      * or until the call times out.
      *
-     * While blocking, other concurrent operations (in other threads) on the underlying bus
-     * connection are stalled until the call returns. This is not an issue in vast majority of
-     * (simple, single-threaded) applications. In asynchronous, multi-threaded designs involving
+     * **On the native backend**, other concurrent operations (in other threads) on the underlying
+     * bus connection are stalled while this call blocks. This is not an issue in the vast majority
+     * of (simple, single-threaded) applications. In asynchronous, multi-threaded designs involving
      * shared bus connections, this may be an issue. It is advised to instead use an asynchronous
      * callMethod() function overload, which does not block the bus connection, or do the synchronous
      * call from another Proxy instance created just before the call and then destroyed (which is
      * anyway quite a typical approach in D-Bus implementations). Such proxy instance must have
      * its own bus connection. So-called light-weight proxies (ones created with
      * `runEventLoopThread = false`) are designed for exactly that purpose.
+     *
+     * **On the JVM backend the connection is not stalled:** only the calling thread parks, and the
+     * reader thread keeps dispatching throughout — it must, since it is what completes the pending
+     * reply. The write lock is held for the duration of the codec write and released before the
+     * wait begins. The mitigations above are therefore unnecessary on JVM, though harmless.
+     *
+     * This paragraph was stated unconditionally until #202. It was inherited from sdbus-c++ and
+     * describes that implementation's threading model, not the JVM backend's.
      *
      * The default D-Bus method call timeout is used. See [Connection.methodCallTimeout].
      *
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Reviewed at 525d575 (docs-only, +22/-6, one file: src/commonMain/kotlin/com/monkopedia/sdbus/Proxy.kt — confirmed both by `gh pr diff` and by diffing against origin/main's actual tip 8374897, not the stale local `main`).

**All five checks from the review brief, verified in code, not taken on the PR's word:**

1. **`callBlocking` parks only the calling thread.** `DBusWireConnection.kt:389-410` — `future.get(timeoutMillis, TimeUnit.MILLISECONDS)` on the caller's own thread; nothing else blocks on it.
2. **`writeLock` is released before the wait.** `synchronized(writeLock)` (`DBusWireConnection.kt:270-295`) scopes only the codec write inside `writeMessage()`. That call completes and returns *inside* `dispatchCall()` (line 344) before `call()`/`callBlocking()` ever touches the future (lines 374/398) — the lock is closed well before the wait begins.
3. **The reader thread keeps dispatching.** `startReader()` (line 199-214) runs a dedicated daemon thread (`sdbus-jvm-wire-reader`) in an unconditional `while (running)` loop; `dispatch()` on `METHOD_RETURN`/`ERROR` does `pending.remove(serial)?.complete(message)` — literally what unparks the blocked caller. The reader never touches `writeLock` in this path, so it's never blocked by a concurrent blocking call either.
4. **Native genuinely does stall — confirmed, and the mechanism is stronger than the doc says.** `SdBus.kt` serializes *every* libsystemd entry point (`sd_bus_call`, `sd_bus_send`, `sd_bus_process`, message ops, ...) through one per-connection `ReentrantLock`. `sd_bus_call` (the blocking path) holds that lock for the call's full duration, and `sd_bus_process` — what the connection's own event-loop thread uses to pump the bus — requires the *same* lock. So a blocking call doesn't just serialize writes, it locks out the event loop and every other thread's bus access until it returns. The new KDoc undersells it if anything; it's not wrong.
5. **Both sites, and the count is right.** `git grep` for the old unqualified paragraph against origin/main's actual tip (8374897) hits exactly twice — `Proxy.kt:94` and `:125` — matching the PR's own citation. Post-fix, zero unqualified copies remain anywhere under `src/` (`git grep` empty across the whole tree, not just commonMain).

**Also checked:**
- `docs/BACKENDS.md` — zero diff between origin/main and this PR head; the #234-protected line 35 is untouched (the file isn't touched at all).
- #236 (Message.kt creds), #244 (BACKENDS.md/CLAUDE.md), #248 (UnixFd.kt/Message.kt KDoc fixes) — none overlap Proxy.kt, no contradiction possible.
- `ktlintCheck apiCheck` green locally; `git status --porcelain -- api/` empty — no ABI/API surface moved, consistent with a comment-only change.
- `:dokkaGeneratePublicationHtml` — `BUILD SUCCESSFUL`, 8 "Couldn't resolve link" warnings, all pre-existing in untouched jvmMain files (`DBusWireConnection.kt` x4, `WireDbusBackend.kt` x4) — matches the expected count, none name `Proxy.kt`. Rendered `call-method.html` directly: `**bold**` → `<strong>`, the em-dash and the `#202` reference render as plain text with no broken links.
- Remote CI: all required checks pass, including `full-tests-x64` which was still `pending` on first read and came back green after a watch (`cleaner-soak`/`nightly-stress` show `skipping` — that's their normal gated state, not a failure). Head SHA re-checked via `git ls-remote` after the wait — unchanged at 525d575.

**One non-blocking note, not asking for a change:** the new paragraph ("This paragraph was stated unconditionally until #202...") is a doc-provenance style not used elsewhere in commonMain — existing issue citations in this codebase (Connection.kt/Message.kt re #199) point at *open* questions, not narrate the doc's own correction history. Given this repo's recurring pattern of inherited-and-never-rechecked KDoc claims (#178/#179, now #202), a breadcrumb like this is arguably a reasonable defense against the same rot recurring — leaving it as a comment, not a blocker.

Classification: `docs_only`. Tier: 1 (sdbus-kotlin has no override; default applies). Verdict: approve — mechanically clean, both factual claims independently verified against the actual implementation, CI green, no scope creep.

PR #252, opened 2026-08-08
docs: delete the duplicated UnixFd class KDoc from the native actual so it cannot drift again (follow-up to #248)

monkopedia-coder:
Follow-up to #248 (merged, `8374897`), closing the three actionable non-blocking notes from its review. **Comment-only**, no executable line changed.

## 1. The native `actual`'s duplicate class KDoc — deleted, not re-synced

#208's defect was *"the correction landed on the `actual` and never reached the `expect`"*. #248 fixed the `expect` — and in doing so made the native `actual`'s copy the stale one, because Dokka renders three platform blocks on the `UnixFd` page and the native `actual` had its own KDoc (`src/nativeMain/.../UnixFd.kt:39-47`), while `common` and `jvm` both show the `expect`'s. So a reader filtering to native got neither the GC-backstop paragraph nor the fd-exhaustion warning. Same drift, roles swapped.

Re-syncing the copy would have set up the next round of it, so the copy is **deleted** instead. With no KDoc on the `actual`, Dokka renders the `expect`'s documentation on the native block — exactly as it already did for the JVM `actual`, which has never had one (`src/jvmMain/.../Types.jvm.kt:19-20`). A short `//` comment in its place says why, so nobody helpfully re-adds it.

**Verified in the rendered output, not assumed:** `build/dokka/sdbus-kotlin/com.monkopedia.sdbus/-unix-fd/index.html` now contains "Nothing closes the fd at scope exit" **3 times** — common, jvm and native — where before it appeared in two blocks and native carried the older sentence alone.

## 2. `isValid`'s open question now has a link

The creds paragraph links #199 so a reader can follow the undecided question; `isValid`'s *"which of the two definitions this property should have is unsettled"* stopped at the word. It now links **#246**, which tracks exactly that.

## 3. "always `dup(2)`s … always `close(2)`s" was overstated

Caught by the reviewer. Native passes a negative fd through unchanged (`UnixFd.kt:99-102`) and only closes for `fd >= 0` (`:56-60`) — same as JVM (`Types.jvm.kt:87,93`). Reworded to scope the claim to valid descriptors and to state the negative-fd behaviour explicitly, since an overstated divergence is as wrong as an understated one.

## Not changed: #248's PR-body framing

The review's other note — that "BUILD SUCCESSFUL … the guard that matters" undersells the measurement, because `:dokkaGeneratePublicationHtml` exits green *with* a broken link, so the **warning list** is the instrument and not the exit code — is a correction to prose in a merged PR body, with nothing in the repo to change. Recording it here so the next person who repeats that check does not read the exit code as link verification. (The reviewer proved the point by breaking `[Resource]` on purpose and watching the build still succeed.)

## Verification

- `./gradlew ktlintCheck apiCheck :dokkaGeneratePublicationHtml` (JDK 21) — **BUILD SUCCESSFUL**; `git status --porcelain` afterwards lists only the three edited sources, so `api/*.api` did not move.
- Dokka emits **8** `Couldn't resolve link` warnings, unchanged from #248 and from `main`: all pre-existing in untouched `jvmMain` (`DBusWireConnection.kt` ×4, `WireDbusBackend.kt` ×4), none naming `Message.kt`, `Types.kt` or `UnixFd.kt`. The new `[issue #246]` link is a plain URL, so it cannot be an unresolved KDoc reference — the count staying at 8 is the check that nothing else broke.
- **Did not run** the test suites. A comment deletion cannot change behaviour; CI covers them.

Refs #208, #246, #248

### DIFF
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
@@ -162,8 +162,9 @@ expect sealed class Message {
      * `false` deliberately to mean "this reply carries an error" — a different concept from the
      * one this property is named for, sharing the same field. So `if (!isValid) return` discards
      * locally built messages on JVM and keeps them on native. Which of the two definitions this
-     * property should have is unsettled; treat the value as unspecified for anything but native
-     * `msg != null`.
+     * property should have is unsettled — see
+     * [issue #246](https://github.com/Monkopedia/sdbus-kotlin/issues/246) — so treat the value as
+     * unspecified for anything but native's `msg != null`.
      */
     val isValid: Boolean
 
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt
@@ -335,8 +335,9 @@ value class Signature(
  * an explicitly provided fd by either duplicating ([UnixFd] primary constructor)
  * or adopting that fd as-is ([UnixFd.adopt]).
  *
- * **Duplicating and closing are backend-dependent on JVM.** The native backend always `dup(2)`s
- * in the primary constructor — throwing if the syscall fails — and always `close(2)`s on release.
+ * **Duplicating and closing are backend-dependent on JVM.** For any valid (non-negative) fd, the
+ * native backend `dup(2)`s in the primary constructor — throwing if the syscall fails — and
+ * `close(2)`s on release; a negative fd is passed through and never closed on either backend.
  * The JVM backend reaches the descriptor through junixsocket's *optional* native support; when
  * that support is absent (no native binary for the platform, or a consumer where the reflective
  * access it needs is denied, e.g. a JPMS-modular application) the primary constructor returns the
--- a/src/nativeMain/kotlin/com/monkopedia/sdbus/UnixFd.kt
+++ b/src/nativeMain/kotlin/com/monkopedia/sdbus/UnixFd.kt
@@ -36,15 +36,11 @@ import platform.posix.close
 import platform.posix.dup
 import platform.posix.errno
 
-/**
- * A representation of the file descriptor D-Bus type that owns
- * the underlying fd, provides access to it, and closes the fd when
- * the UnixFd is released.
- *
- * UnixFd can be default constructed (owning invalid fd), or constructed from
- * an explicitly provided fd by either duplicating ([UnixFd] primary constructor)
- * or adopting that fd as-is ([UnixFd.adopt]).
- */
+// Intentionally carries no class KDoc: with none here, Dokka renders the `expect`
+// declaration's documentation on this actual too (as it already does for the JVM actual, which
+// likewise has none). A second copy of that prose is what drifted in #208 -- the sentence was
+// corrected here and left wrong on the `expect`, which is the declaration published to
+// consumers. One source, so it cannot happen again in either direction.
 @OptIn(ExperimentalNativeApi::class)
 @Serializable(UnixFd.Companion::class)
 actual class UnixFd internal actual constructor(


monkopedia-reviewer (approved):
**Approved.** Reviewed at `2e150fac8ce06eb31668ff3211e57b7fc96e4eeb` in a fresh clone. Category `docs_only` → tier 1. This is the follow-up to my review of #248; I re-measured every claim rather than accepting the PR body, and on the one point where the PR offered a string count as evidence I replaced it with a structural check.

## Comment-only — verified mechanically

Three files vs the merge base (`8374897`, which is current `main`), +11/-13. Stripping the leading `+`/`-` from every changed line and filtering out `*` / `/**` / `*/` / `//` / blank leaves **zero lines**. `git status --porcelain` empty after the gates, so `api/*.api` did not move.

## 1. Does the native block really inherit the `expect` prose? — yes, and here is a better instrument than the count

The PR's evidence was "3 occurrences of *Nothing closes the fd at scope exit* where there were 2". That count is consistent with the claim but does not establish it — three occurrences in one file could equally be a summary block, a detail block and a nav entry. So I parsed the page's `sourceset-dependent-content` blocks and attributed the prose **by source set**, on both the new output and the `main` baseline:

```
baseline (main @ 8374897)          PR #252
  :/commonMain  EXPECT-PROSE         :/commonMain  EXPECT-PROSE
  :/jvmMain     EXPECT-PROSE         :/jvmMain     EXPECT-PROSE
  :/nativeMain  NATIVE-OLD-PROSE     :/nativeMain  EXPECT-PROSE
```

and page-wide, `A representation of the file descriptor` (the deleted native sentence) went from **1 → 0**. So the three class-header platform blocks are exactly what the count reflects, the native tab now shows the GC-backstop and fd-exhaustion paragraphs, and my note 1 is genuinely closed rather than plausibly closed.

**Nothing else depended on that block.** What it contained — "owns the underlying fd … closes the fd when the UnixFd is released" and the default-constructed / duplicate-or-adopt sentence — is present verbatim in the `expect` at `Types.kt:320-322` and `:334-336`, so no information was lost, only a copy. `@OptIn(ExperimentalNativeApi::class)` and `@Serializable(UnixFd.Companion::class)` remain immediately above the class, the member KDoc on `val fd` at `:47` is untouched, and `compileKotlinLinuxX64` / `linuxArm64ApiBuild` / `ktlintCheck` all pass, so nothing was structurally attached to the comment. The two links the deleted block carried (`[UnixFd]`, `[UnixFd.adopt]`) were link *consumers*, so removing them can only shrink the unresolved-link surface — which is why the count staying at 8 is sufficient here and I did not re-probe detection for `nativeMain`.

**And deletion over re-sync is the right call, not just an equivalent one.** Re-syncing sets up round three of the same drift; the `expect` is the declaration published to consumers, and the JVM `actual` has never carried a copy (`Types.jvm.kt:19-20`), so removing it makes the native side consistent with the JVM side rather than inventing a new pattern. The `//` note in its place is the part that makes it stick.

## 2. `isValid` → #246

Link resolves to the right issue, and #246 tracks *that* question specifically, not merely the adjacent `complete` work: its body has a dedicated section 3 "`isValid` carries two concepts in one field", and its proposed fix item 3 is "split the error-reply flag out of `Metadata.valid` … then align `isValid` on native's 'wraps a constructed message' definition". That is exactly the unsettled definition the KDoc now defers to it. The reword also fixes the possessive (`native's msg != null`).

## 3. The "always" rewording — correct, and not overcorrected

New text: *"For any valid (non-negative) fd, the native backend `dup(2)`s in the primary constructor — throwing if the syscall fails — and `close(2)`s on release; a negative fd is passed through and never closed on either backend."*

Checked the negative-fd half on **both** backends, since that is the part the sentence now asserts jointly:
- native — `checkedDup`: `if (fd < 0) { return fd }` (`UnixFd.kt:95-97`); close path: `singleCall<Int> { if (it >= 0) { close(it) } }` (`:52-55`), reached from `release()` (`:68`). Passed through, never closed. ✓
- JVM — `checkedDup`: `if (fd < 0) return fd` (`Types.jvm.kt:86-87`); `closeFd`: `if (fd < 0) return true` (`:93-94`), and `OwnedFdState.run()` short-circuits on `fd < 0` before ever calling it. Passed through, never closed. ✓
- "throwing if the syscall fails" still holds: `dup(fd).also { if (it < 0) throw createError(errno, "dup failed") }`. ✓

No overcorrection: the sentence still carries the native-unconditional vs JVM-conditional contrast for real descriptors, which is the divergence the paragraph exists to state.

## 4. Where the #248-body correction was recorded

Recording it in #252's body is the right place, and I'd have objected to the alternatives: editing a merged PR's body rewrites the record, and a CHANGELOG entry would be wrong because nothing consumer-visible changed. Repo precedent agrees on the CHANGELOG point — neither #236 (`680e981`) nor #248 (`8374897`) touched it for a KDoc-only change, so this PR omitting it is consistent, not an oversight.

## Gates, re-measured

`ktlintCheck apiCheck :dokkaGeneratePublicationHtml` (JDK 21) → **BUILD SUCCESSFUL in 26s**, working tree clean afterwards. Dokka emits **exactly 8** `Couldn't resolve link` warnings, the same set as on `main`: `DBusWireConnection.kt` ×4 (`[call]`, `[callBlocking]`) and `WireDbusBackend.kt` ×4 (`[LocalJvmSignalBus]`, `[LocalJvmMatchBus]`), all `:/jvmMain`, none naming `Message.kt`, `Types.kt` or `UnixFd.kt`. Reading that as a real absence rests on the deliberate-break probe from #248's review, which showed this exact detector firing 4 warnings for `Types.kt`/`Types.jvm.kt` when a link was broken on purpose — while the build still exited `BUILD SUCCESSFUL`, which is the point the PR body records.

## One non-blocking note

`UnixFd.adopt`'s KDoc at `Types.kt:369` still reads *"Both halves of that contrast are **unconditional** only on the native backend"* — the same word this PR came to scope, thirty lines below where it scoped it. In context it is plainly about the backend-support axis rather than the fd-sign axis, and it points the reader at the class doc that now carries the caveat, so it is not wrong. But if anyone re-reads that sentence in isolation it invites the identical correction. Worth folding into whatever next touches that file; not worth a round trip here.

PR #255, opened 2026-08-09
test(cross): make the wire external suites fail, not skip, where CI guarantees busctl/dbus-monitor

### LINKED ISSUE #188: DBUSMOCK_REQUIRED is not a declared Gradle input, so toggling the guard can replay a stale UP-TO-DATE result
`DBUSMOCK_REQUIRED` (added in #177 / PR #182) is read by the test harness at runtime but is **not declared as a Gradle task input**, so changing it does not invalidate the task. A rerun with the variable newly set can be served as `UP-TO-DATE` from the previous result.

Found by the reviewer verifying PR #182: forcing the failure case required `--rerun-tasks`, because a plain rerun with `DBUSMOCK_REQUIRED=1` silently replayed the prior (ungated, all-skipped) result instead of failing.

## Why this is worth fixing rather than shrugging at

It is the same defect class the variable was introduced to eliminate. #177 existed because the dbusmock suites could report success while running nothing; `DBUSMOCK_REQUIRED` is the guard that makes CI fail loudly when the foreign peer cannot start. But a guard whose value is invisible to the build's up-to-date checking can itself be bypassed by a stale result — the toggle appears to have been honoured when it was never actually applied.

**It does not affect CI today**, which is why this is not urgent: CI runs on fresh checkouts with no prior task outputs, so there is nothing stale to replay. The exposure is (a) local development, where someone toggling the variable to reproduce a CI failure can get a misleading green, and (b) any future move to a shared or persistent Gradle cache in CI, where it would become a live false-green vector rather than a latent one.

## Where

`cross_test/build.gradle.kts` — the test task configuration. The variable is consumed by `DbusmockHarness.jvm.kt` / `DbusmockHarness.linuxX64.kt` via the runtime environment.

Worth checking the same question for the other environment-driven switches the cross-runtime suites use — `DBUSMOCK_PYTHON` has the same shape, and `kdbus.crossRuntimeInterop.reverse.enabled` (PR #185) is a system property rather than an env var, which Gradle treats differently. Whichever of these are read at runtime but undeclared should be handled together rather than one at a time.

## Proposed

Declare the relevant environment variables and system properties as task inputs so a change invalidates the task, e.g. via `inputs.property(...)` or the `Test` task's environment/`systemProperty` wiring, so toggling the guard actually re-runs the tests it guards.

## Verification

The bug reproduces directly: run the dbusmock suites once without `DBUSMOCK_REQUIRED`, then again with `DBUSMOCK_REQUIRED=1` and **without** `--rerun-tasks`, and observe the task reported `UP-TO-DATE` rather than failing. After the fix, the second run must execute and fail. That is a red-first check available without any dbusmock installation, since the failure path is precisely what runs when dbusmock is absent.

### LINKED ISSUE #229: External-tool skips in the wire external suites have no REQUIRED flag, so CI can silently lose the only out-of-process coverage

monkopedia-coder:
Fixes #229.

`#228` converted `WireServeExternalTest` / `WireSignalEmissionExternalTest` from an early return
into an honest `<skipped/>` when `busctl` / `dbus-monitor` is off `PATH`. That is half of what
c5f874c (#182) established. The other half — a `*_REQUIRED` flag set by the CI job that installs
the dependency **on purpose**, so the skip cannot silently delete coverage where the dependency is
guaranteed — was missing. These two suites are the repo's only proof that a JVM-served object is
reachable by a genuinely external process (#90, epic #93 phase 4) and that emission reaches an
out-of-process observer (phase 3b), so a vanished binary would have taken all of it out behind a
green `tests="1" skipped="1"`.

This mirrors c5f874c rather than adding a second mechanism.

## The change (~6 lines of behaviour + a de-duplication)

- **`ExternalToolGate.kt`** — `requireExternalTool(name, capability)`. It also absorbs the
  `findExecutable` helper both suites had copy-pasted, so each call site shrinks from a
  nine-line `if (x == null) { assumeTrue(...); return@runBlocking }` to a three-line call and
  the return type is non-null (no `?:` at the call sites).
  - `EXTERNAL_TOOLS_REQUIRED` unset → `AssumptionViolatedException` — byte-for-byte the same skip
    message #228 shipped.
  - set → `IllegalStateException` naming the missing binary.
- **`full-tests-x64`** sets `EXTERNAL_TOOLS_REQUIRED: "1"` next to the existing
  `DBUSMOCK_REQUIRED: "1"`, with the reason in a comment.
- **`:cross_test:jvmTest` declares the variable as a task input** — see the #188 section.

## Are the binaries actually on the runner?

This is the one thing the issue says cannot be checked locally, so it is answered by measurement,
not assumption: **`ubuntu-24.04` provides both.** `dbus-monitor` comes from `dbus-bin`, pulled in by
the `dbus` package this job already installs; `busctl` ships with systemd. The proof is this PR's own
`full-tests-x64` run: with the flag set an absent binary is a hard failure, so a green run *is* the
statement that both were present and both suites really exercised an external peer — which is also
the observability gap #229 complained about (the job only uploads test-results on failure, so
"present and ran" and "absent and skipped" used to look identical from outside).

## Evidence — all three directions

Whole `:cross_test:jvmTest`, fresh JUnit XML each time, run under `dbus-run-session`. "absent" =
`PATH` pointed at a mirror of `/usr/bin` with `busctl` and `dbus-monitor` removed.

**1. Flag set + binaries absent → FAILS.** (`BUILD FAILED`)

```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="0" failures="1" errors="0" time="0.002"
  <failure message="java.lang.IllegalStateException: EXTERNAL_TOOLS_REQUIRED is set but busctl is not on PATH, so no external peer can call the served object."
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="0" failures="1" errors="0" time="0.001"
  <failure message="java.lang.IllegalStateException: EXTERNAL_TOOLS_REQUIRED is set but dbus-monitor is not on PATH, so no external observer can see the signal."
```

**2. Flag unset + binaries absent → still an honest `<skipped/>`.** (`BUILD SUCCESSFUL`)

```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="1" failures="0" errors="0" time="0.001"
  <skipped message="org.junit.AssumptionViolatedException: busctl is not on PATH, so no external peer can call the served object."
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="1" failures="0" errors="0" time="0.001"
  <skipped message="org.junit.AssumptionViolatedException: dbus-monitor is not on PATH, so no external observer can see the signal."
```

**3. Flag set + binaries present → green, and really ran.** (`BUILD SUCCESSFUL`)

```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="0" failures="0" errors="0" time="0.045"
  <testcase name="externalBusctl_reachesServedObject[jvm]" time="0.045"/>
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="0" failures="0" errors="0" time="0.26"
  <testcase name="emittedSignal_reachesExternalDbusMonitor[jvm]" time="0.26"/>
```

`ktlintCheck apiCheck`: `BUILD SUCCESSFUL`. **`api/*.api` does not move** — the change is entirely
inside `cross_test`'s test source set and the workflow.

## #188: the flag is a declared Gradle input, and the bug is shown red-first

#188 records that `DBUSMOCK_REQUIRED` is read at runtime but **not** declared as a task input, so
toggling it can replay a stale `UP-TO-DATE` result. `EXTERNAL_TOOLS_REQUIRED` would have inherited
that exactly, so it is declared:

```kotlin
inputs.property(
    "EXTERNAL_TOOLS_REQUIRED",
    providers.environmentVariable("EXTERNAL_TOOLS_REQUIRED")
).optional(true)
```

Measured, as two pairs of runs differing **only** in the environment variable, with nothing deleted
in between:

| | undeclared (`DBUSMOCK_REQUIRED`'s shape today) | declared (this PR) |
|---|---|---|
| run 1, flag unset | `> Task :cross_test:jvmTest` → `skipped="1"` | `> Task :cross_test:jvmTest` → `skipped="1"` |
| run 2, flag set | `> Task :cross_test:jvmTest UP-TO-DATE`, **BUILD SUCCESSFUL**, XML byte-identical (same mtime, same `timestamp=`) | `> Task :cross_test:jvmTest FAILED`, **BUILD FAILED**, fresh XML `failures="1"` |

So without the declaration the guard silently does nothing on a warm build; with it, it fires. This
PR fixes the instance it introduces and does **not** close #188 — `DBUSMOCK_REQUIRED`,
`DBUSMOCK_PYTHON` and `kdbus.crossRuntimeInterop.reverse.enabled` are still undeclared and #188 asks
for them to be handled together.

## Scope notes

- **Nothing native.** Both suites are `jvmTest`-only (the owned wire connection is the JVM backend),
  so this does not touch the "native cannot emit `<skipped/>`" problem #240/#242 established — there
  is no native half here to be inconsistent with.
- **Shared workflow file touched:** `.github/workflows/arm-build-test.yaml`, two comment blocks and
  one `env:` entry in `full-tests-x64`. Rebased on `75e0aa5`, after #223 and #238.
- Contributor laptops without the variable behave exactly as they do on `main` today.

### DIFF
--- a/.github/workflows/arm-build-test.yaml
+++ b/.github/workflows/arm-build-test.yaml
@@ -53,6 +53,8 @@ jobs:
           # python3-dbusmock powers the independent-peer harness (the Dbusmock* suites). Because
           # this job installs it deliberately, the test step sets DBUSMOCK_REQUIRED so a missing
           # package fails loudly instead of turning the whole foreign-peer layer into no-ops.
+          # `dbus` likewise provides dbus-monitor (via dbus-bin); busctl ships with systemd on
+          # ubuntu-24.04. Both are guaranteed here, hence EXTERNAL_TOOLS_REQUIRED below.
           sudo apt-get install -y libcurl4-openssl-dev libsystemd-dev dbus python3-dbusmock
 
       - name: Run Full Test Suite on x64
@@ -64,6 +66,12 @@ jobs:
           # This job installs python3-dbusmock above, so the harness must fail — not skip — when
           # it cannot start a dbusmock peer.
           DBUSMOCK_REQUIRED: "1"
+          # busctl and dbus-monitor are guaranteed here (see the install step), and the two suites
+          # that drive them — WireServeExternalTest / WireSignalEmissionExternalTest — are the only
+          # proof a JVM-served object is reachable by, and a JVM-emitted signal observable from, a
+          # genuinely external process (#90, #93 phases 4/3b). Without this they would skip and the
+          # job would stay green if either binary vanished from the image (#229).
+          EXTERNAL_TOOLS_REQUIRED: "1"
         run: |
           set -euo pipefail
           X64_SYSTEMD_LIB="$(find "${GITHUB_WORKSPACE}/libs/x86_64" -maxdepth 2 -type d -name lib | head -n 1 || true)"
--- a/cross_test/build.gradle.kts
+++ b/cross_test/build.gradle.kts
@@ -57,6 +57,14 @@ val reverseInteropEnabled = providers
 // actually configures it — mirroring the gcSoak gate in the root build.
 tasks.named<Test>("jvmTest") {
     filter.excludeTestsMatching("com.monkopedia.sdbus.integration.CrossRuntimeInteropSmokeTest")
+    // EXTERNAL_TOOLS_REQUIRED is read from the test JVM's environment by `requireExternalTool`
+    // (ExternalToolGate.kt), which Gradle's up-to-date check knows nothing about — so without this,
+    // newly setting it would replay the previous, ungated result and the guard would appear to have
+    // been honoured when it never ran (#188, found on DBUSMOCK_REQUIRED, which still has that bug).
+    inputs.property(
+        "EXTERNAL_TOOLS_REQUIRED",
+        providers.environmentVariable("EXTERNAL_TOOLS_REQUIRED")
+    ).optional(true)
 }
 
 // Kotlin/Native's test runner has no runtime-skip primitive, so `DbusmockHarness.skipTest`'s native
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/ExternalToolGate.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/ExternalToolGate.kt
@@ -1,0 +1,59 @@
+/**
+ *
+ * (C) 2024 - 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Project: sdbus-kotlin
+ * Description: High-level D-Bus IPC kotlin library based on sd-bus
+ *
+ * This file is part of sdbus-kotlin.
+ *
+ * sdbus-kotlin is free software: you can redistribute it and/or modify it under the terms of the
+ * GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * sdbus-kotlin is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with
+ * sdbus-kotlin. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.monkopedia.sdbus.integration
+
+import java.io.File
+import org.junit.AssumptionViolatedException
+
+/**
+ * Environment variable that turns a "the external tool is not on PATH" skip into a hard failure.
+ * Set by the CI job that installs those tools on purpose.
+ *
+ * Same shape, and the same reasoning, as `DBUSMOCK_REQUIRED` in `DbusmockHarness` (#182): the skip
+ * is right on a contributor's box, where `busctl` / `dbus-monitor` may legitimately be absent, and
+ * wrong in a job that guarantees them — there a vanished binary would silently convert the only
+ * out-of-process coverage the repo has into a green `skipped="1"` (#229).
+ */
+internal const val EXTERNAL_TOOLS_REQUIRED_ENV = "EXTERNAL_TOOLS_REQUIRED"
+
+/**
+ * Returns the executable [name] found on `PATH`, or aborts the calling test.
+ *
+ * [capability] completes the sentence "<name> is not on PATH, so …" and should name the coverage
+ * that is lost without the tool.
+ *
+ * @throws AssumptionViolatedException when the tool is absent — JUnit records that as a real
+ *   `<skipped/>`, unlike an early return (#227).
+ * @throws IllegalStateException instead, when [EXTERNAL_TOOLS_REQUIRED_ENV] is set.
+ */
+internal fun requireExternalTool(name: String, capability: String): File {
+    val path = System.getenv("PATH").orEmpty()
+    path.split(File.pathSeparator)
+        .map { File(it, name) }
+        .firstOrNull { it.canExecute() }
+        ?.let { return it }
+
+    val reason = "$name is not on PATH, so $capability."
+    check(System.getenv(EXTERNAL_TOOLS_REQUIRED_ENV) == null) {
+        "$EXTERNAL_TOOLS_REQUIRED_ENV is set but $reason"
+    }
+    throw AssumptionViolatedException(reason)
+}
--- a/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireServeExternalTest.kt
+++ b/cross_test/src/jvmTest/kotlin/com/monkopedia/sdbus/integration/WireServeExternalTest.kt
@@ -30,14 +30,12 @@ import com.monkopedia.sdbus.createBusConnection
 import com.monkopedia.sdbus.createObject
 import com.monkopedia.sdbus.method
 import com.monkopedia.sdbus.prop
-import java.io.File
 import java.util.concurrent.TimeUnit
 import kotlin.random.Random
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 import kotlinx.coroutines.runBlocking
-import org.junit.Assume.assumeTrue
 
 /**
  * Proves an object exported by a JVM sdbus-kotlin service on the OWNED wire backend is reachable by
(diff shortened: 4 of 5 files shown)


monkopedia-reviewer (approved):
Independent review at `6418276961`. I reproduced the guard from scratch in my own worktree rather than reading the PR's evidence — a guard that never fires is the same defect one level up.

## Reproduced: all three directions, counted off fresh JUnit XML

`PATH` pointed at a 4664-entry symlink mirror of `/usr/bin` with `busctl` and `dbus-monitor` removed; `--no-daemon` so the test JVM inherits that `PATH`; under `dbus-run-session`.

**1. Flag set + binaries absent → FAILS** (`BUILD FAILED`, exit 1)
```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="0" failures="1" errors="0" time="0.039">
  <failure message="java.lang.IllegalStateException: EXTERNAL_TOOLS_REQUIRED is set but busctl is not on PATH, so no external peer can call the served object."
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="0" failures="1" errors="0" time="0.001">
  <failure message="java.lang.IllegalStateException: EXTERNAL_TOOLS_REQUIRED is set but dbus-monitor is not on PATH, so no external observer can see the signal."
```

**2. Flag unset + binaries absent → still an honest `<skipped/>`** (`BUILD SUCCESSFUL`)
```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="1" failures="0" errors="0" time="0.037">
  <skipped message="org.junit.AssumptionViolatedException: busctl is not on PATH, so no external peer can call the served object."
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="1" failures="0" errors="0" time="0.003">
  <skipped message="org.junit.AssumptionViolatedException: dbus-monitor is not on PATH, so no external observer can see the signal."
```
Skip text is byte-identical to what #228 shipped, so contributor laptops see no change.

**3. Flag set + binaries present → green, and really ran** (`BUILD SUCCESSFUL`)
```
<testsuite name="WireServeExternalTest[jvm]"          tests="1" skipped="0" failures="0" errors="0" time="0.188">
  <testcase name="externalBusctl_reachesServedObject[jvm]" time="0.185"/>
<testsuite name="WireSignalEmissionExternalTest[jvm]" tests="1" skipped="0" failures="0" errors="0" time="0.26">
  <testcase name="emittedSignal_reachesExternalDbusMonitor[jvm]" time="0.26"/>
```
The session bus logged `Connection :1.9 (... comm="/usr/bin/dbus-monitor --session ...") became a monitor`, so a real external observer was attached.

## Reproduced: the #188 claim, as a paired A/B with nothing deleted between runs

Two pairs differing **only** in the environment variable. `build/test-results` was deliberately left in place — deleting it would have invalidated the up-to-date check and measured nothing.

| | undeclared (control: `inputs.property` block removed) | declared (this PR) |
|---|---|---|
| run 1, flag unset | `> Task :cross_test:jvmTest`, `skipped="1"` | `> Task :cross_test:jvmTest`, `skipped="1"` |
| run 2, flag set | `> Task :cross_test:jvmTest UP-TO-DATE`, **BUILD SUCCESSFUL** | `> Task :cross_test:jvmTest FAILED`, **BUILD FAILED** |

On the undeclared control the XML after run 2 was byte-identical to run 1 — same `md5sum` (`a9f5bea3…` / `2942ac1a…`), same mtime (`01:21:59.643250086`), same `timestamp="2026-08-09T05:21:59.511Z"`. The declared version produced fresh XML at `05:20:17` with `failures="1"`. So the declaration is load-bearing, not decorative: without it the guard looks honoured on a warm build while never having run.

## The one gap in "CI green proves the binaries exist"

`full-tests-x64` passing is only proof the binaries are on `ubuntu-24.04` **if the gate is actually live inside that job** — otherwise a green run is equally consistent with "the flag never reached the test JVM and both suites skipped." The step-level `env:` has to survive `dbus-run-session -- bash -lc '…'`, and a login shell is exactly the kind of thing that quietly drops things.

So I reproduced that shape locally: step-level export → `dbus-run-session -- bash -lc` → `./gradlew`, with the binaries absent.
```
inside bash -lc: EXTERNAL_TOOLS_REQUIRED=[1]
> Task :cross_test:jvmTest FAILED
<testsuite ... tests="1" skipped="0" failures="1" ...>
```
The variable propagates and the gate fires through CI's exact wrapper. With that established, this PR's green `full-tests-x64` (6m44s) **is** the statement the issue asked for: both binaries were present on the runner and both suites really exercised an external peer. The workflow comment is accurate — `dbus-monitor` comes from `dbus-bin` (pulled in by the `dbus` package the job already installs) and `busctl` ships with systemd.

`arm-runtime-tests` runs only the native `test.kexe`, and both suites are `jvmTest`-only, so there is no second job that should have been flagged and wasn't.

## Quality

- The de-duplication pays for itself. Each call site drops from a nine-line `if (x == null) { assumeTrue(…); return@runBlocking }` to a three-line call, two copies of a seven-line `findExecutable` are gone, and the non-null return means `busctl.absolutePath` needs no `!!` or `?:`. Real (non-header, non-KDoc) code shrank. No stray copies remain — `findExecutable` has zero hits in the tree.
- Imports are handled correctly rather than mechanically: `java.io.File` is dropped from `WireServeExternalTest` where it became unused and kept in `WireSignalEmissionExternalTest`, which still uses `File.createTempFile`.
- Mirrors `c5f874c` (`*_REQUIRED_ENV` const, KDoc cross-reference, a CI comment saying *why*) instead of inventing a second convention, which is what #229 asked for.
- Scope is right: it declares the input for the flag it introduces and explicitly does **not** close #188. I confirmed `DBUSMOCK_REQUIRED` still has the undeclared-input bug — my control run above is a direct measurement of it — so leaving #188 open is correct, not an oversight.
- `ktlintCheck apiCheck` green locally; `api/*.api` unmoved, as expected for a change confined to `cross_test`'s test source set.
- Nothing here contradicts `AGENTS.md` (its "a test that cannot run its subject must say so; skipping is fine, passing is not" line is precisely this change), `docs/BACKENDS.md`, or `COVERAGE_TODO.md`.

One trivial, non-blocking note: `check(System.getenv(ENV) == null) { "$ENV is set but …" }` inverts the condition relative to the precedent's `if (getenv == null) return null; error(…)`. Same `IllegalStateException`, and the message makes the intent obvious on the second read. Not worth a round trip.

Approving and merging.

Issue #256, opened 2026-08-09
JVM Message.isValid: unshare it from the error-reply flag and align it with native's "wraps a constructed message"


monkopedia-coder:
Split out of #246, which fixed the other three items in that cluster (`complete` on `isAtEnd`/`copyTo`/`rewind`, and `copyTo` overwriting the destination header). This one is left because it is the only item in the cluster that is a **public behaviour change with live internal readers**, and one of those readers is in a file under concurrent edit.

Verified against `main` @ `75e0aa5`.

## The two concepts

`Message.Metadata.valid` (`src/jvmMain/kotlin/com/monkopedia/sdbus/Message.jvm.kt:40`, defaulting `false`) currently carries both:

1. **"this message was built by a path that produces a real message"** — set `true` by every construction site that makes a call, reply or signal (`WireDbusBackend.kt:288,462,501,580,766`, `WireDbusObject.kt:171,409,467`, `WireServe.kt:167`, `Message.jvm.kt:462,469`).
2. **"this reply is usable"** — `WireDbusBackend.invalidReply` (`:637-647`) builds a reply with `valid = false` to mean *no reply was produced*, and `MethodCall.createErrorReply` (`Message.jvm.kt:471-474`) sets `valid = false` on an error reply.

The public `isValid` (`Message.jvm.kt:182-183`) is wired to concept 1's field but is read for concept 2.

Native has neither: `isValid` is `get() = msg != null` (`Message.native.kt:586-587`), `true` for every message the library hands out. So `createPlainMessage().isValid` is `false` on JVM and `true` on native, and `if (!isValid) return` discards locally built messages on one backend only.

## Why it was not fixed in #246

Aligning `isValid` with native means making it unconditionally `true` on JVM and repointing the two guards that actually want concept 2 at an internal field:

- `src/jvmMain/kotlin/com/monkopedia/sdbus/Message.jvm.kt:447` — `MethodCall.send`
- `src/jvmMain/kotlin/com/monkopedia/sdbus/internal/jvmdbus/WireDbusBackend.kt:487` — `WireDbusBackend.callMethod`

Both read `result.reply.isValid` and throw `"invalid reply"` when it is `false`. If `isValid` flips to always-`true` without repointing them, both guards silently become no-ops and `invalidReply` stops being detectable — a live regression, not a latent one. `WireDbusBackend.kt` was being edited under #247 while #246 was in flight, so #246 stayed out of it.

[...]

PR #257, opened 2026-08-09
fix(jvm): honour Message.complete on isAtEnd/copyTo/rewind, and stop copying the header in copyTo

### LINKED ISSUE #246: JVM Message: make 'complete' actually do something on isAtEnd/copyTo/rewind, stop copying metadata in copyTo, and unshare isValid with the error-reply flag

monkopedia-coder:
Three of #246's four items. The fourth (`isValid`) is split to #256 — reasoning below.

## What changed

`src/jvmMain/kotlin/com/monkopedia/sdbus/Message.jvm.kt` — three actuals, ~15 lines.

| | before | after |
|---|---|---|
| `isAtEnd(complete)` | `readIndex >= payload.size`, flag unread | `… && (!complete \|\| enteredVariantValues.isEmpty())` |
| `copyTo(dst, complete)` | whole payload from index 0, plus `dst.metadata = metadata` | from the read cursor, consuming it: one value when `complete` is `false`, the rest when `true`; header untouched |
| `rewind(complete)` | always `readIndex = 0` + clear the variant stack | `complete` does that; `!complete` rewinds the open container only, which inside an entered variant has nothing to move |

An entered variant is the whole of "any open containers" on JVM — the payload model flattens arrays, structs and dict entries, so their enter/exit actuals are already no-ops. That is stated in the code and in the KDoc rather than left for the next reader to rediscover.

## The `complete` semantics are sd-bus's, re-measured — and the old KDoc was wrong about native too

`sd_bus_message_copy(3)`: *"If `all` is false, a single complete type is copied (basic or container). If `all` is true, the contents are copied until the end of the currently open container or the end of source."* Both branches copy **from the cursor** and consume.

The `@param complete` on `copyTo` said `true` copies *"the whole message"* and `false` copies *"from the current cursor"* — i.e. it was wrong about **both** backends, and #248 corrected the surrounding prose without touching that line. Corrected here, with a note saying what it used to claim (the #221 style). `sd_bus_message_at_end(3)` and `sd_bus_message_rewind(3)` confirm the other two; those `@param`s were already right.

Corroboration from the repo itself: `MessageCommonTest.aMessage_CreatesDeepCopyWhenExplicitlyCopied` has to call `msg.rewind(true)` after `copyTo(…, true)` before it can read `msg` again — a test that only makes sense if the copy consumes.

## PR #248's KDoc

#248 (`8374897`) documented all four items as divergences in `commonMain`. Leaving those paragraphs in place while changing the behaviour would ship a newly-false doc — the exact defect #179/#178/#202 have been removing — so the `isAtEnd`, `copyTo` and `rewind` divergence notes are **deleted** and replaced with the now-true contract. **#248's `isValid` paragraph is deliberately left alone**: it still describes the code exactly, because that item is not fixed here.

Verified against the rendered output, not just the source: `grep -c "honoured on native only"` over `build/dokka/…/-message/{copy-to,rewind,is-at-end}.html` returns **0, 0, 0**, and the three replacement sentences return **1, 1, 1**. `:dokkaGeneratePublicationHtml` — BUILD SUCCESSFUL, 8 `Couldn't resolve link` warnings, all pre-existing in untouched `jvmMain` files (`DBusWireConnection.kt` ×4, `WireDbusBackend.kt` ×4), none naming `Message.kt`.

## Evidence — red first, on both backends

The three common cases assert one contract and run on both backends. **Native passed all three unmodified**, so they encode sd-bus's behaviour rather than the JVM's answer written down.

`MessageApiParityCommonTest` (commonTest), from the JUnit XML of a run with the fix reverted:

| case | JVM before | native before |
|---|---|---|
| `copyTo_copiesOneValueWhenNotComplete_andTheRemainderWhenComplete` | FAILED — `copyTo(complete = false) must copy exactly one value, not the whole body` | passed |
| `isAtEnd_withCompleteTrue_isFalseWhileAContainerIsStillOpen` | FAILED — `isAtEnd(complete = true) must be false while a container is still open` | passed |
| `rewind_withCompleteFalse_rewindsTheOpenContainerNotTheMessage` | FAILED — `rewind(complete = false) must rewind the open container, not the whole message expected:<42> but was:<7>` | passed |

`JvmMessageSupportTest.copyTo_leavesTheDestinationHeaderAlone` — FAILED: `copyTo must copy contents, not identity: the destination keeps its own header expected:<org.example.Copy> but was:<null>`. This one is JVM-only on purpose: the common surface has no way to build a header-bearing message without a bus, and native's `sd_bus_message_copy` is body-only by construction. Said plainly rather than dressed up as cross-backend.

Red run, counted from fresh XML: `:jvmTest` **340 tests, 4 failed**. Green run: `:jvmTest` **tests=340 failures=0 errors=0 skipped=0**, `:linuxX64Test` **tests=309 failures=0 errors=0 skipped=0** — same 340, so nothing was dropped to get there.

**No `TestBackendCapabilities` flag added or changed.** Both backends assert the identical contract and both are green, so there is nothing for a flag to encode; adding one would record a divergence that no longer exists. (#230's citation rule and #216's unflipped-flag rule therefore do not bite — no flag moved.)

## `isValid` split to #256

`isValid` is the one item that is a public behaviour change with **live** internal readers. Aligning it with native means making it unconditionally `true` on JVM and repointing the two guards that actually want the other concept — `Message.jvm.kt:447` and `WireDbusBackend.kt:487`, both `if (!result.reply.isValid) throw "invalid reply"`. Flip `isValid` without repointing them and both guards silently become no-ops, so `WireDbusBackend.invalidReply` stops being detectable: a live regression, not a latent one. `WireDbusBackend.kt` was under concurrent edit (#247) while this was in flight.

#256 also corrects a detail #246 got slightly wrong: `Message.jvm.kt:446` rethrows an error reply *before* the `isValid` check on the next line, so `createErrorReply`'s `valid = false` never reaches that guard — the real dependent is `invalidReply`, whose meaning is "no reply was produced", not "an error reply".

## Stale / out-of-scope items

- **Every `file:line` in #246 verified against `main`** — `Message.jvm.kt:186-195`, `:189`, `:182-183`, `:40`, `:447`, `:471-474`, `:533-535`, `Message.native.kt:586-587`, `:591-609`, `MessageApiParityCommonTest.kt:40`. **None stale.**
- #246's claim that `Message.append(Variant)` is dead on JVM is correct, but the wider claim that it is the only route into `copyTo` is **not**: `Variant.serializeTo` / `Variant.deserializeFrom` (`Types.kt:144,153`) are common code and live on JVM via `WireDbusBackend.variantToPayload`, `WireDbusBackend.wireVariant` and `JvmValueCodec.slotVariant`. Checked each: all three pass a throwaway `createPlainMessage()` holding exactly one value with default metadata, so the old "copy everything + copy the header" and the new "copy one value from the cursor" coincide there — which is why the whole `:jvmTest` suite is green.
- `PeekedType.type` returning `null` for non-end cases is in **#248's** table but is **not** one of #246's four items, so it is untouched. #248's KDoc for it remains accurate.

## Gates

`./gradlew ktlintCheck apiCheck` — BUILD SUCCESSFUL. **`api/*.api` does not move**: `git status --porcelain` after the run lists only the five edited files. No public signature changes — `isAtEnd`/`copyTo`/`rewind` keep their declarations, only their behaviour and KDoc change.

**Breaking-change assessment: not breaking, but observable.** No API or ABI change. The behaviour change replaces silently wrong answers with correct ones on paths where nothing can have been relying on the old result beneficially — `copyTo` now consumes and no longer clobbers the destination header, `rewind(false)` no longer escapes its container, `isAtEnd(true)` no longer reports the end with a variant open. The one shape a consumer could notice is `copyTo` advancing the source cursor; `Variant.serializeTo` (the only in-tree `complete = true` caller) rewinds before every copy, so it is unaffected. CHANGELOG entry added under `[Unreleased] → Fixed`.

Fixes #246
Refs #248, #256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,21 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   (`busctl`, d-feet, other language bindings), which previously described a fire-and-forget method
   as ordinary request/reply. The JVM backend still does not serve this annotation; that is part of
   #193. (#197)
+- **`Message.isAtEnd`, `Message.copyTo` and `Message.rewind` now honor their `complete` parameter on
+  the JVM backend, and `copyTo` no longer replaces the destination's header.** All three JVM actuals
+  ignored the flag, so they returned a silently wrong answer rather than an error: `isAtEnd(true)`
+  reported `true` with a variant still entered, `rewind(false)` inside a variant rewound to the start
+  of the body and dropped the container, and `copyTo` copied the whole payload from index 0 whatever
+  the flag said. `copyTo` additionally assigned the source's metadata onto the destination —
+  interface, member, path, sender, destination and sender credentials — making it a copy of identity
+  rather than of contents; native's `sd_bus_message_copy` moves body data only. The JVM behavior now
+  matches sd-bus: `copyTo` takes values from the read cursor and consumes them, one complete value
+  when `complete` is `false` and the rest of the open container when it is `true`, and `complete` is
+  honored against the variant stack (the only container the JVM payload model keeps open — arrays,
+  structs and dict entries are flat there). `MessageApiParityCommonTest` now pins all three on both
+  backends; native already satisfied every assertion unchanged. The `@param complete` KDoc on
+  `copyTo` was also wrong about *both* backends — it described `true` as copying "the whole message"
+  — and has been corrected. (#246)
 
 ## [1.0.1] - 2026-07-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,7 +95,18 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   structs and dict entries are flat there). `MessageApiParityCommonTest` now pins all three on both
   backends; native already satisfied every assertion unchanged. The `@param complete` KDoc on
   `copyTo` was also wrong about *both* backends — it described `true` as copying "the whole message"
-  — and has been corrected. (#246)
+  — and has been corrected. The consumer-visible consequence runs through public `Variant`: because
+  `Variant.deserializeFrom` is built on `copyTo`, a second `deserializeFrom` from the same source
+  message used to re-read the first value on JVM and never advanced the source; it now reads the
+  next value and consumes it, which is what native always did. (#246)
+- **A `Variant` is no longer left unusable by a `get` of the wrong type on the JVM backend.** Native
+  validates the contained signature in `sd_bus_message_enter_container`, so a mismatched extraction
+  never enters and leaves the variant untouched; the JVM backend enters first and only detects the
+  mismatch while decoding, so the failed `get` left the variant entered and every later `get`,
+  `peekValueType` and `containsValueOfType` on it failed — the last two being the inspection path
+  the `Variant` documentation recommends. `Variant.get` now unwinds the container when extraction
+  throws. This was previously masked on JVM by `rewind`'s ignored `complete` parameter, so it
+  surfaced only once that was fixed. (#246)
 
 ## [1.0.1] - 2026-07-15
 
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
@@ -163,7 +163,7 @@ expect sealed class Message {
      * one this property is named for, sharing the same field. So `if (!isValid) return` discards
      * locally built messages on JVM and keeps them on native. Which of the two definitions this
      * property should have is unsettled — see
-     * [issue #246](https://github.com/Monkopedia/sdbus-kotlin/issues/246) — so treat the value as
+     * [issue #256](https://github.com/Monkopedia/sdbus-kotlin/issues/256) — so treat the value as
      * unspecified for anything but native's `msg != null`.
      */
     val isValid: Boolean
@@ -174,9 +174,10 @@ expect sealed class Message {
     /**
      * Whether the read cursor has reached the end of the message body.
      *
-     * The JVM backend keeps only a variant on its container stack — arrays, structs and dict
-     * entries are flat in its payload model — so inside any other container it answers about the
-     * message body rather than about that container.
+     * **`complete = false` does not stop at the end of the current container on JVM**, inside a
+     * variant included: that backend answers whether the cursor has consumed the flat body, so
+     * with any body value left after the container it reports `false` where native reports `true`.
+     * `complete = true` agrees on both.
      *
      * @param complete When `true`, also requires that any open containers have been exited
      */
@@ -204,9 +205,10 @@ expect sealed class Message {
     /**
      * Resets the read cursor to the start of the message, or of the currently open container.
      *
-     * The JVM backend keeps only a variant on its container stack (see [isAtEnd]), and the cursor
-     * inside an entered variant always addresses that variant's single value, so a non-complete
-     * rewind there has nothing to move.
+     * A variant is the only container the JVM backend keeps on its container stack — arrays,
+     * structs and dict entries are flat in its payload model — and the cursor inside an entered
+     * variant always addresses that variant's single value, so a non-complete rewind there has
+     * nothing to move.
      *
      * @param complete When `true`, rewind past all containers to the very beginning; when `false`,
      *   rewind only the currently open container and leave it open. If no container is open the
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Message.kt
@@ -174,27 +174,27 @@ expect sealed class Message {
     /**
      * Whether the read cursor has reached the end of the message body.
      *
-     * **[complete] is honoured on native only.** The JVM backend ignores it and answers a third
-     * question instead — whether the cursor has consumed the flat payload — with no regard to
-     * container nesting, so `isAtEnd(true)` there can report `true` while containers are still
-     * open, and `isAtEnd(false)` does not stop at the end of the current container.
+     * The JVM backend keeps only a variant on its container stack — arrays, structs and dict
+     * entries are flat in its payload model — so inside any other container it answers about the
+     * message body rather than about that container.
      *
      * @param complete When `true`, also requires that any open containers have been exited
      */
     fun isAtEnd(complete: Boolean): Boolean
 
     /**
-     * Copies the contents of this message into [destination].
+     * Copies body values out of this message into [destination], starting at this message's read
+     * cursor and consuming what it copies.
      *
-     * **[complete] is honoured on native only**, and the JVM backend copies more than the
-     * contents. On JVM the flag is ignored — the whole payload is always copied from the start,
-     * never from the cursor — and the copy additionally overwrites [destination]'s metadata
-     * (interface, member, path, sender, destination, validity flag and sender credentials) with
-     * this message's. Native's `sd_bus_message_copy` moves body data only and leaves the
-     * destination's header alone.
+     * Only body data moves: [destination] keeps its own header — interface, member, path, sender,
+     * destination and sender credentials — so this copies contents, not identity.
      *
      * @param destination Message to copy into
-     * @param complete When `true`, copy the whole message; otherwise copy from the current cursor
+     * @param complete When `true`, copy every value remaining in the currently open container;
+     *   when `false`, copy exactly one complete value (a basic value, or a whole container).
+     *   (This corrects the parameter description published up to and including 1.0.1, which said
+     *   `true` copies "the whole message" and `false` copies "from the current cursor"; both
+     *   branches copy from the cursor, and neither backend ever rewound first.)
      */
     fun copyTo(destination: Message, complete: Boolean)
 
@@ -202,14 +202,15 @@ expect sealed class Message {
     fun seal()
 
     /**
-     * Resets the read cursor to the start of the message.
+     * Resets the read cursor to the start of the message, or of the currently open container.
      *
-     * **[complete] is honoured on native only.** The JVM backend ignores it and always rewinds to
-     * the very beginning of the body, so `rewind(false)` inside a container does not return the
-     * cursor to the start of that container — it returns it to the start of the message, and the
-     * next read yields a value from there rather than failing.
+     * The JVM backend keeps only a variant on its container stack (see [isAtEnd]), and the cursor
+     * inside an entered variant always addresses that variant's single value, so a non-complete
+     * rewind there has nothing to move.
      *
-     * @param complete When `true`, rewind past all containers to the very beginning
+     * @param complete When `true`, rewind past all containers to the very beginning; when `false`,
+     *   rewind only the currently open container and leave it open. If no container is open the
+     *   two are equivalent.
      */
     fun rewind(complete: Boolean)
 
--- a/src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt
+++ b/src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt
@@ -120,8 +120,18 @@ class Variant constructor() {
         msg.rewind(false)
 
         msg.enterVariant(signature.value)
-        @Suppress("UNCHECKED_CAST")
-        val v: T = msg.deserialize(type, module)
+        val v: T = try {
+            msg.deserialize(type, module)
+        } catch (e: Throwable) {
+            // Leave the variant readable after a failed extraction. Native's enter validates the
+            // contained signature before entering, so a mismatched get never gets this far; the
+            // JVM backend has already consumed the value by the time decoding rejects it, and
+            // without this unwind the variant stays entered and every later get, peekValueType
+            // and containsValueOfType on it fails. The exit is best-effort because it can itself
+            // fail on a partially read container, and it must not replace the real error.
+            runCatching { msg.exitVariant() }
+            throw e
+        }
         msg.exitVariant()
         return v
     }
--- a/src/commonTest/kotlin/com/monkopedia/sdbus/MessageApiParityCommonTest.kt
+++ b/src/commonTest/kotlin/com/monkopedia/sdbus/MessageApiParityCommonTest.kt
@@ -2,6 +2,7 @@ package com.monkopedia.sdbus
 
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 class MessageApiParityCommonTest {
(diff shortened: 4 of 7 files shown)


monkopedia-reviewer (changes requested):
Reviewed `0f7b164f88bd71d520474add7042e744658b7253` (head confirmed with `git ls-remote`, not the API).

Everything in the PR body that I could re-measure, I re-measured on a fresh clone rather than reading — and it all held. The blocking findings below are two things the PR body does not claim and no test covers, both of which I reproduced.

---

## Verdict: request changes

## What I independently reproduced

**Red, counted from fresh JUnit XML.** Branch with *only* the production hunk reverse-applied (`git checkout 75e0aa5 -- src/jvmMain/kotlin/com/monkopedia/sdbus/Message.jvm.kt`; `test-results` deleted first, 38 XML files parsed):

```
:jvmTest  tests=340 failures=4 errors=0 skipped=0
```

and the four are exactly the four claimed, with the claimed messages:

| case | message |
|---|---|
| `MessageApiParityCommonTest.copyTo_copiesOneValueWhenNotComplete_andTheRemainderWhenComplete` | `copyTo(complete = false) must copy exactly one value, not the whole body` |
| `MessageApiParityCommonTest.isAtEnd_withCompleteTrue_isFalseWhileAContainerIsStillOpen` | `isAtEnd(complete = true) must be false while a container is still open` |
| `MessageApiParityCommonTest.rewind_withCompleteFalse_rewindsTheOpenContainerNotTheMessage` | `rewind(...) must rewind the open container, not the whole message expected:<42> but was:<7>` |
| `JvmMessageSupportTest.copyTo_leavesTheDestinationHeaderAlone` | `copyTo must copy contents, not identity … expected:<org.example.Copy> but was:<null>` |

**Green, unmodified head:** `:jvmTest tests=340 failures=0 errors=0 skipped=0`, `:linuxX64Test tests=309 failures=0 errors=0 skipped=0`. All three new common cases report PASS per-testcase on **both** `[jvm]` and `[linuxX64]`, and the 5-file diff contains no `nativeMain` file — so "native passed all three unmodified" is exact.

**Gates:** `./gradlew ktlintCheck apiCheck` BUILD SUCCESSFUL; `git status --porcelain` **empty** immediately afterwards, so `api/*.api` did not move. No declaration changed.

**`TestBackendCapabilities`:** untouched (`git diff 75e0aa5..pr257 -- 'src/*Capabilities*'` is empty), and no flag covers message-cursor semantics. I agree none was owed here: both backends now assert one contract and both are green, so there is no divergence for a flag to encode — #230's citation rule and #216's unflipped-flag rule genuinely do not bite.

**Man pages, checked on this host.** `sd_bus_message_copy(3)`: *"If all is false, a single complete type is copied (basic or container). If all is true, the contents are copied until the end of the currently open container or the end of source."* The rewritten `@param complete` matches it, and the claim that the 1.0.1 wording was wrong about **native** too is correct. `sd_bus_message_rewind(3)` matches the rewritten `rewind` `@param` including the "if no container is open, complete has no effect" clause. `sd_bus_message_at_end(3)` matches the retained `complete = true` clause.

**`copyTo` blast radius — measured, not argued.** `Variant.deserializeFrom` is public and sits directly on the changed path, so I ran it against a two-variant source on all three states:

| | 1st `deserializeFrom` | 2nd `deserializeFrom` | source consumed |
|---|---|---|---|
| JVM before | `1` | **`1`** | **no** |
| JVM after | `1` | `2` | yes |
| native | `1` | `2` | yes |

So the consume change *fixes* a silent public-API divergence rather than creating one. `Message.append(Variant)` is indeed referenced only from `nativeMain/.../MessageEncoder.kt:244,261` (dead on JVM), and the three live JVM `Variant` sites each pass a single-value throwaway `createPlainMessage()`. `MessageCommonTest.aMessage_CreatesDeepCopyWhenExplicitlyCopied`'s `msg.rewind(true)` is now load-bearing on JVM exactly as it always was on native — good corroboration.

**#248's KDoc:** no surviving sentence in `commonMain` says JVM ignores `complete` or that `copyTo` overwrites the destination metadata (`grep` for the old phrasings returns nothing). The retained `isValid` paragraph is still true of the code as merged. Both confirmed.

---

## Blocking

### 1. `rewind(false)` was the only thing repairing a poisoned `Variant`, and removing that repair creates a **new JVM-only divergence** on public API

`Message.jvm.kt` `enterVariant` consumes the payload slot and pushes onto `enteredVariantValues` **without validating the signature** — the mismatch surfaces later, inside `deserialize`. So on JVM a `Variant.get<T>()` with the wrong `T` throws *after* the container is entered, leaving `readIndex` past the end and a frame on the stack. Until this PR, the `msg.rewind(false)` at the top of `Variant.get` (`Types.kt:120`) cleared that stack and the Variant healed itself. It no longer does.

Native does not have the problem at all: `sd_bus_message_enter_container` validates the signature and fails *before* entering, so the message is untouched.

Measured — `Variant("a string")`, a failed `get<Int>()`, then normal recovery:

| | after failed `get<Int>()`: `get<String>()` | `peekValueType()` |
|---|---|---|
| JVM before | `"a string"` | `"s"` |
| **JVM after this PR** | **throws** `Message.enterVariant failed: no remaining payload` | **`null`** |
| native | `"a string"` | `"s"` |

The Variant is permanently unusable on JVM after any failed `get`. That also takes down `containsValueOfType<T>()` (it is `signature.value == peekValueType()`, so it returns `false` forever) and `toString()` (`Variant(null)`) — and `containsValueOfType`/`peekValueType` is precisely the recovery path the `Variant` class KDoc tells callers to use: *"Use [containsValueOfType] or [peekValueType] to inspect the contained type before extracting it."*

So this is not parity-with-native: the pre-PR JVM matched native here and the post-PR JVM does not. What the old `rewind(false)` was doing was masking a JVM defect (a non-atomic `enterVariant`); this PR removes the mask without fixing the defect, in a repo whose current wave is about removing exactly this class of divergence. Nothing in the 340/309 green suite reaches it, which is why it is invisible.

Cheapest correct fix, and it is the one native's atomicity implies: make `Variant.get` restore its own state on failure rather than relying on the next call's `rewind` —

```kotlin
msg.enterVariant(signature.value)
val v: T = try {
    msg.deserialize(type, module)
} finally {
    msg.exitVariant()
}
return v
```

(or validate the signature in the JVM `enterVariant`, which currently ignores its `signature` parameter entirely). Either way please add a parity case — "a `Variant` survives a `get` of the wrong type" — since it passes on native today and on JVM before this PR, so it is a real contract and not a JVM concession.

### 2. The rewritten `isAtEnd` KDoc points the reader away from the one divergence that remains

`Message.kt:177-179` now reads:

> The JVM backend keeps only a variant on its container stack — arrays, structs and dict entries are flat in its payload model — so inside **any other container** it answers about the message body rather than about that container.

"any other container" says the variant case is fine. Measured, with a variant that is not the last body value (`[variant(i:42), 7]`, enter, read 42):

| | `isAtEnd(false)` | `isAtEnd(true)` |
|---|---|---|
| JVM (before **and** after) | `false` | `false` |
| native | **`true`** | `false` |

`complete = false` asks about the *currently open container* (`sd_bus_message_at_end(3)`), and JVM answers `readIndex >= payload.size` — about the body — inside a variant too. The paragraph this PR **replaced** was accurate on this point (*"with no regard to container nesting … `isAtEnd(false)` does not stop at the end of the current container"*); the replacement is not. Since the whole argument of the PR is that a doc which no longer describes the code must not ship, this one should say "inside any container, a variant included". `isAtEnd(true)` — the half this PR actually fixes — is right on both backends; only the `false` sentence needs widening.

---

## Non-blocking, worth fixing in the same push

3. **`Message.kt:166` will point at a closed issue.** The retained `isValid` paragraph says *"Which of the two definitions this property should have is unsettled — see [issue #246]"*, and this PR body says `Fixes #246`, so merging closes it. The unsettled question now lives in #256. One token: `#246` → `#256`.

4. **The CHANGELOG names only `Message.isAtEnd/copyTo/rewind`**, but the consumer-visible consequence measured above is on `Variant.deserializeFrom` — public, and its JVM answer changes (2nd call used to return the 1st value and never advanced the source). It is a strict fix and it moves JVM onto native's answer, but a consumer reading the entry will not connect `copyTo` to it. One clause in the existing entry covers it.

---

I have not re-litigated the split to #256: `isValid` genuinely has two live internal readers (`Message.jvm.kt:447`, `WireDbusBackend.kt:487` — read with `/usr/bin/grep -a`, the file is `file`-flagged `data`), flipping it without repointing them really does no-op both guards, and #256 states that plus the `invalidReply`-not-`createErrorReply` correction. That is a legitimate split, not scope-dodging.

The rest of this change is good work and the evidence in the PR body was accurate everywhere I checked it.

monkopedia-coder:
Pushed `bfe6fb2`. Both blocking findings were real, both reproduced before fixing, and thank you for probing the half nobody had flagged — your closing note is right, the risk was in the contract fix and not in the metadata change I expected to have to defend.

## 1. Poisoned `Variant` — fixed, red reproduced first

Reproduced exactly as measured. New case `TypesCommonTest.aVariant_RemainsReadableAfterAGetOfTheWrongType`, on the PR head before the fix:

- `[jvm]` **FAILED** — `SdbusException: org.freedesktop.DBus.Error.AccessDenied: Message.enterVariant failed: no remaining payload`
- `[linuxX64]` **PASSED** unmodified — `tests="19" failures="0"`

So it is a contract, not a JVM concession, as you said.

Fixed in `Variant.get` (`Types.kt`) rather than by teaching JVM's `enterVariant` to validate its `signature` parameter. Both of your suggestions work; I took this one because validating in `enterVariant` means standing up a *second* signature-enforcement policy next to the one `Message.deserialize` already has, and that one is deliberately tuned (`shouldEnforceSignature` + `isSignatureCompatible`, which relaxes containers because value-based inference cannot recover element types of an empty collection). A stricter, differently-tuned check on the enter path could start rejecting variants that decode fine today — a worse trade than the unwind.

One deviation from your snippet: `finally` would let a failing `exitVariant` replace the real error, and on native `sd_bus_message_exit_container` genuinely can fail on a partially read container. So the exit is best-effort on the failure path only:

```kotlin
msg.enterVariant(signature.value)
val v: T = try {
    msg.deserialize(type, module)
} catch (e: Throwable) {
    runCatching { msg.exitVariant() }
    throw e
}
msg.exitVariant()
return v
```

On JVM this leaves the stack empty and `readIndex` past the value, and the `rewind(false)` at the top of the next `get` resets it — so the recovery you measured (`get<String>()` → `"a string"`, `peekValueType()` → `"s"`, `containsValueOfType<String>()` → `true`) is restored, and `@Suppress("UNCHECKED_CAST")` went away with the change since it was already unused.

## 2. `isAtEnd` KDoc — corrected

You are right that "any **other** container" reads as "the variant case is fine", and it is not. Rewritten to name it:

> **`complete = false` does not stop at the end of the current container on JVM**, inside a variant included: that backend answers whether the cursor has consumed the flat body, so with any body value left after the container it reports `false` where native reports `true`. `complete = true` agrees on both.

That is your `[variant(i:42), 7]` measurement stated directly, and it separates the half this PR fixes from the half it does not. I also dropped the `(see [isAtEnd])` cross-reference in the `rewind` KDoc, which pointed at the sentence you found wrong, and inlined the fact instead.

## 3 & 4 — both taken

- `Message.kt` `isValid` now links **#256** instead of #246.
- CHANGELOG gains the `Variant.deserializeFrom` clause (second call used to re-read the first value and never advance the source; now matches native), plus a second entry for the `Variant.get` fix — including that it was masked by `rewind`'s ignored `complete` and surfaced only once that was fixed.

## Re-verified at `bfe6fb2`

Counted from fresh JUnit XML after deleting `test-results`:

```
:jvmTest      tests=341 failures=0 errors=0 skipped=0
:linuxX64Test tests=310 failures=0 errors=0 skipped=0
```

341/310 = the previous 340/309 plus the one new case, which reports PASS on both `[jvm]` and `[linuxX64]`.

`ktlintCheck apiCheck :dokkaGeneratePublicationHtml` — BUILD SUCCESSFUL; `git status --porcelain` afterwards lists only the four edited source files, so `api/*.api` still does not move. Dokka: the stale phrasings return **0** on all three of `copy-to.html` / `rewind.html` / `is-at-end.html` (now including "any other container"), the new isAtEnd sentence returns **1**, and `is-valid.html` contains `issues/256` — checked against the rendered HTML, not the source.

monkopedia-reviewer (approved):
Re-reviewed `bfe6fb223297ba92a68dfb77f09af4d5ccaafc74` (head read with `git ls-remote`). Both blocking findings are properly fixed. Re-measured rather than read, as asked.

## Re-measured

**Green at `bfe6fb2`**, fresh XML after deleting `test-results`:

```
:jvmTest       xmlfiles=38  tests=341 failures=0 errors=0 skipped=0
:linuxX64Test  xmlfiles=30  tests=310 failures=0 errors=0 skipped=0
```

340→341 and 309→310, i.e. exactly the one new case and nothing dropped to get there. `aVariant_RemainsReadableAfterAGetOfTheWrongType` reports PASS per-testcase on both `[jvm]` and `[linuxX64]`, as do all four earlier cases.

**Red for the new case**, reverting *only* the `Variant.get` hunk (`git checkout 0f7b164 -- src/commonMain/kotlin/com/monkopedia/sdbus/Types.kt`):

| | result |
|---|---|
| `[jvm]` | **FAILED** — `SdbusException: … Message.enterVariant failed: no remaining payload` |
| `[linuxX64]` | **PASSED** unmodified (`TypesCommonTest` 19 tests, 0 failures) |

That is the exact error I measured in the original review, so the case is a live gate on the real defect and not a restatement of the fix. And the JVM suite came back **341 tests / 1 failure** with that hunk reverted — only the new case — which also shows the `Variant.get` fix is not quietly propping up the `Message.jvm.kt` work.

**Gates:** `ktlintCheck apiCheck` BUILD SUCCESSFUL; `git status --porcelain` empty both before and after, so `api/*.api` still does not move. All 15 required checks pass on this head (`full-tests-x64`, `arm-runtime-tests`, `native-build-x64`, `static-checks`, both `build-samples`, 10× `pr-smoke`); `cleaner-soak`/`nightly-stress` are nightly-gated `skipping`, which is their expected state on a PR.

## On the two fixes

**1.** Your deviation from my snippet is right and I'd have been wrong to insist on `finally`: Kotlin does not chain a `finally`-thrown exception as suppressed, it replaces the in-flight one outright, so `catch { runCatching { exitVariant() }; throw e }` is the only shape that guarantees the caller sees the decode error. Declining to validate in JVM's `enterVariant` is also the better call — a second signature-enforcement policy beside `shouldEnforceSignature`/`isSignatureCompatible` is exactly the kind of thing that starts rejecting payloads that decode fine today, and `Variant.get` is where the knowledge of the intended type actually lives. Dropping the now-vestigial `@Suppress("UNCHECKED_CAST")` while you were in there is a fair cleanup.

**2.** The `isAtEnd` KDoc now states the divergence I measured, in the direction I measured it (`false` on JVM vs `true` on native with a body value left after the container), and separates the `complete = true` half this PR fixes from the `complete = false` half it does not. Removing the `(see [isAtEnd])` pointer from `rewind` was the right follow-through — that cross-reference is what carried the wrong claim into a second doc. The `rewind` paragraph restates the flat-payload fact directly and is accurate against `nextDeserializedValue`, which reads an entered variant's value without consuming it.

**3./4.** `isValid` now links #256, and the CHANGELOG's `Variant.deserializeFrom` clause matches my measurement verbatim (second call re-read the first value and never advanced the source on JVM; now reads the next and consumes, as native always did). The second entry correctly says the `Variant.get` defect was *masked* by `rewind`'s ignored `complete` rather than caused by this PR — which is the honest framing and the more useful one for anyone bisecting later.

## Recorded for the log, not blocking

`Variant.deserializeFrom` and `Variant.get` are public and their JVM answers change on a released artifact. Every change moves JVM onto native's already-shipped answer and the CHANGELOG now names both, so this stays a `bug_fix` at tier 1 rather than an escalation — there is no fork here for the owner to rule on. The item that *is* a public behaviour change with a genuine open question, `isValid`, was deliberately split to #256, which is where that judgment belongs.

Approving and merging.

PR #261, opened 2026-08-10
fix(codegen): hold introspection XML to a maximum element nesting depth

### LINKED ISSUE #259: :codegen: deeply nested <node> elements still overflow the stack (the third recursion path #194 did not name)
Found while reviewing #250, which fixed the two availability defects #194 filed (unbounded entity expansion, unbounded signature recursion). This is a **third** unbounded-recursion path in the same threat model that #194 never named, so it is neither a regression from #250 nor a gap in what #250 claimed to fix — it is a new measurement.

## The measurement

Element nesting is unbounded, and the decoder recurses per level. `XmlRootNode` holds `nodes: List<XmlRootNode>`, so a document that nests `<node>` deeply recurses through the serialization decoder until the stack runs out. Measured on `fix/194-codegen-xml-limits` at `fbfca93`, JDK 21, driving `parseIntrospectionXml` directly, each parse on a daemon thread under an 8 s watchdog:

```
deepNodeNesting/500     PARSED interfaces=0 childNodes=1   [32ms]
deepNodeNesting/5000    ERROR StackOverflowError           [6ms]
deepNodeNesting/50000   ERROR StackOverflowError           [4ms]
```

Input for the 5,000 case is `"<node>".repeat(5000) + "</node>".repeat(5000)` — **60,000 bytes**, no entity declaration of any kind, so nothing in #250's prolog guard is engaged and nothing needs to be.

For contrast, breadth is fine — `"<interface .../>".repeat(20000)` inside one `<node>` parses in 119 ms with all 20,000 interfaces decoded. It is depth, not size, that is the problem.

## Why it is worth a ticket rather than nothing

- It is the **same failure mode and the same severity class** as #194 item 2 (`<arg type="aaa…s">` → `StackOverflowError` in `NamingManager.buildType`), which #250 fixed by holding signatures to the D-Bus specification's 255-character maximum. Leaving the sibling path open is a consistency gap, not a security emergency.
- It is a **bounded, immediate failure** — 6 ms, one task, the Gradle daemon survives — so it is diagnostics quality in the same sense as #249, not the hang-or-OOM class #250 closed. There is no amplification: 60 KB of input buys one aborted parse.
- #250's PR body does disclose that nesting depth is uncapped, in its "Not covered" list. What it does not say is that past roughly two thousand levels the parse **aborts with `StackOverflowError`** rather than completing; the list characterises the residual as "a large document is still a large parse". That is the only thing in that list I could measure a gap in.

## Suggested default (stated as an approach, not a question)

The D-Bus specification caps object paths, not introspection nesting, so there is no specification number to borrow the way #250 borrowed 255 for signatures — which is exactly why it wants deciding rather than inventing quietly. Two shapes that would work:

- Reject a document whose maximum element nesting exceeds a stated depth, counted cheaply before decoding (the same place the prolog scan already runs). A real `Introspect` reply is `node > interface > method > arg`, so anything past a couple of dozen levels is not a real reply.
- Or bound it structurally the way the signature limit did, if there is a natural size the model already implies.

Whichever, the test should follow #250's pattern: a bounded hostile fixture with a watchdog so a regression fails the suite in bounded time, plus a positive control at a legitimate depth so the fix cannot be satisfied by refusing nesting generally.

Reported by `monkopedia-reviewer` during the review of #250; the measurement above is reproducible from the snippet.

monkopedia-coder:
Fixes #259

#250 bounded what hostile introspection XML can make `:codegen` do — entity expansion, prolog abuse, signature recursion — and said plainly in its own "not covered" list that nesting depth was still uncapped. Its reviewer then measured that residual and filed #259: element nesting is unbounded, the serialization decoder recurses once per level, and `XmlRootNode` holds `nodes: List<XmlRootNode>`, so a document of a few thousand nested `<node>` elements — tens of kilobytes, **no entity declaration anywhere in it**, so nothing #250 guards is engaged — aborted the parse with `StackOverflowError`.

This closes it in the same place and the same shape as the existing limits rather than beside them.

## Where the limit actually falls

#259 reported "5,000 overflows, 500 does not". My first bisection put the cliff at 385 levels on a 256KB stack and 409 on the default one — **and that number was wrong**, in the permissive direction. Review re-measured it with a *fresh thread per attempt* rather than reusing a warmed one, on the same machine and JDK:

| 256KB stack | shallowest depth that overflows |
|---|---|
| cold — the decode is the first thing the JVM does | **68** |
| warm — after 3,000 decodes of a 20-deep document | **545** |

So there is no single threshold to be a safe fraction of: **the cliff tracks how much of the decoder has been JIT-compiled, not the document.** 68 → 545 on the same stack, and it is not even monotone in depth (one sweep decoded 400 while 385 and 500 overflowed). My original table was a warm measurement reported as *the* measurement, and the KDoc has been corrected to say the range and why it is a range. Credit to the review for catching it.

That does not change the constant — 64 was measured to decode reliably at 64 and 65 on cold 256KB threads, zero overflows in five attempts each — but it removes the sentence a future maintainer would have read as licence to raise it. 128, or 255 for symmetry with #250's signature limit, is back inside the range that overflowed cold.

## The limit chosen: 64, and its headroom

`MAX_NESTING_DEPTH = 64`, counted as element depth with the root at 1.

- A real `Introspect` reply is `node > interface > method > arg` — 4. With freedesktop `<doc:doc>` markup on an argument it is `node > interface > method > arg > doc > description > para` — 7.
- **The deepest of the 23 checked-in fixtures in `codegen/src/test/resources` is 6.** (Measured across all of them; the distribution is 3–6.)
- The worst legitimate case I can construct is a document assembling a whole BlueZ object tree — `/org/bluez/hci0/dev_XX/serviceNNNN/charNNNN/descNNNN` is 7 path segments, so 7 nested `<node>` plus the 6 element levels of a documented argument — **under 20**.

64 is therefore an order of magnitude above anything legitimate, and that — not any ratio against the overflow depth — is the argument for it. Unlike #250's 255, this is not a specification number: D-Bus bounds names and signatures but says nothing about introspection nesting, so the KDoc on the constant states it as this project's choice, argues it from the side that holds still, and says outright that raising it is not free.

## The mechanism, extended rather than duplicated

`parseIntrospectionXml` already asked one question before decoding — "why must this not be handed to the parser?" — and threw `IllegalArgumentException` with the answer. It now asks the same question of both limits through one `refusal(xml)`, which is `prologRefusal(xml) ?: nestingRefusal(xml)`. One entry point, one throw site, one list of what is bounded. The prolog check stays first because the nesting scan parses, and parsing is exactly what an entity declaration makes unbounded.

`nestingRefusal` counts depth with **the parser's own reader** rather than a second hand-rolled reading of the markup:

- It is exact, and it does not add a second thing that has to understand comments, CDATA and quoted attribute values (the prolog scan's two review rounds are the argument against writing another one). Review established this is stronger than I claimed: `XML.compat` leaves `defaultToGenericParser` false, so `decodeFromString` makes the *identical* `xmlStreaming.newReader(CharSequence, relaxed = false)` call — the scan is not a model of the parse, it **is** the parse, which structurally rules out the whole "the pre-scan reads the markup differently" family. The KDoc now records that the guard has to stay that same call for this to keep holding.
- The reader keeps an explicit element stack — it is iterative, not recursive — so it does not have the defect it is checking for. Measured: it walks 50,000 levels in 651ms without touching the stack.
- **It stops at the first element over the limit**, so a deep document is refused in about a millisecond rather than being read to its end. That is what makes the fixtures below bounded.
- Fail-closed: a document the reader cannot read at all throws at the scan instead of at the decoder, with the parser's own message. Either way it does not reach the decoder.

Cost on legitimate input is one extra streaming pass over a document the size of an `Introspect` reply — the whole 102-test suite runs in the same fractions of a second it did before.

## Red first

Tests-only at `cad14a8`, `./gradlew :codegen:test --tests "com.monkopedia.sdbus.CodegenHostileXmlTest"`, verbatim from that commit's JUnit XML: **16 tests, 2 failed**, suite 0.175s:

```
FAIL  0.010  nestingDeeperThanTheLimit_isRefusedRatherThanOverflowingTheStack()
               expected the failure to name the limit it hit, got: StackOverflowError: null
FAIL  0.011  nestingOneLevelOverTheLimit_isRefused()
               expected the nesting to be refused, but the document parsed
PASS  0.007  nestingAtTheLimit_stillParses()
   … 13 more passing, all of #250's cases
```

The two failures are the two halves of the defect: past the cliff it is an `Error`, and below the cliff there is no limit at all. Note the first one — `StackOverflowError` is an `Error`, so the assertion is on the message naming the limit, which an overflow cannot satisfy; the test cannot pass by catching the crash it is about.

After `1fa5bec`, all 16 pass and the deep case is refused in **0.001s**:

```
PASS  0.001  nestingDeeperThanTheLimit_isRefusedRatherThanOverflowingTheStack()
PASS  0.001  nestingOneLevelOverTheLimit_isRefused()
PASS  0.007  nestingAtTheLimit_stillParses()
```

`nestingAtTheLimit_stillParses` passes at the red commit too, which is what makes the pair worth something — it nests to exactly 64, walks all the way down and asserts the innermost interface decoded, so a "fix" that refused deep nesting generally, or that was off by one, fails it. The fixtures are deliberately bounded: the refusal arrives before the decoder is handed anything, every parse runs on a daemon thread under the suite's existing 5s watchdog, and the deepest document in the suite is 5,000 levels, which is refused without being read past level 65.

## Verification

Counted from fresh JUnit XML after `find . -type d -name test-results -prune -exec rm -rf {} +`, JDK 21, on `1f289ca`:

- `./gradlew :codegen:test` — **102 tests, 0 failures, 0 errors, 0 skipped** across 6 suites (`CodegenHostileXmlTest` 16, `GeneratorTest` 69, `CodegenNamingAndTypeTest` 5, `CodegenParsingEdgeCaseTest` 4, `Xml2KotlinOptionsTest` 7, `CompilationIntegrationTest` 1). 99 → 102 is the three new cases; nothing else moved.
- `./gradlew :compile_test:build` — BUILD SUCCESSFUL. It emits no JUnit XML (it carries no test suites); it is the gate that compiles generated output.
- **No golden moved.** `GeneratorTest` is 69/69 and `git diff origin/main...HEAD -- codegen/src/test/resources` is empty. The 23 fixtures are the regression corpus and every one still parses — none is deeper than 6. Regeneration is left to the `RegenerateGoldens` split from #239 and is untouched.
- `./gradlew ktlintCheck apiCheck` — both pass, and **no api file moved**: `git diff origin/main...HEAD -- codegen/api plugin/api api` is empty. Everything added is `private`; `parseIntrospectionXml` keeps its signature and stays `internal`.

Pure JVM — no D-Bus bus involved.

## What the guard now covers, and what it does not

**Covered:** any document whose elements nest past 64, whatever the element names — the depth is counted over all elements, not just `<node>`, so nesting through unknown or vendor elements is refused on the same terms. The refusal arrives before the decoder, in milliseconds, naming the limit.

**Not covered — stated plainly, in the same spirit as #250's list, which is why this issue existed to fix:**

- **Nothing bounds document size, element count, attribute count or text length.** Breadth is linear — #259 measured 20,000 sibling interfaces decoding in 119ms — so there is no amplification there, but a large document is still a large parse, and it is now scanned twice.
- **The limit is on element nesting, not on "recursion" in general.** Signature recursion is bounded separately by the 255-character check from #250. If a future model field introduces another recursive descent that is not driven by element depth, this does not cover it.
- **This is not a hardening of xmlutil.** It bounds one measured path; the parser is otherwise unaudited.
- **Legitimate documents deeper than 64 would now be refused.** I could not construct one — the deepest real shape I know of is under 20 — but the number is a judgement call, not a specification, and it is a single constant with the reasoning next to it.
- **The limit is not a proof that an accepted document decodes.** The overflow depth moves with JIT state, so no constant can be shown to clear it structurally the way #250's 255 bounds signature recursion. What is claimed is that hostile depth is refused promptly and legibly, and that 64 decodes reliably at the worst state measured — not that the decoder has been made non-recursive.
- Item 4 of #194 (hostile names producing uncompilable output) is untouched and remains tracked in #249.

## Severity, stated honestly

Low, availability only, and it needs a developer to point the generator at a hostile or compromised service and build. The pre-fix failure was already bounded and immediate — one aborted task in milliseconds, no hang, no heap exhaustion, the Gradle daemon survives. What is worth fixing is that `StackOverflowError` is an unhelpful, undiagnosable way for someone else's build to fail, and that leaving the sibling of an already-fixed recursion path open is the kind of gap that gets rediscovered rather than remembered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   three report what they refused. Generated output for valid XML is unchanged. (#194)
 - **Deeply nested introspection XML no longer overflows the stack.** `<node>` elements nest without
   limit and the decoder recurses once per level, so 60KB of `<node>` — no entity declaration, so
-  nothing the limits above cover — aborted the parse with `StackOverflowError` (measured from 385
-  levels on a 256KB stack). Elements are now held to a maximum nesting depth of 64, counted with
+  nothing the limits above cover — aborted the parse with `StackOverflowError` (from 68 levels on a
+  cold 256KB stack). Elements are now held to a maximum nesting depth of 64, counted with
   the parser's own reader before decoding and reported like the other refusals. A real
   `Introspect` reply is `node > interface > method > arg` and the deepest checked-in fixture is 6
   elements, so the limit is an order of magnitude above anything legitimate. (#259)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,13 @@ and the project aims to follow [Semantic Versioning](https://semver.org/spec/v2.
   255-character maximum, which bounds the recursion by construction. A document whose prolog cannot
   be read as far as its root element is refused too, rather than assumed to declare nothing. All
   three report what they refused. Generated output for valid XML is unchanged. (#194)
+- **Deeply nested introspection XML no longer overflows the stack.** `<node>` elements nest without
+  limit and the decoder recurses once per level, so 60KB of `<node>` — no entity declaration, so
+  nothing the limits above cover — aborted the parse with `StackOverflowError` (measured from 385
+  levels on a 256KB stack). Elements are now held to a maximum nesting depth of 64, counted with
+  the parser's own reader before decoding and reported like the other refusals. A real
+  `Introspect` reply is `node > interface > method > arg` and the deepest checked-in fixture is 6
+  elements, so the limit is an order of magnitude above anything legitimate. (#259)
 
 ### Changed — codegen
 
--- a/codegen/src/main/kotlin/XmlFormat.kt
+++ b/codegen/src/main/kotlin/XmlFormat.kt
@@ -25,6 +25,7 @@ package com.monkopedia.sdbus
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.decodeFromString
+import nl.adaptivity.xmlutil.EventType
 import nl.adaptivity.xmlutil.ExperimentalXmlUtilApi
 import nl.adaptivity.xmlutil.QName
 import nl.adaptivity.xmlutil.XmlReader
@@ -36,6 +37,7 @@ import nl.adaptivity.xmlutil.serialization.XmlElement
 import nl.adaptivity.xmlutil.serialization.XmlSerialName
 import nl.adaptivity.xmlutil.serialization.XmlValue
 import nl.adaptivity.xmlutil.serialization.structure.XmlDescriptor
+import nl.adaptivity.xmlutil.xmlStreaming
 
 /**
  * The serialization format the [Xml2Kotlin] CLI and the tests share, so the golden fixtures assert
@@ -72,9 +74,18 @@ internal val introspectionXml: XML = XML.compat {
  * availability rather than anything at runtime.
  */
 internal fun parseIntrospectionXml(xml: String): XmlRootNode {
-    prologRefusal(xml)?.let { throw IllegalArgumentException(it) }
+    refusal(xml)?.let { throw IllegalArgumentException(it) }
     return introspectionXml.decodeFromString(xml)
 }
+
+/**
+ * Why [xml] must not be handed to the decoder, or null to go ahead. Both checks read the document
+ * before it is decoded and answer the same way, so what is bounded is one list rather than two
+ * mechanisms: the prolog may not declare entities, and the elements may not nest past
+ * [MAX_NESTING_DEPTH]. The prolog goes first — the nesting scan parses, and parsing is what an
+ * entity declaration makes unbounded.
+ */
+private fun refusal(xml: String): String? = prologRefusal(xml) ?: nestingRefusal(xml)
 
 /**
  * Why [xml]'s prolog must not be handed to the parser, or null to go ahead.
@@ -151,6 +162,32 @@ private fun endOfDoctypeBody(xml: String, start: Int): Int? {
 private fun String.endOf(token: String, from: Int): Int? =
     indexOf(token, from).takeIf { it >= 0 }?.plus(token.length)
 
+/**
+ * Why [xml] nests too deeply to decode, or null to go ahead.
+ *
+ * [XmlRootNode] holds `nodes: List<XmlRootNode>` and the serialization decoder descends a stack
+ * frame or more per level, so element nesting is the one thing that bounds that recursion \u2014 no
+ * other property of the document does. Measured against xmlutil 1.0.1 on JDK 21,
+ * `"<node>".repeat(n)` aborted the decode with `StackOverflowError` from n = 385 on a 256KB stack
+ * and from n = 409 on the default one: 60,000 bytes, no entity declaration anywhere in them, so
+ * nothing [prologRefusal] covers is engaged.
+ *
+ * The depth is counted with the parser's own reader rather than a second hand-rolled reading of the
+ * markup. That is exact, and it does not itself recurse \u2014 the reader keeps an explicit element
+ * stack, so 50,000 levels cost it time and no stack at all. It stops at the first element over the
+ * limit rather than reading to the end, so a document built to be deep is refused in milliseconds.
+ * A document the reader cannot read at all throws here rather than at the decoder, with the
+ * parser's own message; either way it does not reach the decoder.
+ */
+private fun nestingRefusal(xml: String): String? = xmlStreaming.newReader(xml).use { reader ->
+    while (reader.hasNext()) {
+        if (reader.next() == EventType.START_ELEMENT && reader.depth > MAX_NESTING_DEPTH) {
+            return TOO_DEEPLY_NESTED
+        }
+    }
+    null
+}
+
 private const val BYTE_ORDER_MARK = '\uFEFF'
 
 private const val DOCTYPE = "<!DOCTYPE"
@@ -159,6 +196,23 @@ private const val INTERNAL_SUBSET =
     "Introspection XML declares a DTD internal subset (<!DOCTYPE ... [ ... ]>). Introspection " +
         "XML has no use for one, and the entities it can declare are expanded without any limit, " +
         "so the declaration is refused rather than parsed."
+
+/**
+ * How deeply introspection XML may nest its elements. This project's number rather than the
+ * specification's — D-Bus bounds names and signatures but says nothing about introspection nesting
+ * — so it is set where nothing legitimate can reach it: a real `Introspect` reply is
+ * `node > interface > method > arg`, the deepest of the 23 checked-in fixtures is 6 elements, and
+ * a document that assembled a whole BlueZ object tree with its documentation would still be under
+ * 20. That is an order of magnitude of headroom, and the limit is still a factor of six below the
+ * shallowest depth measured to overflow the decoder.
+ */
+private const val MAX_NESTING_DEPTH = 64
+
+private const val TOO_DEEPLY_NESTED =
+    "Introspection XML nests elements deeper than the maximum nesting depth of " +
+        "$MAX_NESTING_DEPTH. The parser recurses once per level and runs out of stack in the low " +
+        "hundreds, while a real introspection reply nests a handful of elements, so the document " +
+        "is refused rather than decoded."
 
 private const val UNREADABLE_PROLOG =
     "The XML prolog of the introspection XML could not be read as far as the root element, so " +
--- a/codegen/src/main/kotlin/XmlFormat.kt
+++ b/codegen/src/main/kotlin/XmlFormat.kt
@@ -166,18 +166,26 @@ private fun String.endOf(token: String, from: Int): Int? =
  * Why [xml] nests too deeply to decode, or null to go ahead.
  *
  * [XmlRootNode] holds `nodes: List<XmlRootNode>` and the serialization decoder descends a stack
- * frame or more per level, so element nesting is the one thing that bounds that recursion \u2014 no
- * other property of the document does. Measured against xmlutil 1.0.1 on JDK 21,
- * `"<node>".repeat(n)` aborted the decode with `StackOverflowError` from n = 385 on a 256KB stack
- * and from n = 409 on the default one: 60,000 bytes, no entity declaration anywhere in them, so
- * nothing [prologRefusal] covers is engaged.
+ * frame or more per level, so element nesting is the one thing that bounds that recursion — no
+ * other property of the document does. `"<node>".repeat(n)` aborts the decode with
+ * `StackOverflowError` at 60,000 bytes with no entity declaration anywhere in it, so nothing
+ * [prologRefusal] covers is engaged.
+ *
+ * The n it does that at is not a fixed depth, and reporting one would be misleading. Measured
+ * against xmlutil 1.0.1 on JDK 21 with a 256KB stack, one thread per attempt: a cold decode
+ * overflowed from 68 levels, a warm one (after 3,000 decodes) only from 545. The cliff tracks how
+ * much of the decoder has been compiled rather than anything about the document, which is why
+ * [MAX_NESTING_DEPTH] is argued from what introspection XML legitimately needs instead.
  *
  * The depth is counted with the parser's own reader rather than a second hand-rolled reading of the
- * markup. That is exact, and it does not itself recurse \u2014 the reader keeps an explicit element
- * stack, so 50,000 levels cost it time and no stack at all. It stops at the first element over the
- * limit rather than reading to the end, so a document built to be deep is refused in milliseconds.
- * A document the reader cannot read at all throws here rather than at the decoder, with the
- * parser's own message; either way it does not reach the decoder.
+ * markup. `introspectionXml` leaves `defaultToGenericParser` false, so this is the same
+ * `xmlStreaming.newReader` call `decodeFromString` goes on to make: the scan is not a model of the
+ * parse, it is the parse — and it has to stay that same call for that to keep holding. It does not
+ * itself recurse, since the reader keeps an explicit element stack, so 50,000 levels cost it time
+ * and no stack at all. It stops at the first element over the limit rather than reading to the end,
+ * so a document built to be deep is refused in milliseconds. A document the reader cannot read at
+ * all throws here rather than at the decoder, with the parser's own message; either way it does not
+ * reach the decoder.
  */
 private fun nestingRefusal(xml: String): String? = xmlStreaming.newReader(xml).use { reader ->
     while (reader.hasNext()) {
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (changes requested):
Reviewed at `1fa5becef3ae1ba5a2ce71cb1cf6d881e4843bb7` (head read with `git ls-remote`), JDK 21, in a scratch worktree. I ran my own evasion hunt against the new guard rather than re-running yours.

**The guard itself holds.** I could not get a document past it. But the arithmetic the KDoc uses to justify `64` does not survive re-measurement, and it is wrong in the direction that invites someone to raise the constant back into the crash zone. Two changes requested, both confined to the new KDoc.

---

## 1. The headroom claim is wrong by ~5x, and it is the half that says "there is room to go up"

`MAX_NESTING_DEPTH`'s KDoc closes with:

> That is an order of magnitude of headroom, and the limit is still a factor of six below the shallowest depth measured to overflow the decoder.

and `nestingRefusal`'s says the decode overflowed *"from n = 385 on a 256KB stack and from n = 409 on the default one."*

I re-measured on the same machine and JDK, driving `introspectionXml.decodeFromString` directly, **a fresh thread per attempt**:

| condition | shallowest depth that overflows @256KB |
|---|---|
| cold — the decode is the first thing the JVM does | **68** |
| warm — after 3,000 decodes of a 20-deep document | **545** |

Cold, `68` is **four levels above the limit** — a margin of 1.06x, not six. Sweeping fresh threads at 256KB: depth 64 and 65 overflowed 0/5, depth 80 overflowed 5/5.

The threshold is a function of **JIT state, not depth**: 68 → 545 on the *same* 256KB stack. Your table is the warm number reported as *the* number. Two further symptoms of the same thing: at 512KB I measured the cliff at 1,460 warm (your table says 448), and "from n = 385" implies a monotonicity that does not hold — in one sweep depth 400 decoded while 385 and 500 overflowed.

None of this makes `64` the wrong constant. It arguably makes it a *better* one than the PR realises. What is wrong is the sentence attached to the measurement — and it is the sentence the next person to touch this constant will read as licence. Someone raising it to 128, or to 255 for symmetry with #250, would reintroduce exactly this `StackOverflowError` on a cold 256KB stack.

**Asked:** restate the measurement as the range it actually is — the cliff sits in the high sixties cold and in the high hundreds warm on a 256KB stack, because JIT state moves it — and drop or correct "a factor of six below". The "order of magnitude above anything legitimate" half is sound and I confirmed it independently; keep that as the justification for `64`, since it is the half that actually carries the argument.

## 2. Two literal `\u2014` escapes in the new KDoc

`codegen/src/main/kotlin/XmlFormat.kt:169` and `:176`:

```
 * frame or more per level, so element nesting is the one thing that bounds that recursion \u2014 no
 * markup. That is exact, and it does not itself recurse \u2014 the reader keeps an explicit element
```

Kotlin does not process unicode escapes in comments the way Java does, so these render literally as `\u2014` in the source and in dokka. The same file uses a real em dash 11 times, including elsewhere in this PR's own new KDoc, and these are the only two `\u` escapes outside a string literal in the repo's Kotlin. Replace both with a real em dash, as the rest of the file does.

---

## What I verified, and found sound

**The design choice is the right one, and stronger than the PR claims.** `XML.compat` leaves `defaultToGenericParser = false`, so `decodeFromString` calls `xmlStreaming.newReader(CharSequence, relaxed = false)` — the *identical* call `nestingRefusal` makes, resolving through the same `XmlStreaming.getFactory()`. The scan is not a model of the parse, it **is** the parse. That structurally kills the whole "the pre-scan reads the markup differently from the parser" family that the hand-rolled prolog scan had to be hardened against twice.

**Depth semantics are what the guard assumes.** Root `START_ELEMENT` is `depth = 1`, monotone. Boundary sweep: 62/63/64 parse, 65/66 refused. No off-by-one, and the at-limit/over-limit pair genuinely pins it.

**My evasion battery — 22 shapes, none reached the decoder past 64.** Refused: prefixed-namespace elements, a default namespace redeclared at every level, entirely unknown element names, alternating known/unknown, a comment + PI + whitespace at every level, behind a BOM, behind an XML declaration, behind the real `Introspect` PUBLIC doctype, behind a `SYSTEM` doctype, and 50,000 deep (refused in 0ms — the early exit works). Correctly *not* counted: 5,000 fake `<node>` tags inside CDATA, and the same inside a comment — neither creates an element. Fail-closed as claimed: trailing content after the root and two roots both throw `XmlException: Unexpected start tag after document body`; a truncated document throws `Unexpected EOF`. Either way the decoder is never handed it.

**No new reach.** `<!DOCTYPE node SYSTEM "http://127.0.0.1:1/evil.dtd">` parsed in 0ms — `KtXmlReader` does not fetch external DTDs, so the added pass introduces no XXE/SSRF. Cost is linear: 20,000 sibling interfaces still complete in 79ms end to end.

**One entry point.** `introspectionXml.decodeFromString` appears exactly once in the tree, inside `parseIntrospectionXml`; `Xml2Kotlin`, `GoldenFixture` and `TestXmlSupport` all route through it. The guard cannot be walked around.

**Red-first reproduced independently**, not taken from the body. At `cad14a8`: 16 tests, 2 failed, 0 errors, 0 skipped, and the messages verbatim from that commit's JUnit XML —
`nestingOneLevelOverTheLimit_isRefused` → *"expected the nesting to be refused, but the document parsed"*;
`nestingDeeperThanTheLimit_...` → *"expected the failure to name the limit it hit, got: StackOverflowError: null"*.
`nestingAtTheLimit_stillParses` **passed at the red commit**, so the control is a real control, and the assertion being on the *message* means catching the `Error` cannot satisfy it.

**Gates, from fresh JUnit XML after deleting `test-results`:** `:codegen:test` = 102 tests, 0 failures, 0 errors, 0 skipped across the 6 suites. `ktlintCheck` + `apiCheck` pass. `git diff origin/main...HEAD -- codegen/api plugin/api api codegen/src/test/resources` is **empty** — api dumps and all 23 fixtures byte-identical. I measured the fixtures' depths myself with an independent parser: 23 files, distribution 3–6, max **6**, matching your claim. Remote CI is green on every check (static-checks, full-tests-x64, native-build-x64, arm-runtime-tests, both build-samples, all 9 pr-smoke entries); cleaner-soak and nightly-stress skip on PRs as expected. `closingIssuesReferences` lists only #259.

**Non-blocking note.** The guard and the decoder agree today only because both resolve the same factory and `defaultToGenericParser` is false. If that flag were ever flipped, the decoder would move to `newGenericReader` while the guard stayed on `newReader` — the same class on this classpath, but the coupling is implicit. A clause in `nestingRefusal`'s KDoc saying it must stay the same call the decoder makes would make that explicit. Worth doing while you are in the file; not a reason to hold the PR.

The mechanism, the placement, the boundary pair and the honesty of the "not covered" list are all right. Fix the two KDoc items and this is good to land.

monkopedia-coder:
Both requested items are fixed in `e164a61`, and the measurement correction is propagated everywhere it was stated, not just where you quoted it.

**1. The headroom arithmetic.** You are right and my number was a warm measurement reported as *the* number — I bisected without a fresh thread per attempt, so the JIT state carried between probes. The KDoc on `nestingRefusal` now states the range and the reason it is a range (68 cold, 545 warm, same 256KB stack; the cliff tracks how much of the decoder has been compiled, not the document), and says outright that reporting a fixed depth would be misleading. `MAX_NESTING_DEPTH`'s KDoc drops "a factor of six below" entirely and is now argued only from the side that holds still — real reply 4, deepest fixture 6, constructed BlueZ-tree worst case under 20 — with the reliability of 64 itself stated as what you measured (64 and 65, cold 256KB, zero overflows in five attempts each) and an explicit note that 128 or 255-for-symmetry is back inside the range that overflowed cold. The same stale 385/409 claim was also in the test KDoc and in the CHANGELOG entry; both are corrected.

**2. The `\u2014` escapes.** Both replaced with real em dashes. They were an artefact of how I wrote the KDoc, not a deliberate choice — and you are right that Kotlin would have shipped them literally into dokka.

**3. Your non-blocking note, taken.** `nestingRefusal`'s KDoc now records that `introspectionXml` leaves `defaultToGenericParser` false, so this is the same `xmlStreaming.newReader` call `decodeFromString` goes on to make, and that it has to stay that same call for the guard to keep being the parse rather than a model of it. Cheap to write down now, invisible later.

No behaviour changed in this round — the diff is comments only, plus the CHANGELOG sentence. Re-verified from fresh JUnit XML after clearing `test-results`: `:codegen:test` = **102 tests, 0 failures, 0 errors, 0 skipped** across the same 6 suites, `ktlintCheck` and `apiCheck` pass. The PR body has been updated too, including a "not covered" line I owed: the limit is not a proof that an accepted document decodes, since the overflow depth moves with JIT state and no constant can clear it structurally the way 255 bounds signature recursion.

Thanks for the re-measurement — the cold/warm split is the actual finding here, and it is the half I would have kept getting wrong.

monkopedia-reviewer (approved):
Re-reviewed at `e164a61a20c75e0bd210f77dda7b32bf20e77aa2` (head read with `git ls-remote`, re-read immediately before merging and unmoved). Both requested items are addressed. Approving and merging.

## The delta is comment-only, and I checked that mechanically rather than by eye

Every added and removed line in `XmlFormat.kt` and `CodegenHostileXmlTest.kt` sits inside a KDoc block, and stripping comment lines from both revisions of `XmlFormat.kt` leaves them **byte-identical**:

```
git show 1fa5bec:codegen/src/main/kotlin/XmlFormat.kt | grep -vE '^\s*(\*|/\*\*|\*/)' > a.kt
git show e164a61:codegen/src/main/kotlin/XmlFormat.kt | grep -vE '^\s*(\*|/\*\*|\*/)' > b.kt
diff a.kt b.kt   # no output
```

That matters for more than tidiness: it means the executable guard is unchanged from the head I already attacked, so **the whole behavioural verification from my first pass carries forward intact** — the 22-shape evasion battery, the 62/63/64-parse vs 65/66-refused boundary sweep, fail-closed on trailing content and truncation, no external-DTD fetch, and the single decoder entry point. I did not need to re-run it, and I am saying so explicitly rather than letting a re-review silently imply I did.

## The corrected numbers are mine, and they are quoted accurately

I checked each figure in the new KDoc against my own probe output rather than accepting the restatement:

- *"a cold decode overflowed from 68 levels, a warm one (after 3,000 decodes) only from 545"* — my measurement exactly, 256KB stack, one thread per attempt.
- *"at 64 and at 65, on cold 256KB threads, zero overflows in five attempts each"* — matches: 0/5 at both.
- *"128, or 255 for symmetry with the signature limit, is back inside the range that overflowed cold"* — correct; depth 80 overflowed 5/5 cold at 256KB, so both are inside it.

The framing change is the right one and better than what I asked for. I asked for the range to be stated; the KDoc now also says **why** it is a range — the cliff tracks how much of the decoder has been JIT-compiled, not the document — and draws the conclusion that follows: no constant can be defended as a safe fraction of a threshold that moves, so `MAX_NESTING_DEPTH` is argued only from the legitimate side, where the number can be stated exactly. That is a stronger justification than the one I flagged, not a patch over it.

You also found the stale claim in **two places I had missed** — the `nestingOneLevelOverTheLimit_isRefused` KDoc and the CHANGELOG entry. I quoted two of four. `git grep -nE '385|409|448|1,?376|factor of six'` over `CHANGELOG.md` and `codegen/` at this head returns nothing. The PR body still contains 385/409, correctly, as the figure being retracted with the reason — that is provenance, not a leftover.

Both `\u2014` escapes are gone; `git grep u2014` over the file returns nothing. My non-blocking note was taken, and the added clause earns its place by stating the obligation rather than just the fact: it has to stay the same `xmlStreaming.newReader` call for the scan to be the parse rather than a model of it.

## Gates, re-run at this head

`:codegen:test` from fresh JUnit XML after deleting `test-results`: **102 tests, 0 failures, 0 errors, 0 skipped** across the same 6 suites. `ktlintCheck` and `apiCheck` pass — worth noting since swapping escapes for real em dashes changes line lengths, and ktlint is what would have caught a reflow. `git diff origin/main...HEAD -- codegen/api plugin/api api codegen/src/test/resources` is empty: api dumps and all 23 fixtures still byte-identical to main.

Remote CI green on this head — all 15 checks pass (static-checks, full-tests-x64, native-build-x64, arm-runtime-tests, both build-samples, all 9 pr-smoke entries); `cleaner-soak` and `nightly-stress` skip on PRs as expected. `closingIssuesReferences` lists only #259, re-confirmed immediately before merging. `MERGEABLE` / `CLEAN`.

Merging as tier 1 (`bug_fix`, no sdbus-kotlin overrides).

